Issue #104, opened 2022-04-05
Validate UEIs


Jkrzy:
## Background
As a user I can enter an UEI and receive feedback on the validity of the value so that I can be sure I've entered the correct UEI.

**Note** 

- UEIs are [specifically structured](https://www.gsa.gov/about-us/organization/federal-acquisition-service/office-of-systems-management/integrated-award-environment-iae/iae-systems-information-kit/uei-technical-specifications-and-api-information) and maintained by SAM.gov.
- Sam.gov offers an Entity API which may be helpful here: https://open.gsa.gov/api/entity-api/
- We'll need to handle all validation on the backend but, where possible we want to also perform validations as close as possible to data entry.  
  - For example, we can check that an entry is 12 character before a user submits a POST via the frontend.

**Current mock of interface where UEIs are entered**
![Screen Shot 2022-04-05 at 3 23 07 PM](https://user-images.githubusercontent.com/3485564/161833591-1e816616-c3fa-4f90-b4cf-4cec5c376bf6.png)

## We'll know we're done when...
- [ ] When a user submits a UEI not meeting the standards, they receive a helpful error message 

## Tasks
- [ ] Examine spec and identify checks which may be feasible on the client side (Engineering)
- [ ] Propose and implement a layered approach to UEI validation (Engineering)
- [ ] Mock up error states for UEI validation errors (Design)
- [ ] Front end implementation to perform validation and render error message(s) (Engineering)

NicholasKramer:
Sharing our proposed approach to validating UEIs: @Jkrzy and I contemplated a layered approach to UEI validation where we followed the UEI spec to validate locally before verifying with the SAM.gov API.  The following checks can be performed locally, and if they pass, finalize verification by reaching out to the SAM.gov API:

* The Unique Entity ID is a 12-character, alphanumeric value that adheres to the following rules:
* The letters “O” and “I” are not used to avoid confusion with zero and one.
* The first character is not zero to avoid cutting off digits that can occur during data imports, for example, when importing data into spreadsheet programs.
* Nine-digit sequences are not used in the identifier to avoid collision with the nine-digit DUNS Number or Taxpayer Identification Number (TIN).

Any thoughts?  Suggestions?  I'm happy to document this elsewhere if needed.

jadudm:
Makes sense to me. Saves us from sending complete randomness to the API. And, the proposed local checks are light/straight-forward. I don't think this level of decision requires an ADR, if that's part of the question.

[...]

PR #142, opened 2022-04-19
Added local UEI validators

NicholasKramer:
Added local validators to ensure adherence to UEI standards.

According to the [UEI specification](https://www.gsa.gov/about-us/organization/federal-acquisition-service/office-of-systems-management/integrated-award-environment-iae/iae-systems-information-kit/uei-technical-specifications-and-api-information), there are several checks to determine a valid UEI. 

This performs the initial checks to ensure a valid UEI, including: 

> * The letters “O” and “I” are not used to avoid confusion with zero and one.
> * The first character is not zero to avoid cutting off digits that can occur during data imports, for example, when importing data into spreadsheet programs.
> * Nine-digit sequences are not used in the identifier to avoid collision with the nine-digit DUNS Number or Taxpayer Identification Number (TIN).

### DIFF
--- a/backend/api/serializers.py
+++ b/backend/api/serializers.py
@@ -14,6 +14,12 @@ class EligibilitySerializer(serializers.ModelSerializer):
         return data
 
 
+class UEISerializer(serializers.ModelSerializer):
+    class Meta:
+        model = SingleAuditChecklist
+        fields = ['uei']
+
+
 class SingleAuditChecklistSerializer(serializers.ModelSerializer):
     class Meta:
         model = SingleAuditChecklist
--- a/backend/api/test_serializers.py
+++ b/backend/api/test_serializers.py
@@ -1,5 +1,5 @@
 from django.test import SimpleTestCase
-from .serializers import EligibilitySerializer
+from .serializers import EligibilitySerializer, UEISerializer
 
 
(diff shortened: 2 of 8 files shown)


NicholasKramer:
Alright, I think this is at a good point to ship what we have now as a foundational piece that allows us to start validating UEI's for #104.

The `/sac/ueivalidation` endpoint returns a json "errors" object with field named attributes whose values are the corresponding error message:

Given an invalid UEI input

```JSON
{"uei": "000000000O*"}
```

The API returns

```JSON
{
    "valid": false,
    "errors": {
        "uei": [
            "The UEI should be alphanumeric",
            "The letters “O” and “I” are not used to avoid confusion with zero and one.",
            "The first character is not zero to avoid cutting off digits that can occur during data imports.",
            "Nine-digit sequences are not used in the identifier to avoid collision with the nine-digit DUNS Number or Taxpayer Identification Number (TIN).",
            "Ensure this field has at least 12 characters."
        ]
    }
}
```
Given input

```JSON
{"uei": "ABC123DEF456"}
```

Success looks like this:

```JSON
{
    "valid": true
}
```

Jkrzy (approved):
Good start, thanks @NicholasKramer

PR #144, opened 2022-04-20
Jadudm/public

jadudm:
Opening a PDR pull request for discussion. 

This has been discussed in Slack and meetings, so it may be that discussion in this space is minimal. It documents what feels like consensus from the team at this point.


docs/product/decisions/0003-P-public-repos.md
```diff
@@ -0,0 +1,24 @@
+# 3. Make `gsa-tts/fac` repository public
+
+Introduced: 20220420
+
+## Status
+
+Discussion
```
Jkrzy, line 7:
```suggestion
Decided
```

docs/product/decisions/0002-P-uei.md
```diff
@@ -0,0 +1,42 @@
+# 4. EIN or UEI before recording audits?
```
Jkrzy, line 1:
```suggestion
# 2. Capture only Universal Entity Identifier (UEI)
```

Jkrzy (approved):
A few copy edits, I'm going to accept these suggestions to apply the changes. 

Also assuming the inclusion of the UEI here is intentional.

Issue #128, opened 2022-04-12
Audit submission access page


MelissaBraxton:
**Background**

In the future, non-federal users will have to log in via login.gov. 

It covers the following stories:

1. As an auditor or auditee, I need to be able to specify who will need to certify my new audit (who will "check the box" at the end)
2. As an auditor or auditee, I need to be able to specify who should have access to my new audit
3. As a user, I need information on what will happen when I enter email addresses here so that I feel confident moving to the next step. (this is a front end/content change for now--not anything having to do with the mechanics of notifications)
4. As an auditor or auditee creating a new audit submission, I need the submission access page to make it clear what a “contact” is and what kind of access to the submission they’ll have so that I can give the right access to the right people. 

**We'll know we're done when...**
- [x] A user can enter a valid email address for (validated means any email address for now, for this bit of work. Later we'll need to figure out how we handle whether a particular users can access a particular audit submission)
- [x] When a user enters an invalid email address, they get a helpful error message that helps them recover. 
  
**Tasks**
- [x] [DESIGN/CONTENT] Update mocks based on sprint 1 research (#180) 
- [x] [DESIGN] Attach refined mocks to the issue
- [x] https://github.com/GSA-TTS/FAC/issues/200
- [x] https://github.com/GSA-TTS/FAC/issues/246
- [ ] https://github.com/GSA-TTS/FAC/issues/317
- [x] https://github.com/GSA-TTS/FAC/issues/168

**Definition of Done**
- [ ] Meets acceptance criteria
- [ ] Screen reader - Listen to the experience with a screen reader extension, ensure the information presented in order
- [ ] Keyboard navigation - Run through acceptance criteria with keyboard tabs, ensure it works. 
- [ ] Text scaling - Adjust viewport to 1280 pixels wide and zoom to 200%, ensure everything renders as expected. Document 400% zoom issues with USWDS if appropriate.

Jkrzy:
🤔 Thinking about assumptions here

* At this point in the flow, the Names/Emails being specified **are not** linked directly to User Accounts in the FAC.
* At **some time in the future** we expect a person associated with a given email would authenticate via Login.gov and then be automatically associated with an `SF-SAC` created here.

MelissaBraxton:
Yes, that matches my thinking @Jkrzy.

MelissaBraxton:
One new need statement from sprint 1 research to discuss:

[...]

PR #152, opened 2022-04-22
Thru line from Eligibility->Audit Info->Access steps to create an SF-SAC

Jkrzy:
We expand backend functionality to cover API request-response cycles across and stub out spaces for future enhancement of validation:

> Eligibility -> Auditee Info -> User Access Roles

A large change here is the addition of a `UserProfile` model, where we store user's responses to the initial forms leading up to the creation of a `SingleAuditChecklist`. 

This is an intentional decision to manage the state of the intake process on the backend, specifically for the Frontend being developed in `/FAC-Frontend`. Initially, we anticipate these end-points to be used only by that frontend to provide a multi-step user experience. 

Also included is a new model laying the foundations for #128. There's work to be done here to finalize the handling of inbound access, this PR gives us a starting point and is not intended as final decisions.


backend/audit/test_models.py
```diff
@@ -9,6 +9,6 @@
 class SingleAuditChecklistTests(TestCase):
 
     def test_str_is_id_and_ein(self):
```
NicholasKramer, line 11: Minor: consider changing function name to "test_str_is_id_and_uei"

NicholasKramer:
Awesome work so far @Jkrzy!

backend/audit/test_models.py
Jkrzy: Thanks @NicholasKramer, let's use :octocat: suggestion for things like this.

```suggestion
    def test_str_is_id_and_uei(self):
```

NicholasKramer (approved):

Jkrzy:
Thanks @NicholasKramer, I'll work on working resolving the conflicts here and ping you when ready for another quick look.

NicholasKramer (approved):
Nice work @Jkrzy!

PR #163, opened 2022-04-28
Make test JSON into dicts for better legibility

tadhg-ohiggins:
Long strings: hard to read.
So turn them into real dicts
And then use JSON.


NicholasKramer (approved):
Awesome

PR #181, opened 2022-05-04
Minor tweaks to SAM.gov UEI call

tadhg-ohiggins:
So False is not None
And we need to handle that.
Also, move the key.

-----

*   Change conditional to check for `None` explicitly rather than for a false-y value.
*   Move API key into auth headers.
*   Make the SAM URL an explicit parameter so that we're only accessing global variables from one place.


Jkrzy (approved):
Thanks @tadhg-ohiggins

PR #183, opened 2022-05-04
Add docs about flake8, black, and bandit.

tadhg-ohiggins:
Now that we use them
We need documentation.
This is a first pass.

-----

Add info about how we use and configure `bandit`, `black`, and `flake8`, and how they fit into our GitHub workflows.


Jkrzy (approved):
Thanks @tadhg-ohiggins, looks good!

PR #194, opened 2022-05-10
Tell Docker to migrate before running server

geekygirlsarah:
This adds a line to `docker-compose.yml` so that when creating a new image or `docker-compose up`ing an existing container, that it tries to run all migrations before running the server. Without this, Django complains that the database is out of date and can possibly throw errors later on.


NicholasKramer (approved):
Looks good!!

PR #205, opened 2022-05-13
Upgrade Python 3.10.1 to .4 for cloud.gov builds to resume

geekygirlsarah:
This should fix the Github Action that's failing because cloud.gov buildpacks don't have a Python 3.10.1 version anymore.


NicholasKramer (approved):
Do we have to specify the exact version in test.yml?  Or does 3.10 work?

geekygirlsarah:
> Do we have to specify the exact version in test.yml? Or does 3.10 work?

via https://docs.cloudfoundry.org/buildpacks/python/index.html :
> To request the latest Python version in a patch line, replace the patch version with x: `3.6.x`. To request the latest version in a minor line, replace the minor version: `3.x`.

So looks like we can. I'll update.

geekygirlsarah:
NO! It broke again! 
```
   -----> Python Buildpack version 1.7.53
   **WARNING** buildpack version changed from 1.7.49 to 1.7.53
   -----> Supplying Python
   **ERROR** Could not install python: no match found for 3.10 in [3.7.12 3.7.13 3.8.12 3.8.13 3.9.11 3.9.12 3.10.3 3.10.4]
```
Looks like it has to be a specific version :(

PR #207, opened 2022-05-13
Create 0004-P-migration-plan.md

petermarks-gsa:
Documented this conversation from Matt Jadud which I think will be helpful to clarify with the team

### DIFF
--- a/docs/product/decisions/0004-P-migration-plan.md
+++ b/docs/product/decisions/0004-P-migration-plan.md
@@ -1,0 +1,32 @@
+# 4. FAC Migration Plan
+
+Introduced: 20220513
+
+## Status
+
+Proposed
+
+## Context
+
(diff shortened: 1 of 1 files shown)


geekygirlsarah (approved):
This looks good to merge to me (though not sure if I'm the approval you want or not)

jadudm:
@geekygirlsarah, I think we can / should all read ADRs, because we might all find/see things that need fixed. So, the "+1" is always good on the ADR pathway.

I don't *yet* know for certain if the CFAC (Census FAC) will accept FY22 data. So, as an ADR, we are indicating what we hope/want, but we don't actually have an assurance there yet.

As an ADR in May '22, modulo waiting for a response from Census, this looks good to me. If our world changes in surprising or substantial ways, we can revisit the ADR.

PR #229, opened 2022-05-19
Modify SingleAuditChecklist to match Figma mockup

### LINKED ISSUE #216: Back End - SingleAuditChecklist model updates to store SF-SAC General Information section data
Breakout task for https://github.com/GSA-TTS/FAC/issues/185

**Background**
The current `SingleAuditChecklist` model contains fields covering Auditee and Auditor Information. We need to review the current implementation to ensure we're able to receive and store all fields present in the SF-SAC form and our current mock-ups.

- [ ] Enumerate form fields present on our [current mock-up of the General Info page](https://www.figma.com/file/ty7iZnrnpBO5OlhRoDQd4m/FAC-Playground?node-id=1899%3A8010)
- [ ] Review existing, update where necessary, and create new [field definitions](https://docs.djangoproject.com/en/4.0/ref/models/fields/) within  the `SingleAuditChecklist` model so that we can store all data for each of the fields identified above

geekygirlsarah:
Fixes #216 

This correctly matches up the `SingleAuditChecklist` model to match the [Figma mockup](https://www.figma.com/file/ty7iZnrnpBO5OlhRoDQd4m/FAC-Playground?node-id=1899%3A8010).

Changes include (but not limited to):
* Removed a duplicate `auditee_name`
* Removed `address2` from all addresses
* Added `auditor_country`
* ~~Renamed `uei` to `auditee_uei` to be more specific~~
* Rearranged items to be in the order they're asked on the mockup
* Renamed `firm_name` to `auditor_firm_name` to distinguish between contact and organization name


geekygirlsarah:
Though maybe I didn't want to rename `uei` to `auditee_uei`? It broke a _lot_ of things...

geekygirlsarah:
I think we may want to still consider a rename of `uei` to `auditee_uei` in case there's other UEIs involved, but also to match the naming of other auditee-related fields

backend/audit/models.py
```diff
@@ -82,39 +66,55 @@ class SingleAuditChecklist(models.Model):
     multiple_eins_covered = models.BooleanField(
         verbose_name=_("Multiple EINs covered"), null=True
     )
+    uei = models.CharField(
```
NicholasKramer, line 69: Changing to auditee_uei would be ideal, if we can make it work.

geekygirlsarah: It means changing up some tests and the API, which I can do. Just references to "UEI" are plentiful, so I want to change the model, not the API endpoint functionality (just the API backend that reads the model)

NicholasKramer (approved):

PR #265, opened 2022-06-10
Product Roadmap Revisions

petermarks-gsa:
From follow up discussion with @MelissaBraxton and @tadhg-ohiggins to break up MVP features into themed and sequenced milestones.

### DIFF
--- a/docs/agile-process.md
+++ b/docs/agile-process.md
@@ -1,9 +1,9 @@
 Agile Process
 ===================
 
-# Overview
+### Overview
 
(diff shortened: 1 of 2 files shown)


jadudm:
It might be good to note in the revised roadmap (which, to be clear, looks 👨‍🍳 💋 , because GH does not have a chef's kiss emoji) that we need features from the "full" release in order to be a functioning Clearinghouse, and that the MVP should not be construed as "version 1" or a system that meets our full obligations as the Clearinghouse.

Apologies for not suggesting an edit directly...

petermarks-gsa:
> 

Sounds good, @jadudm! Added this disclaimer about that: https://github.com/GSA-TTS/FAC/pull/265/commits/c26b7122bbb28a284d78019974637238ae0bcfe5

tadhg-ohiggins (approved):

Issue #249, opened 2022-05-31
Look into need for revised submissions in V1


petermarks-gsa:
we are talking about the SAC revision process, not the form data. If someone has already submitted a SAC and was approved, but something needs to be changed, what happens? What are backend considerations.

petermarks-gsa:
I have a proposal for handle revisions at the roadmap level in this pull request that I've assigned @MelissaBraxton to review: https://github.com/GSA-TTS/FAC/pull/269

PR #269, opened 2022-06-14
Product roadmap revisions from 6/13 census call

petermarks-gsa:
https://github.com/GSA-TTS/FAC/issues/249

### DIFF
--- a/docs/product/roadmap.md
+++ b/docs/product/roadmap.md
@@ -74,3 +74,4 @@ IMPORTANT NOTE: This system does not meet our full obligations as the Clearingho
 * Data import from old FAC
 * Login for government users (via Max.gov)
 * Not handling changes to form over time/in future (it’s likely that in some future year the questions on * the form will change, and we’re not trying to design for that for MVP)
+* Submitting revisions to an audit package


docs/product/roadmap.md
```diff
@@ -56,7 +57,6 @@ IMPORTANT NOTE: This system does not meet our full obligations as the Clearingho
     * Very basic, like year in excel match year submitted? UEI’s match?
 * Non-xls input to audit packages
 * Copying over FAQ’s and other pages
-* Submitting revisions to an audit package
```
MelissaBraxton, line 59: We'll want to make sure we reflect the need for this down the road, but maybe not for MVP.

MelissaBraxton (changes requested):
Looks good, but let's also make sure we copy the revision need as part of a post MVP phase.

docs/product/roadmap.md
petermarks-gsa: My impression from yesterday's [meeting](https://docs.google.com/document/d/1YP2K-WD1Uw1D8RvMYvxj3w57DV2uPyG4guqWu7AumLo/edit#) is that auditee/user-generated revisions is for sure not MVP given how tough, imperfect and unused solutions in this space seem to be. When we asked Mark directly what the cost of no revisions is, he responded that "For the user generated ones, Id imagine a lot aren’t material[y harmful]". I'm not sure those kind of revisions belong in the "Full FAC functionality release" either. This seems more like a wish list item we could potentially track its own section. 

That said, it sounds like we need to keep invalid resubmission flow on the roadmap and also provide a resubmission workaround for user-generated revisions. I've added that to milestone 3 as it seems to go along with the status and user business logic theme.

jadudm: Something Mark said in passing (I used this as an example, not realizing it was part of a current discussion), and he pointed out that once it is submitted, it starts its journey into the backend, and there's... *process* that they then invoke? Like, they message all the cognizant agencies when a grantee/auditor says they made a material error, and need to pull it back and revise?

Only adding this to the convo here to signal that it sounds like there's complexity (but, also ack'ing that y'all may have already uncovered that complexity in your research/convos).

MelissaBraxton (approved):
LGTM!

PR #282, opened 2022-06-27
Backend Login.gov integration

### LINKED ISSUE #135: As a user I can authenticate with Login.gov
**Background**
We're using Login.gov to authenticate

**Stories**

1. As an Auditee or Auditor I know how and where I need to authenticate so that I may submit new and modify existing single audit packages.

**Acceptance Criteria (We'll know we're done when..._**
- [ ] When a user can authenticate with login.gov and be associated with a Django user account
- [ ] For a user who has authenticated with login.gov and does not yet have a Django user account, the application creates one
  
**Tasks**
- [x] https://github.com/GSA-TTS/FAC/issues/259
- [ ] https://github.com/GSA-TTS/FAC/issues/267
- [x] https://github.com/GSA-TTS/FAC/issues/268
- [x] https://github.com/GSA-TTS/FAC/issues/308
- [x] Make contact with login.gov partnerships team

**QA**
- [ ] Meets acceptance criteria
- [ ] Screen reader - Listen to the experience with a screen reader extension, ensure the information presented in order
- [ ] Keyboard navigation - Run through acceptance criteria with keyboard tabs, ensure it works. 
- [ ] Text scaling - Adjust viewport to 1280 pixels wide and zoom to 200%, ensure everything renders as expected. Document 400% zoom issues with USWDS if appropriate.

timoballard:
Partially resolves #135

Backend portion of Login.gov integration. Authentication flow requires the JWT from Login.gov to be exchanged for a DRF token from our backend. All subsequent API calls are authenticated using the DRF token:

`---> POST /api/auth/token Authorization: Bearer <Login.gov JWT>`
`<--- { token: <DRF Token> }`
`...`
`---> /api/... Authorization: Token <DRF Token>`

- Adds `JWTUpsertAuthentication`, which upserts users who present a Login.gov JWT (extends `JWTAuthentication` backend from [djangorestframework-simplejwt](https://github.com/jazzband/djangorestframework-simplejwt))
- Adds `ExpiringTokenAuthentication`, which extends DRF's builtin `TokenAuthentication` so that tokens have a configurable TTL
- Adds view (`/api/auth/token`) for exchanging a Login.gov JWT for a DRF token


backend/users/models.py
```diff
@@ -4,7 +4,11 @@
 User = get_user_model()
 
 
-# Create your models here.
+class LoginGovUser(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
```
geekygirlsarah, line 8: Is a cascade what we want? If we delete a user, that makes sense it should delete the Login.gov info, but is there a case where we want to delete users? Would SACs be unaccessible after this? (Maybe this doesn't matter after they're certified and done through all the phases?)  I ask as maybe this should be `models.PROTECT` instead, but not sure I can answer that.

backend/users/models.py
```diff
@@ -4,7 +4,11 @@
...
+    login_id = models.CharField(max_length=100, unique=True)
```
geekygirlsarah, line 9: [NB] is the ID a 100-char field? I'd imagine this ID from login.gov is a standard size but figure maybe this should be the actual length of whatever it is. (if it's 100, then great!)

backend/users/models.py
```diff
@@ -4,7 +4,11 @@
...
+
+
 class UserProfile(models.Model):
     user = models.OneToOneField(User, related_name="profile", on_delete=models.CASCADE)
```
geekygirlsarah, line 13: [NB] Same here about cascading as above for line 8

geekygirlsarah (approved):
I have a few questions, but I don't know that they have to be answered to delay my approval of this.

backend/users/models.py
timoballard: Yeah, login.gov IDs are UUIDs, so 36 characters long. My tendency in these scenarios, where we're referencing an external field that we don't have control over, is to pad the length a bit in case something about the ID format changes on their end, or if out initial assumptions were somehow incorrect. Thoughts?

backend/users/models.py
timoballard: Same reply as above

backend/users/models.py
geekygirlsarah: yeah, I get that, though I also think if you can fit nearly 3 UUIDs in one field, it's probably too big. Also if you specify a size, the DB can allocate that space, which can get inefficient over time

timoballard: Yep that's fair - cut it in half to a max of 50

backend/users/models.py
geekygirlsarah: Then this seems fine as it is!

PR #327, opened 2022-07-07
Allow submissions without UEI

### LINKED ISSUE #286: 🅢 Back end - Create new submission without valid UEI
Breakout of https://github.com/GSA-TTS/FAC/issues/278

**Acceptance criteria:**

* We can accept null or blank UEI values existing endpoint

timoballard:
Resolves #286 

Allows SAC submission to proceed without an auditee UEI.

Adds tests to make explicit the current required field behavior on the `/auditee` endpoint. Looking for feedback specifically on `auditee_name`, which can be `null` or missing, but cannot be blank. Is this desired?

- `auditee_uei`
  - `null`: allowed
  - `blank`: allowed
  - `missing`: allowed
- `auditee_name`
  - `null`: allowed
  - `blank`: not allowed
  - `missing`: allowed
- `auditee_fiscal_period_start/end`
  - `null`: not allowed
  - `blank`: not allowed
  - `missing`: not allowed


tadhg-ohiggins:
My understanding is that `auditee_name` (and the fiscal period fields) have to allow absent/null values because of the possibility of submitting a UEI that SAM.gov recognizes but returns no data for, so this looks correct to me.

Is the [0004_alter_logingovuser_login_id.py](https://github.com/GSA-TTS/FAC/pull/327/files#diff-54e9bcbdca96376c90f131a86aa7345533ce6221c1872fb24eed6a6ec5b20b63) inclusion stray or is there a relation to the `logingovuser` model that I'm missing?

tadhg-ohiggins (approved):

PR #347, opened 2022-07-12
Tweak Makefile so that coverage is invoked by python executable

tadhg-ohiggins:
Avoid confusion
By being more explicit.
This should help with that.

-----

I ran into a situation locally where the `coverage` executable was associated with a different version of Python, and this should fix that problem.


timoballard (approved):
LGTM

Issue #320, opened 2022-07-06
🅜 Change login.gov flow to look for users with matching email addresses


tadhg-ohiggins:
Breakout issue of #246

### Background

When users log in with login.gov and do not already have associated `User` objects, we need to check against the list of users that have been created without any non-email data. (These are the users that have been added to their SAC submissions by other users.)

The current login.gov flow only looks at primary email addresses supplied by login.gov, and we need to change that to look at all email addresses associated with the user. (Both primary and a set of other email addresses are provided in the user info passed along by login.gov.)

### Acceptance Criteria (We'll know we're done when...)

- [ ] We can add a user to a SAC when that user doesn't exist in the system and then log in using the login.gov flow and verify that that user has been added to the system.
- [ ] And that that user can edit that SAC.

### Tasks

- [ ] Check assumption that email address can only exist in one login.gov account at a time.
- [ ]

petermarks-gsa:
@timoballard I just added this to the board. Can you please advise on t-shirt size of this?

timoballard:
@petermarks-gsa M

tadhg-ohiggins:
## 2022-07-08 discussion on some edge cases around login.gov users and `Access` objects.
Participants: @geekygirlsarah @tadhg-ohiggins @timoballard 

1.  `SingleAuditChecklist` object is created with `Access` objects that have null `user_id` fields but they have the email addresses the user submitted for auditee, etc.
1.  We do not check the email addresses in the newly-created `Access` objects to see if there are `User` objects with matching email addresses—we only check for that during the login.gov login flow.
1.  When a user logs in with login.gov, we:
    1.  See if there is already a `User` for that login.gov identifier. If not, we create that `User` with their login.gov primary email address as their email address property on `User`.
    1.  Look at all their login.gov email addresses and at all `Access` objects with `null` `user_id` fields to see if any of the email addresses match.
    1.  If any match, we associate those `Access` objects with their `user_id`.

## Scenario 1

*   Alice adds bob@agency1.gov to SAC 1. `Access` object is created with bob@agency1.gov as email address.
*   Chris adds fac-liason@agency2.gov to SAC 2. `Access` object is created with fac-liaison@agency2.gov as email address.
*   Bob logs into FAC for the first time, `all_emails` contains `[bob@agency1.gov, fac-liaison@agency2.gov]`
*   We create `User` for Bob with `bobprimaryemail@agency.gov` as the canonical email address.
[...]

PR #348, opened 2022-07-13
Claim pending audit access during Login flow

timoballard:
Addresses #320 

Updates the backend Login.gov authentication flow so that:
- If a matching `User` record already exists, but no `LoginGovUser` record exists, a `LoginGovUser` record is created and associated with the existing `User` record
- Existing `Users` are looked up by finding all records where the email address is found in the LoginGov `all_emails` collection
  - If a single matching `User` is found, use it
  - If no matching `User` is found, create one
  - If multiple matching `Users` are found, use the one that matches the LoginGov primary email address
    - If none of them match the LoginGov primary email address, use the first one
- Find any outstanding `Access` records associated with an email address found in the `all_emails` collection
  - Link the `Access` record to the `User` ID


rachelschneiderman (approved):

backend/users/auth.py
```diff
@@ -10,17 +10,59 @@
...
 
 class JWTUpsertAuthentication(JWTAuthentication):
+    def get_or_create_auth_user(self, email, all_emails):
+        # find all Users where email is in LoginGov all_emails
+        users = self.user_model.objects.filter(email__in=all_emails)
+
+        if users.count() > 1:
+            # more than one found, take the one that matches the LoginGov primary email
+            try:
+                user = users.get(email=email)
```
tadhg-ohiggins, line 39: Is it possible that there might be more than one user with the same email address?

backend/users/auth.py
```diff
@@ -10,17 +10,59 @@
...
+                # none of the Users use the LoginGov primary email, take the first one
+                user = users[0]
+        elif users.count() == 1:
+            # found a single User match
+            user = users[0]
+        else:
+            # didn't find an existing User, create one
+
+            # generate a random password
+            password = "".join(
```
tadhg-ohiggins, line 50: Do we want to note here that this password is never going to be used, since our auth is all via login.gov?

timoballard: yes - done!

backend/users/auth.py
timoballard: per our Slack discussion: uniqueness of `User` emails is not enforced at the database level, but this scenario should not be possible at the application level, given the way new `Users` are created. It is possible, though, if `Users` are created manually, either via direct database insertions or through Django admin. So, we explicitly catch and log that failure mode and ultimately reject the authentication attempt

tadhg-ohiggins (approved):

tadhg-ohiggins:
Looks good to me, nice work!

PR #362, opened 2022-07-14
PDR - Determine certify audit submission path

petermarks-gsa:
Per engineering discussion today, we're ready to decide on the "Special roles/UX in the app to sign digitally" path that design and product recommended as the preferred solution. Creating this PDR for formalize that product decision and explain the thinking since we, as a team, have been going back and forth on this.


MelissaBraxton (approved):
Awesome to hear! Thanks for this update!

PR #364, opened 2022-07-15
Add note on resolving .python-version conflicts via symlink

tadhg-ohiggins:
What is a version?
Tools have their own ideas.
Fix it with symlinks.

-----

`pyenv` wants explicit patch versions, like `3.10.0`, but our deployment process wants only minor versions, like `3.10`, so that any available patch version can be used (and so we don't have to keep track of which patch versions drop out of being supported). We don't care about which patch version is being used, but need to appease `pyenv` in some local development situations; this adds a note to the docs on how to do that.

### DIFF
--- a/docs/product/roadmap.md
+++ b/docs/product/roadmap.md
@@ -1,7 +1,7 @@
 Product Roadmap
 ===================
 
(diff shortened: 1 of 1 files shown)


timoballard (approved):

rachelschneiderman (approved):

geekygirlsarah (approved):
👍 


To clarify the description a tiny bit, GitHub Actions and/or cloud.gov need either a specific patch version (`3.10.5`) or a general minor version (`3.10`), otherwise it'll break and not deploy due to missing buildpacks. (I fought this with #205 and #208 back in May after Cloud.gov deprecated an old patch version.)

Issue #379, opened 2022-07-19
🅢 Backend implementation of logging out


timoballard:
Breakout issue of #359 

### Background
When a user logs out from the FAC-Frontend, the backend should be notified so that it can delete the auth token associated with that user. This will prevent the token from being reused and force the user to re-authenticate with their Login.gov token before continuing to use the rest of the API.

### Acceptance Criteria (We'll know we're done when...)

- [ ] An authenticated `DELETE` request can be sent to `/auth/token`
- [ ] After sending a `DELETE` request to `/auth/token`, the previously issued FAC token can no longer be used for API requests

### Tasks

- [ ] 
- [ ]

petermarks-gsa:
@timoballard I see https://github.com/GSA-TTS/FAC/pull/393 got merged. Are we good to close this?

timoballard:
@petermarks-gsa yep!

PR #393, opened 2022-07-25
Add Token delete endpoint

timoballard:
Addresses #379 

Adds a `DELETE` endpoint at `/api/auth/token`, which deletes the `Token` associated with the authenticated user, effectively logging them out of the backend. Before making additional API calls, a JWT from Login.gov would need to be exchanged for a new API token.

### DIFF
--- a/backend/users/test_views.py
+++ b/backend/users/test_views.py
@@ -23,7 +23,7 @@ class AuthTokenViewTests(TestCase):
         """
         view = AuthToken()
 
-        # AuthToken views should allow JWTUpsertAuthentication 
+        # AuthToken views should allow JWTUpsertAuthentication
         self.assertIn(JWTUpsertAuthentication, view.authentication_classes)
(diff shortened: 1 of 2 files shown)


backend/users/test_views.py
```diff
@@ -57,3 +67,21 @@ def test_existing_token(self):
         self.assertEqual(response.status_code, 200)
         self.assertIn("token", data)
         self.assertNotEqual(token.key, data["token"])
```
tadhg-ohiggins, line 69: Maybe a little much, but would adding

```
tokens = Token.objects.filter(user=user)
self.assertEqual(tokens.count(), 1)
```

or something similar here make sense?

timoballard: makes sense to me

geekygirlsarah (approved):
👍  from me

tadhg-ohiggins (approved):

PR #397, opened 2022-07-26
391: Endpoint for retrieving SACs submitted by current user

rachelschneiderman:
Step 1 of populating the 'My audit submissions' table. This PR creates the endpoint for populating the table, for now only return SACs that have been submitted by the current user.

### DIFF
--- a/backend/api/views.py
+++ b/backend/api/views.py
@@ -193,22 +193,6 @@ class SubmissionsView(APIView):
     def get(self, request):
         current_user = request.user
 
-        # Report ID, Status, Auditee UEI, Fiscal Year End Date,Auditee Name
-        # all_submissions = SingleAuditChecklist.objects.filter(submitted_by=current_user)
-        #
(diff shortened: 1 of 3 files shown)


backend/api/views.py
```diff
@@ -181,3 +182,24 @@ def post(self, request):
...
+            submitted_by=current_user
+        ).values(
+            "report_id",
+            "submission_status",
+            "auditee_uei",
+            "auditee_fiscal_period_end",
+            "auditee_name",
+        )
+
+        return JsonResponse(list(all_submissions), safe=False)
```
geekygirlsarah, line 205: [NB] Sweet, you figured out how to do this right!

geekygirlsarah (approved):
👍  from me!

backend/api/views.py
rachelschneiderman: Only with your help!!! 

teamwork

Issue #344, opened 2022-07-12
PDR/ADR - File Upload storage


petermarks-gsa:
We need to come up with a solution to store Uploaded PDFs.

timoballard:
@petermarks-gsa @tadhg-ohiggins 

As a conversation starter: the [Civil Rights Portal](https://github.com/usdoj-crt/crt-portal) for USDOJ has the same backend stack that we do (Django backend on Cloud.gov). We built a file upload/storage solution there which includes validations around file size, supported file types, and virus scanning. Files are stored in S3 (via [cloud.gov's managed services](https://cloud.gov/docs/services/s3/)).

Seems like at a minimum, we could use this as a reference model for our solution

jadudm:
Part of the answer to this might be informed by some conversation with Federal users. I don't *think* it will radically shift storage choices, but knowing how people want/need to access the data may help inform the design. In particular, I'm thinking about questions around persistent links, bulk access, and so on that have surfaced in piecemeal ways over the past few months... I don't know if we have any dedicated UX conversations around this?

PR #405, opened 2022-08-01
ADR - File uploads

timoballard:
Address #344


jasnakai (approved):
LTGM.

petermarks-gsa:
@timoballard Do I understand correctly that the solution you were a part of on the DOJ application made it to production? Do you recall if any of the technology choices needed special consideration for ATO or was it pretty smooth sailing?

timoballard:
@petermarks-gsa Correct, this solution is in production for the DOJ-CRT app and didn't create ATO complications (although to be fair, in the case of DOJ-CRT, the application had already been ATO'd and in production for several months before this feature was added).

This technical stack should be ATO-friendly, as S3 is both FedRAMPed and is offered as a managed service through cloud.gov. ClamAV has been used for similar purposes on a number of other past TTS projects, so if need be we can reach out to colleagues involved in that work to learn from their ATO experiences in this area.

petermarks-gsa (approved):
@timoballard Thanks for context. Adding my ✅

jasnakai (approved):
Added frontend considerations.

petermarks-gsa:
@tadhg-ohiggins Are we good to merge and call this approved?

tadhg-ohiggins (approved):

Issue #376, opened 2022-07-19
Sam.gov connection errors


petermarks-gsa:

tadhg-ohiggins:
*   We don't know what the cause of the errors (403s?) is yet.
*   It's possible our existing key has expired.
*   We may have access to another key already, and should try that.
*   If not, we may have to apply for another key; until we have an ATO SAM.gov will only provide keys tied to individuals (not to applicatios) which may have shorter lifespans.
*   We likely need another issue for “track SAM.gov key lifetime and build process for renewing key”.

heymatthenry:
This is what I'm seeing on the front end: 

`"valid":false,"errors":{"auditee_uei":["SAM.gov API response status code invalid: 403"]}}`

petermarks-gsa:
@tadhg-ohiggins Who should we assign this blocking issue to?

tadhg-ohiggins:
Switching the API key via the cloud.gov dashboard (but not via the GitHub secrets plus deploy.yml process) worked. Closing.

PR #419, opened 2022-08-04
Change SamGov's key to be set by UPSI for better security

### LINKED ISSUE #384: Figure out why GitHub action to deploy SAM.gov API key didn't work
### Background

#376 revealed that the GitHub deployment action didn't do exactly what we expected. Figure out why that is.

- [ ] The deployment action does what we expected, OR
- [ ] We have an alternative mechanism for changing the API keys.

geekygirlsarah:
Fixes #384 

Problem:
#376 revealed that the GitHub actions didn't update `SAM_API_KEY` env variable upon deploy.

Cause:
The GitHub action [18f/cg-deploy-action](https://github.com/18f/cg-deploy-action) used to deploy had Sam's key passed in as an environment variable, which is what [`cf push`](https://cli.cloudfoundry.org/en-US/v6/push.html) implies is allowed. However, `--var` only replaces variables in a manifest file, not env vars on the CF app. So nothing was happening.

Fix:
Upon research, Cloud Foundry says that [User-Provided Service Instances](https://docs.cloudfoundry.org/devguide/services/user-provided.html) (or UPSI) should be used to deliver credentials instead of environment variables as env vars can be put into logs by accident. This fix involves me creating one (called `fac-key-service`), and modifying the GitHub action to update credentials in the UPSI instead of the environment variable. Upon push, the app should restage and access the updated key.

In code, I updated `settings.py` to pull the key from the UPSI instead of the environment variable with a simple package called `cf-env` [https://docs.cloudfoundry.org/buildpacks/python/index.html](recommended by Cloud Foundry). It connects to the service and pulls out credentials.

Any future API keys or other secure strings needed to be stored should be put in here and can be easily looked up with these changes.

Note:
This should work. In that case, the environment variable will need to be manually deleted on the cloud.gov app.

It's entirely possible this could fail after merging. Restoring the old settings file should fix this as the environment variable will still exist on cloud.gov.


timoballard (approved):
LGTM - added #420 to capture our (eventual) need for a production GH deployment workflow

geekygirlsarah:
> LGTM - added #420 to capture our (eventual) need for a production GH deployment workflow

good catch!

PR #428, opened 2022-08-09
Add endpoint to return all accessible audits for a user

### LINKED ISSUE #410: Ⓛ Backend for all SACs associated with current user
Breakout issue of [#343](https://github.com/GSA-TTS/FAC/issues/343)

### Background

The “My audit submissions” page should show all submissions for a user—not merely all submissions *by* a user. We need an endpoint for that.

*   Since we're filtering by user by definition:
    *   We need to handle auth.
    *   We need to get the user from the login/auth machinery.
    *   We need to get the submissions for that user, which will be looking for Access objects that have that user associated with them, and then looking for the SAC objects that have that user in any of these fields:
        *   `certifying_auditee_contact`
        *   `certifying_auditor_contact`
        *   `auditee_contacts`
        *   `auditor_contacts`
        * `creator`

        and also looking for submissions submitted by that user.

        OR we need to change the model so that many Access instances can point to a single SAC (but it's one-to-many and not many-to-many because each Access instance itself can only point to one SAC instance, and SAC instances wouldn't point to Access instances at all, only the reverse). This is broken out in #418.

        Discussion between @tadhg-ohiggins, @timoballard, and @geekygirlsarah resulted in broad consensus that it's better to rewrite the models.
*   Filtered by fiscal year of submission (maybe—probably future enhancement)
*   No pagination.
*   No sorting.
*   The response should include the list of audits associated with the user and the role the user has for that audit.
     *    What do we do if a user has multiple roles on the same audit? Ideally we would return the user's roles as an array but this is an implementation detail to be worked out.
*   Create tests:
    *   Create a new user, hit endpoint, see no results.
    *   Create a new user, create an SAC (with Access for the user), see one result.
    *   Create a new user, create *n* SACs (each with Access for the user), see *n* results.
    *   Create a new user, create *n* SACs (each with Access with different roles for the user), see *n* results, each reflecting the correct role for a given SAC.
    *   Create a new user, create 1 SAC and *n* different Accesses with different roles for the user, see *n* results, one for each role assigned to the user for the SAC.
    *   Go through creation cycle, delete an SAC instance, verify that it's no longer listed.
    *   Go through creation cycle, delete an Access instance, verify that the corresponding SAC is no longer listed.
    *   … which means that to do this properly we need to have an endpoint for removing access. (Related to https://github.com/GSA-TTS/FAC/issues/333)

- [x]   We can hit the endpoint with an authenticated user and see their submissions.
- [x]   We can add and remove access for that user to specific SACs and see this reflected in the results from the endpoint.
- [x]   We have tests for the above.

### Tasks

- [x]   #418.
- [x]   Decision on what role the user who created the SAC is supposed to have. (decision: `creator`)
- [x]   Since that decision may take a while, in the interim, add a new role with a label of `Creator` and a value of `creator` and use that during the rewrite in the next step.
- [x]   Rewrite the `SingleAuditChecklist` and `Access` models, and the associated code/tests, so that we use the many-`Access` -> one-`SingleAuditChecklist` approach for associating users with SACs.
- [x]   Write the endpoint that returns all SACs associated with the current user.

timoballard:
Resolves #410

Adds `/access-list` endpoint, which returns all of the `SingleAuditChecklists` for which the authenticated `User` has `Access` (see return format below). If a `User` holds multiple roles for a `SingleAuditChecklist`, that `SingleAuditChecklist` will appear in the list once for each role the user holds. 

```
[
    {
        "auditee_uei": "ABC123DEF456",
        "auditee_fiscal_period_end": "2022-08-01",
        "auditee_name": "a new one",
        "role": "auditee_contact",
        "report_id": "2021XC70001000003",
        "submission_status": "in_progress"
    },
    {
        "auditee_uei": "ABC123DEF456",
        "auditee_fiscal_period_end": "2022-08-01",
        "auditee_name": "a new one",
        "role": "auditee_cert",
        "report_id": "2021XC70001000003",
        "submission_status": "in_progress"
    },
    {
        "auditee_uei": "ABC123DEF456",
        "auditee_fiscal_period_end": "2022-08-01",
        "auditee_name": "somethingelse",
        "role": "creator",
        "report_id": "2021E060001000002",
        "submission_status": "in_progress"
    }
]
```

### DIFF
--- a/backend/api/serializers.py
+++ b/backend/api/serializers.py
@@ -148,4 +148,11 @@ class AccessListSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Access
-        fields = ["auditee_uei", "auditee_fiscal_period_end", "auditee_name", "role", "report_id", "submission_status"]
+        fields = [
+            "auditee_uei",
+            "auditee_fiscal_period_end",
+            "auditee_name",
+            "role",
+            "report_id",
+            "submission_status",
+        ]
--- a/backend/api/serializers.py
+++ b/backend/api/serializers.py
@@ -122,3 +122,30 @@ class SingleAuditChecklistSerializer(serializers.ModelSerializer):
     class Meta:
         model = SingleAuditChecklist
         fields = "__all__"
+
+
+class AccessListSerializer(serializers.ModelSerializer):
+    auditee_uei = serializers.SerializerMethodField()
+    auditee_fiscal_period_end = serializers.SerializerMethodField()
+    auditee_name = serializers.SerializerMethodField()
+    report_id = serializers.SerializerMethodField()
+    submission_status = serializers.SerializerMethodField()
+
+    def get_auditee_uei(self, access):
+        return access.sac.auditee_uei
+
+    def get_auditee_fiscal_period_end(self, access):
+        return access.sac.auditee_fiscal_period_end
+
+    def get_auditee_name(self, access):
+        return access.sac.auditee_name
+
+    def get_report_id(self, access):
+        return access.sac.report_id
+
+    def get_submission_status(self, access):
+        return access.sac.submission_status
+
+    class Meta:
+        model = Access
+        fields = ["auditee_uei", "auditee_fiscal_period_end", "auditee_name", "role", "report_id", "submission_status"]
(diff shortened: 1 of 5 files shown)


tadhg-ohiggins (approved):
Nice work!

PR #434, opened 2022-08-11
First draft of policy readme

petermarks-gsa:
Per discussion with the Grants QSMO and others on today's OMB call, we think it's a good idea to track policy decision in this repo and communicate to government stakeholders that we are tracking changes. I've added a README with a few examples of policies to decide on after the transition.


jadudm (approved):
Looks good to me.

PR #437, opened 2022-08-12
Handle field-level SAC validation errors

tadhg-ohiggins:
Validation rules
Are there for some good reasons
We should explain them.

-----

Add tests that enter data which should be rejected by our field-level validations rules for `SingleAuditChecklist`, and ensure that we reject them and provide explanatory errors.

Also add view logic to catch and handle those errors.

### DIFF
--- a/backend/api/test_views.py
+++ b/backend/api/test_views.py
@@ -554,6 +554,102 @@ class SingleAuditChecklistViewTests(TestCase):
 
         self.assertEqual(response.status_code, 404)
 
+    def test_put_bad_data(self):
+        """
+        If we have edit access and we're submitting to the allowed fields but
+        we're submitting bad data, we should get errors.
+        """
+
+        def check_response(bad_data, expected):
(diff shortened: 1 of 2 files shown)


backend/api/test_views.py
```diff
@@ -554,6 +554,102 @@ def test_put_bad_report_id(self):
...
+        for key in ternary_keys:
+            message = "“invalid_boolean” value must be either True, False, or None."
+            expected = {"errors": {key: [message]}}
+            with self.subTest():
+                check_response({key: "invalid_boolean"}, expected)
+
+        for key in length_100_keys:
+            one = "a value over one hundred characters long is annoying to enter "
+            two = "succinctly when your Python line length limit is eighty-eight"
+            value = f"{one}{two}"
```
geekygirlsarah, line 643: [NB] ha!

geekygirlsarah (approved):

PR #442, opened 2022-08-23
Add ADR for Excel generation and validation

### LINKED ISSUE #328: 🅛 PDR/ADR - Excel file generation
## Background
The process for submitting entries to the FAC relies on uploading audit data via Excel templates. Design has been clear that this is so critical to workflows that it's a priority over web forms and could actually be done in place of them for MVP. There is already an accepted PDR for ["how to handle bulk data uploads"](https://github.com/GSA-TTS/FAC/blob/main/docs/architecture/decisions/0008-bulk-data-uploads.md) which uncovered an Excel File validator as well as a corresponding [spike](https://github.com/GSA-TTS/FAC-xlsx-validator) in that effort. The part that we don't have clarity around, which is the subject of this ticket, is creating Excel file templates.

## Understanding of generated templates
Excel template forms downloaded from the Census FAC are custom generated by the server to validate against previous data entered, including:

* **Validating EIN** - Currently, EIN's are baked into the template and validated when they are uploaded. Per Census, this is helpful for auditors doing a bunch of audits at once to ensure they don't upload a sheet for another organization they're uploading an audit for. 
* **Previously entered data options** - We understand there are cases where data entered on previous steps is surfaced as options in the template on subsequent steps. 

## Maintainability

The Excel process under discussion would have at least three levels of validation:

1.  The backend validation rules: any `SAC` submitted, regardless of how it is submitted, must pass backend validation.
2.  The frontend validation rules: the frontend libraries being used to translate Excel into JSON for submission need to validate that JSON and flag errors before anything is submitted to the backend.
3.  The Excel validation rules: the Excel files need to contain validation rules so that users know within Excel that their entries are valid or have errors.

These constitute the base level of “dynamic generation”—not generation based on user data, but generation based on whatever the validation rules are, as these validation rules themselves are likely to change (for example, the rules include lists of valid answers to specific questions, and those lists change).

Any system we come up with for this must have a single source of truth for all three levels. While it's possible we could use the Python backend code itself as this source, it's more likely that we would use some well-documented, well-understood system for defining rulesets and then use its tooling as part of translation into the various domains.

One possible approach to this is [JSON Schema](https://json-schema.org/), but many other options doubtless exist.

The base questions for maintainability would then be:
*   What rule-definition tool(s) would we use?
*   How difficult would it be to translate the expected rules (such as those contained in the existing Census systems and the existing Excel templates) into that system?
*   How difficult would it be to translate the rules from that system into our three levels?
*   How difficult would it be to add that translation approach to our build pipeline?

In addition, each of those three levels of validation requires automated testing. We have good understanding for how to approach that for the frontend and backend, but no idea how this would be accomplished for Excel.

## Potential solutions (Non-exhaustive list)
  * Option 1: Find a way to generate Excel file templates (current Census FAC approach, Design team recommendation)
    * It's unclear this is an option on our stack
  * Option 2: Static Excel file templates
    * Risk: Potentially more confusing for the user to re-enter information
    * Risk: CPAs working on multiple audits might upload the wrong sheet
  * Option 3: Single Excel file templates
    * Risk: Limited / no client side validation, which could lead to more audit churn
    * Opportunity: More simple experience for users
  * Other options: any way to do this on CSV? 

## Acceptance Criteria:
- [ ] We have identified a solution or set of solutions that meets the maintainability criteria laid out above.
- [ ] Compared a few libraries for excel file generation and considered security/compliance considerations. 
- [ ] Consider the complexity of current excel templates / validation when making a recommendation
- [ ] Estimate effort various approaches
- [ ] (as a team) Compare estimates with user needs and make a technology decision for MVP

## Tasks
- Research Census's current implementation of template generation
- Look into libraries for template generation and compare them

geekygirlsarah:
Resolves #328 

This adds the ADR about Excel files with research that @jasnakai and I did.


docs/architecture/decisions/0011-excel-generation-validation.md
```diff
@@ -0,0 +1,46 @@
...
+2. We need to determine how the system will save data on the backend. This will determine what kind of data we need to get from the Excel file and save to the backend. 
+3. Depending on how we decide to move forward regarding UEI status and associated data (received from Sam.gov, manually typed in, etc.) may have an impact on how the files are generated.
+4. Depending on how this is implemented along with the rest of the FAC workflow, it may require some user research to determine the best way to do it.
+5. We could reuse Census' old template, or we could make our own. If we make our own, it may also involve user research and/or policy revisions to make sure it's adequate for the needs of all users involved (like auditors, auditees, processors, etc.)
+6. Given the flow of processing Excel files on the frontend and sending the important data to the backend, Excel files won't ever get saved. This means the concerns of the File Upload ADR won't apply here.
```
jadudm, line 45: Are there policy/legal implications to not storing the original document as a source of truth? Can the team make that decision alone, or should it be a question to OMB/the steering committee?

geekygirlsarah: Maybe a good ? for our Census friends on Thursday's meeting.

geekygirlsarah: Maybe a good ? for our Census friends on Thursday's meeting. Unless @tadhg-ohiggins knows?

tadhg-ohiggins: I think it's a question for Census; I feel like they've mentioned it in the past but I can't recall whether they save, for instance, the last version of Excel submitted or every version, although I think the latter is unlikely.
Since they have full bidirectional translation I think the representation of the data in their system is the canonical submitted “document”, not any Excel file that might have been part of that process.

tadhg-ohiggins: In any case I do not think we should regard any Excel submissions as “documents” in a formal sense, just as we would not regard a curl `POST` body as a document in a formal sense. The “documents” are the contents of the SAC as represented in our system, and the contents of the submitted PDF. Anything else is ephemeral.

petermarks-gsa: Added the following question to tomorrow's [Census agenda](https://docs.google.com/document/d/1DxuyVOusBIIRrXCaOmmzQp-Q54bthuSirmwR4j4jTbc/edit#): 

What is the current process for storing user-uploaded excel templates? Do you currently store these?

petermarks-gsa:
Per comments at standup today, I think it would be helpful to review this alongside the (WIP) implementation story for this:

https://github.com/GSA-TTS/FAC/issues/443

Do we have the unknowns answered to confidently scope and move forward with this implementation issue?

tadhg-ohiggins:
At development huddle today we discussed this and agreed that the following need to be added to or made more prominent in this ADR:

*   The importance of a single source of truth for all validation.
*   A reference implementation of JSON Schema for this purpose to verify that it fits the use we're trying to put it to.
*   The possibility of needing to maintain two separate schemas, one for finalized submissions and one for in-progress submissions.
*   An outline of how the test framework/pipeline for this would work (that is, how do we make a change to the validation rules and verify that it's reflected in all of the consumers of the validation rules?).

@jasnakai and @geekygirlsarah  are going to add these elements. @jasnakai is taking point on the JSON Schema prototype implementation but we will all pitch in if that gets tricky.

docs/architecture/decisions/0011-excel-generation-validation.md
tadhg-ohiggins: Census stated that they do not store user-uploaded Excel fiiles.

tadhg-ohiggins (approved):

Issue #399, opened 2022-07-27
🅛 Implement form to view/edit SAC data


heymatthenry:
Frontend ticket to accompany #398

## Background

Users will have input some data into the pre-SAC forms, and other data may have come from the SAM.gov API. If the user was unable to provide a valid UEI during the pre-SAC flow, some fields may be empty.

## Technical overview

On load, the page will request data from the endpoint described in #398 and, if a field from that endpoint is editable, its value will appear in a form field. If a field is not-editable, it will appear as rendered, non-editable text on the page.

## Acceptance criteria

- [x] Frontend is able to `GET` data from the appropriate endpoint
- [x] Data returned from endpoint will render in either a form field (if editable) or as non-editable text (if not).
- [x] This form implements the same client-side validation as the SAC creation step 
- [x] Form is able to `POST` new data to the endpoint
- ~[ ] Server-side errors from the `POST` will be displayed to the user~

heymatthenry:
Going to suggest this is at least a `L` rather than an `M`. This ticket is probably where we'll start integrating whichever client-side JS component framework we end up choosing, and will be a template for how we do hydration going forward.

petermarks-gsa:
@heymatthenry Can you please update on progress here? Sounds like we might need to choose a new framework?

heymatthenry:
@petermarks-gsa Yes, we're going to add a new library. This isn't going to be a huge change and won't affect any existing code. It's something we can sprinkle in as needed going forward. I have an ADR in progress: #464

Draft pull request for the Frontend code is: GSA-TTS/FAC-Frontend#44

heymatthenry:
@tadhg-ohiggins & I were talking about the piece of this ticket that covers showing errors from the server, and we agreed that it ought to be its own ticket. I'll make the ticket but there are some things I want to nail down first since I think that work may apply to more than just the SAC form.

That said, the outstanding tasks for this ticket are to:

- [ ] Wire up radio buttons to reflect previously entered data
- [ ] When #468 lands, wire up the fields it touches
- [ ] Add e2e tests for the above

heymatthenry:
Broke out the error messaging piece into #484.

Some fields aren't yet being populated (particularly those whose data comes in from SAM.gov in step 2). Once those data make it into the JSON returned by `/sac/edit/:report_id`, they should ✌️just work✌️ and populate the relevant fields. Given that, this should be in good shape.

petermarks-gsa:
[...]

PR #468, opened 2022-09-12
399 backend support

tadhg-ohiggins:
We need all the things
Not only a mere subset.
Add all the info

-----

We were not returning all of the data entered during the pre-SAC steps from the `/sac/edit/[report_id]` endpoint. This adds them, and also makes some field name changes for consistency.

Should mean the backend supports what #399 needs.


heymatthenry (approved):
Needs more lambdas but otherwise LGTM

timoballard (approved):

geekygirlsarah (approved):
👍 but definitely needs more lambdas like Matt said

PR #471, opened 2022-09-14
Fix UEI validation not returning all the info it should

geekygirlsarah:
[Updated]

The address fields weren't propagating from SamGov to the saved SAC. This should resolve it.

Also added more robust tests that ensure individual fields are there and match as expected from SamGov (rather than "did I get a thing? I got a thing, tests pass")

Relates to [#468](https://github.com/GSA-TTS/FAC/pull/468), [#399](https://github.com/GSA-TTS/FAC/issues/399) and [FAC-Frontend/#44](https://github.com/GSA-TTS/FAC-Frontend/pull/44)


heymatthenry:
Relates to #468, #399 and GSA-TTS/FAC-Frontend#44

heymatthenry (dismissed):
This LGTM but I'll defer to our friends with more backend expertise.

backend/api/views.py
```diff
@@ -76,7 +76,7 @@ def post(self, request):
             return Response(
                 {
                     "valid": True,
-                    "response": json.loads(serializer.data.get("auditee_uei")),
+                    "response": json.dumps(serializer.data.get("auditee_uei")),
```
timoballard, line 79: Should we update (or create) tests around this to assert that the expected fields are coming back?

tadhg-ohiggins (approved):

PR #490, opened 2022-09-23
Add federal awards placeholder endpoint + tests

### LINKED ISSUE #478: Backend endpoint for data to generate Excel (placeholder)
Breakout issue of #443 

### Background

The FAC has to generate an Excel file for the user to download to fill in with their federal awards. This task is specifically for an API endpoint that the frontend will gather data for the Excel spreadsheet. (For now, this API endpoint won't return anything.)

Endpoint should be `/sac/edit/[report_id]/federal_awards`

- [ ] New endpoint is created
- [ ] Test to ensure it's able to be contacted and return an empty `{}`

### Tasks

- [ ] New endpoint is created
- [ ] Test to ensure it's able to be contacted and return an empty `{}`

geekygirlsarah:
Resolves #478 

This adds in the placeholder that will become the federal awards endpoint. For now, it just returns a `{}`.

### DIFF
--- a/backend/api/test_views.py
+++ b/backend/api/test_views.py
@@ -879,7 +879,29 @@ class SacFederalAwardsViewTests(TestCase):
         """Convenience method to get the path for a report_id)"""
         return reverse("sacfederalawards", kwargs={"report_id": self.sac_report_id})
 
-    def test_get_placeholder_data(self):
+    def test_get_authentication_required(self):
+        """
+        If a request is not authenticated, it should be rejected with a 401
+        """
+
+        # use a different client that doesn't authenticate
+        client = APIClient()
+        response = client.get(self.path(self.sac_report_id), format="json")
+        self.assertEqual(response.status_code, 401)
+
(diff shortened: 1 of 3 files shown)


backend/api/test_views.py
```diff
@@ -841,6 +841,55 @@ def test_edit_inappropriate_fields(self):
                 self.assertEqual(response.json(), expected)
 
 
+class SacFederalAwardsViewTests(TestCase):
```
timoballard, line 844: Could we add a couple of tests around auth/z? Perhaps one for an unauthenticated request, and another for a user who doesn't have access to the SAC in question

timoballard (commented):
Left a comment suggesting a couple additional tests we might want to add, otherwise LGTM!

backend/api/test_views.py
geekygirlsarah: Sure, can work on that.

timoballard (approved):

PR #502, opened 2022-09-28
Return report_id in submission response

### LINKED ISSUE #501: Return report ID after completing pre-SAC flow
When completing the last step of the pre-SAC flow, the front-end `POST`s to `/sac/accessandsubmission` and gets back a response that looks like:

```
{
next: "TBD"
sac_id: 18
}
```

As a consequence of #399, the SAC form needs a `reportId` query param in order to load the right data along with the form. This ID is not the `sac_id` above, but the longer report ID that we're using for access control, etc. (having the form of something like `2022ZEL0001000006`). The `/sac/accessandsubmission` endpoint should give back this latter ID in addition to/instead of the `sac_id` so the frontend can build the URL to the next step to the SAC form.

### Tasks
- [ ] Update `POST /sac/accessandsubmission` view to return `report_id` instead of `sac_id`
- [ ] Update tests for this view to accommodate this change

timoballard:
Resolves #501 

Replaces `sac_id` with `report_id` in the response returned from `POST /sac/accessandsubmission`

### DIFF
--- a/backend/api/views.py
+++ b/backend/api/views.py
@@ -192,7 +192,7 @@ class AccessAndSubmissionView(APIView):
             request.user.profile.entry_form_data = {}
             request.user.profile.save()
 
-            return Response({"sac_id": sac.id, "next": "TBD"})
+            return Response({"report_id": sac.report_id, "next": "TBD"})
 
         return Response({"errors": serializer.errors})
 
--- a/backend/api/test_views.py
+++ b/backend/api/test_views.py
@@ -385,7 +385,7 @@ class AccessAndSubmissionTests(TestCase):
         )
         data = response.json()
 
-        sac = SingleAuditChecklist.objects.get(id=data["sac_id"])
+        sac = SingleAuditChecklist.objects.get(report_id=data["report_id"])
 
         creator_access = Access.objects.get(sac=sac, role="creator")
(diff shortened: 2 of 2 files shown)


geekygirlsarah (approved):
I feel like it's cheating if you just copy and paste the same thing 8 times, lol

But LGTM 👍

backend/api/views.py
```diff
@@ -192,7 +192,7 @@ def post(self, request):
             request.user.profile.entry_form_data = {}
             request.user.profile.save()
 
-            return Response({"sac_id": sac.id, "next": "TBD"})
+            return Response({"report_id": sac.report_id, "next": "TBD"})
```
geekygirlsarah, line 195: [NB] Should "next" still exist? Or point to whatever the screen to review it is or something? Or should it still be TBD since an actual future step is undetermined?

heymatthenry (approved):
Seems like that should do the trick AFAICT

backend/api/views.py
heymatthenry: Fair question. As of now this doesn't do anything (clearly, since "TBD") because we're letting the front-end determine where to go next. I can imagine a world where we'd prefer to have that in the DB instead.

PR #505, opened 2022-09-30
Endpoint for retrieving JSON Schema 

### LINKED ISSUE #481: Backend endpoint for JSON schema file
Breakout issue of #443 

### Background

The FAC has to generate an Excel file for the user to download to fill in with their federal awards. This task is specifically for an API endpoint that will provide a JSON schema file that the frontend and backend can use for validating data and generating Excel files.

The endpoint should be something like `/schemas/2023/federal_awards`.

- [ ] API endpoint exists
- [ ] Endpoint returns JSON schema file
- [ ] Test(s) to ensure file is sent

### Tasks

- [ ] Create API endpoint
- [ ] Send JSON schema as return item
- [ ] Add test(s) to ensure this works as expected

timoballard:
Resolves #481

- Adds an endpoint for retrieving the JSON schema for a given fiscal year and type (`GET /schemas/<fiscal_year>/<schema_type>`)
- Adds a placeholder JSON file (`/backend/audit/schemas/2023-metadata.json`) for testing purposes


backend/api/test_views.py
```diff
@@ -1190,3 +1190,68 @@ def test_detail_match_unsubmitted_returns_404(self):
...
+
+        self.user = baker.make(User)
+        self.client = APIClient()
+        self.client.force_authenticate(user=self.user)
+
+    def path(self, fiscal_year, type):
+        """Convenience method to get the path for a particular year and schema type)"""
+        return reverse("schemas", kwargs={"fiscal_year": fiscal_year, "type": type})
+
+    def test_authentication_required(self):
```
geekygirlsarah, line 1213: Hmm, _is_ authentication required to get schemas? Should they be publicly available?

backend/api/test_views.py
```diff
@@ -1190,3 +1190,68 @@ def test_detail_match_unsubmitted_returns_404(self):
...
+
+                    self.assertEqual(response.status_code, 200)
+
+    def test_invalid_fy_returns_404(self):
+        """
+        Requests for an invalid fiscal_year return a 404
+        """
+        for schema_type in self.schema_types:
+            with self.subTest():
+                url = self.path("1899", schema_type)
```
geekygirlsarah, line 1243: [adjusts monocle] 1899 t'was but a good year, indeed!

timoballard: 🧐

backend/api/test_views.py
timoballard: Good question. Here's how I'm thinking about it - as things stand, authentication is (at least indirectly) required here, because you need to login in order to access the audit creation/submission steps. And more generally, I tend to lean toward a `deny by default, permit by exception` approach, where there need to be specific reasons to *not* require auth, rather than the other way around.

That said, there could very well be good reasons for these to be public. Any thoughts on this @tadhg-ohiggins ?

geekygirlsarah: My thought is that we're opening up the API to the world (with auth, of course) to interact with the site, so the schemas should be available to the world so they know what they're doing. I think your approach is good, but this is one of those things that felt it could be open since it doesn't deal directly with any data.

timoballard: Fair point, and agreed (especially in this case, since the schemas will be publicly available in any case with this being an open source project). I'll let @tadhg-ohiggins weigh in in case there are any other considerations here, but otherwise I'm happy to remove auth here

timoballard: Removing auth from this endpoint - thanks for feedback @geekygirlsarah!

geekygirlsarah (approved):

PR #508, opened 2022-10-03
Backend: add Makefile target to define the scope of tests to run

tadhg-ohiggins:
When all is too much
We want to narrow the scope
And ignore cover.

-----

Add `make nctest` to run test without coverage, and to support something like `make nctest fac.test.scope=audit.test_models` to run a subset of tests.


geekygirlsarah (approved):
Wait, we had a Makefile in here the whole time???

👍 to the PR, 👎 to me not realizing this was here.

timoballard (approved):

PR #574, opened 2022-11-02
Add cms

LindsayYoung:
First pass at content pages

This creates a cms app so that we can have a clear place for content pages. This will also help with assigning scoped permissions later.

There are 2 example pages to get the structure going. The home page lists the content pages. The content pages are at their own URL.

We would still like to move the base template out of the current app directory and in the template directory above. We can do that as we add the real content.


geekygirlsarah (approved):

geekygirlsarah:
@LindsayYoung The CI pipeline runs `flake8 .` and `black .` both and basically this failed tests for just formatting things and extra unused imports. I fixed those. This is good to merge.

LindsayYoung:
Closing this in favor of `adding-static` branch

PR #587, opened 2022-11-14
Adding static

LindsayYoung:
This has:
- Static files and assets are built in the docker container
- A script that builds assets  locally
- CMS app that can be used for future content, FAQs resources etc
- Includes static setup from @geekygirlsarah 
- Using `collectstatic` and `load static` to serve assets 

Next steps after this PR:
- Get this working on cloud.gov
    - Confirm names (I would suggest making app and service names consistent across the spaces)
    - Trying out the build on cloud.gov, making sure we have all the build steps


tadhg-ohiggins:
Converted back to draft while thinking about history cleanup.

PR #611, opened 2022-11-22
Login.gov re-integration

### LINKED ISSUE #568: Update Login.gov integration for unified Django application
Breakout issue of #556

### Background
The current LoginGov integration follows the PKCE flow and is driven primarily by the static FAC-Frontend site. Moving to a more traditional server-rendered Django application will require us to use a different [OpenID Connect flow](https://developers.login.gov/oidc/#choosing-an-authentication-method) (specifically `private_key_jwt`). With this flow, the server is involved in the initial authentication exchange with LoginGov and session creation. This should obviate the need for the [current token-based API authentication](https://github.com/GSA-TTS/FAC/blob/main/backend/users/auth.py) in favor of [session authentication](https://www.django-rest-framework.org/api-guide/authentication/#sessionauthentication).

- [x] Our Login.gov configuration has been updated to use the `private_key_jwt` flow
- [x] Login/Logout functionality works as before (nothing about the authentication experience should appear different to end-users) 

### Tasks

- [x] Update LoginGov configuration to  use `private_key_jwt` flow
- [x] Remove `ExpiringTokenAuthentication` backend + references
- [x] Remove `JwtUpsertAuthentication` backend + references
- [x] Remove `djangorestframework-simplejwt` dependency + references
- [x] Select and incorporate an OIDC authentication backend to manage the authentication flow with LoginGov (`mozilla-django-oidc` is one option)
- [ ] Revisit #320 and document any outstanding edge cases related to LoginGov users and Access

timoballard:
Resolves #568 

Updates our Login.gov integration to work with the [unified Django application](https://github.com/GSA-TTS/FAC/blob/main/docs/architecture/decisions/0012-switch-to-unified-django-web-app.md). This primarily involved switching our authorization flow from `PKCE` to `private_key_jwt`


backend/audit/migrations/0014_alter_singleauditchecklist_federal_awards.py
```diff
@@ -0,0 +1,23 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="singleauditchecklist",
+            name="federal_awards",
+            field=models.JSONField(
+                blank=True,
+                null=True,
+                validators=[audit.validators.validate_federal_award_json],
```
timoballard, line 20: FYSA @tadhg-ohiggins looks like we missed the validator migration in #588

tadhg-ohiggins: Yes, I saw that in #608; might pull it out as part of git history cleanup.

tadhg-ohiggins: Addressed via https://github.com/GSA-TTS/FAC/pull/613 and the magic of `git rebase`.

LindsayYoung:
We will want to get rid of the default Django auth and replace it entirely with Login. Right now /admin is prompting me for username and password. This could be a seperate follow-up PR. Just wanted to note we need this.

timoballard:
thanks @LindsayYoung! I think this might be worth looking into now. Setting `is_staff=True` on your user allows you into `/admin` as expected, so I think this might just be an issue with what happens when `is_staff=False` - I'm assuming it should just be some kind of generic forbidden page? cc: @tadhg-ohiggins @melchoyce

LindsayYoung:
I think disabling would be fine for now. Ideally, we would want it to throw you to Login.gov so they can login in.

timoballard:
@LindsayYoung just pushed a change to redirect to Login.gov if you arrive at `/admin` unauthenticated. If you're authenticated but not staff, you still get the default Django login page (see below), so maybe adding a custom template for that scenario is a separate PR? 

image

melchoyce:
I can whip together a design for a custom "permission denied" screen if that helps?

LindsayYoung:
The authentication is working well! One last thing with the redirects to iron out:

1. Not signed in 
2. Go to`/admin` 
3. That takes me to login where I signed in 
4. Then, I am getting a redirect loop

tadhg-ohiggins (approved):

PR #614, opened 2022-11-28
Adding static assets

tadhg-ohiggins:
PR to just add the static assets from https://github.com/GSA-TTS/FAC/pull/587


tadhg-ohiggins:
Closed because we decided on a slightly different approach.

PR #615, opened 2022-11-28
Adding static content step #1

tadhg-ohiggins:
Breaking #587 into three parts, of which this is the first.
-----
Adding app to settings

First pass at content pages

This creates a cms app so that we can have a clear place for content pages. This will also help with assigning scoped permissions later.

There are 2 example pages to get the structure going. The home page lists the content pages. The content pages are at their own URL.

We would still like to move the base template out of the current app directory and in the template directory above. We can do that as we add the real content.

Moving from a app based template approach to a consolidated, project based approach

(wip) Enable static serving, add index.html from 11ty repo. (rest of repo goes in backend/static folder)

Linting/formatting.


LindsayYoung (approved):
This was approved by @tadhg-ohiggins and written by @LindsayYoung 

Extra thanks to @tadhg-ohiggins for the git commit clean up!

PR #616, opened 2022-11-28
Adding static content step #2

tadhg-ohiggins:
Breaking https://github.com/GSA-TTS/FAC/pull/587 into three parts, of which this is the second.


LindsayYoung (approved):
static files from @heymatthenry

PR #617, opened 2022-11-28
Adding static content step #3

tadhg-ohiggins:
Breaking https://github.com/GSA-TTS/FAC/pull/587 into three parts, of which this is the third.


tadhg-ohiggins:
@heymatthenry Lindsay and I have been staring at this for a while so would appreciate your eyes on it.
#615 was the pre-asset step, in case you want to see that. 616 was the asset check-in PR; I didn't link it here because clicking on it might slow your browser to a crawl and it's just adding 2,600+ files to the tree.

LindsayYoung (approved):

heymatthenry (approved):

PR #620, opened 2022-11-29
Adding static content step #4

LindsayYoung:
This has the real styling and assets and breaks out templates into manageable components. 

Before making this a PR, @heymatthenry can you make sure this is in good order? Thanks


heymatthenry:
This PR 

* re-implements the homepage template
* adds tooling for processing assets
* configures the build tools to process assets

localhost_8000_ (1)

heymatthenry:
@LindsayYoung LGTM! I pulled it down to make sure it still does what it's supposed to, and it seems like we're good. Thank you!

PR #629, opened 2022-12-02
Wire login button to login url

### LINKED ISSUE #628: Connect frontend login buttons to backend endpoints
### Background
Now that the Login.gov re-integration is complete, we need to update the frontend login buttons to point to the appropriate login URLs.

- [ ] Clicking login takes the user to Login.gov to initiate the authentication flow.

timoballard:
Resolves #628 

Connects the login button in the primary nav to the backend login URL.


tadhg-ohiggins:
This looks good, but it also looks like this is supposed to be a modal rather than sending the user to a different page. I think we can check this in and modify the modality later, but I am going to add changing the homepage links to this PR.

tadhg-ohiggins (approved):

PR #630, opened 2022-12-02
Add whitenoise to requirements

tadhg-ohiggins:
Sigh… dependencies.
We use `whitenoise` locally
But cite it in prod.

-----

If this works we can consider adjusting our references to `whitenoise` so that they only occur if we're not in production.


timoballard (approved):

LindsayYoung:
closed with #631

PR #632, opened 2022-12-05
Fix pa11y config

heymatthenry:
This PR makes some small changes to get pa11y tests running at the command line. It does not make those tests pass.


tadhg-ohiggins:
@heymatthenry @LindsayYoung AFAICT this needs to adjust for the fact that we removed `workflows/test.yml` and need to account for that in this PR's workflow. Also it might (not sure) need to call `test:admin` or `test` instead of `test:all` in this step:

```yml
  - name: Run pa11y-ci against admin
        working-directory: ./backend
        run: npm run test:all
 ```
 
 Although I might be reading the details wrong on that one. Any chance you could find time to pair on sorting the merge conflict out? Hopefully it won't take long with both of you looking at it, now that `workflows/test_and_deploy.yml` is (hopefully…) more stable.

PR #640, opened 2022-12-07
Pattern validation for Federal Awards program number

### LINKED ISSUE #633: Update Federal Award JSON Schema approach to handling prefixes/extensions
Breakout issue of #519 / #530

### Background

After discussion with Mark Dixon, we concluded that we cannot constrain agency prefixes and program numbers to a set list, but must support the following edge cases:

-   If the user's intended prefix isn't in the list, they're supposed to choose `99` as their agency prefix, but there are no valid entries starting with `99` in the SAM.gov list.
-   CFDA three digit extension can apparently be `RD`, which is neither three characters long nor composed of digits. There are no valid entries with that extension in our list, so users would not be able to enter this.
-   CFDA three digit extension can apparently be `U` followed by two digits if the user doesn't know their actual extension. Obviously there are no valid entries with those extensions in our list, so users could not enter those either.
-   The instructions for Additional Award Identification state that it's required if either of the previous two cases are true, which isn't possible with our current validation.
-   Federal Program Name can (must) be manually entered if the extension is `RD` or `U` plus some random numbers, which also doesn't apply for us.
*   Following discussion with Mark, we also decided to support an optional trailing letter for extension, since some agencies use this and have requested support for it.
*   ~~Agency prefixes start at 10, and program numbers start at 100~~. Apparently that’s not true, and they start at 001.

A JSON Schema rule for the above looks like this:

```json
{
    "type": "string",
    "pattern": "^[1-9]{1}[0-9]{1}\\.([0-9]{3}[a-zA-Z]{0,1}|U[0-9]{2}|RD)$"
}
```

- [x]   We’ve changed the Federal Awards JSON schema to use this regular expression (or an improved one if we find problems).
- [x]   We’ve updated the tests accordingly.

### Tasks

- [x]   Change the schema.
- [x]   Update the tests.

timoballard:
Resolves #633 

Updates the Federal Awards JSON Schema to use a pattern validator for ProgramNumber.


backend/audit/test_validators.py
```diff
@@ -33,6 +40,229 @@ def test_validation_is_applied(self):
...
+    def test_prefix_over_99(self):
+        """
+        Prefixes over 99 should fail
+        """
+        simple = jsoncopy(SchemaValidityTest.SIMPLE_CASE)
+
+        # pick a prefix between 100 and 999 (invalid)
+        prefix = f"{randrange(100, 1000):03}"
+        # pick an extension beteween 100 and 999 (valid)
+        extension = f"{randrange(100, 1000):03}"
```
tadhg-ohiggins, line 69: Since we discovered that `001` was a valid extension, I think this should be `extension = f"{randrange(1,1000):03}"`

timoballard: fixed - thanks!

tadhg-ohiggins (approved):

Issue #584, opened 2022-11-09
Grant team access to OROS cloud.gov spaces


tadhg-ohiggins:
Breakout issue of [#579](https://github.com/GSA-TTS/FAC/issues/579)

### Background
After [#583](https://github.com/GSA-TTS/FAC/issues/583) is done, we need to use the processes it defines to actually grant the required access to team members

This should be done by someone with the administrative rights to do it…

### Acceptance Criteria (We'll know we're done when...)

- [x]   All team members who need admin access have it.
- [x]   All team members who need developer access have it.

### Tasks

- [x]   Grant appropriate access to all team members.

Issue #583, opened 2022-11-09
Develop processes for adding and removing access (we need those for ATO)


tadhg-ohiggins:
Breakout issue of [#579](https://github.com/GSA-TTS/FAC/issues/579)

### Background
We need to have well-defined, well-documented processes for controlling access to the OROS FAC cloud.gov spaces.

### Acceptance Criteria (We'll know we're done when...)

- [x]   We know who’s responsible for maintaining access administration for these spaces.
- [x]   We know who to go to if there’s some issue with the current administrator(s) or their access.
- [x]   We have documentation on who should have access, how to get access, how to remove access, and how often those things are reviewed (and/or what events might also trigger review).
- [x]   We’ve updated our on-boarding and off-boarding documentation to reflect the above.

### Tasks

- [x]   Identify whoever has current administrative rights.
- [x]   Identify who should have administrative rights.
- [x]   Address “who administrates the administrators?” question?
- [x]   Identify backup personnel for administration.
- [x]   Determine cadence for periodically reviewing who has administrative rights and who is listed as backup(s).
- [x]   Determine what events trigger access review and who’s responsible for doing that review.
- [x]   Document all of the above for ATO purposes.
- [x]   Update on-boarding and off-boarding documentation accordingly.

tadhg-ohiggins:
2022-12-14 currently blocked on deployment/terraform PRs.

mogul:
@tadhg-ohiggins are you okay with closing this one without breaking the last checkbox out? (I though about adding a GitHub action that creates an issue, and assigns it to the fac-admins, but really I can just put a reminder in the calendar too...)

mogul:
_ping_ @tadhg-ohiggins ...

tadhg-ohiggins:
@mogul If there's a calendar reminder then I think that qualifies as closing the item for **Determine cadence for periodically reviewing who has administrative rights and who is listed as backup(s).**

mogul:
Done. 
![image](https://user-images.githubusercontent.com/47762/222645658-beb7968a-2c8f-4af6-978b-58a20007b2af.png)

Issue #582, opened 2022-11-09
Create new spaces for dev, staging, and production within the cloud.gov OROS org


tadhg-ohiggins:
Breakout issue of [#579](https://github.com/GSA-TTS/FAC/issues/579)

### Background
We need to create new spaces for `dev`, `staging`, and `prod` within the cloud.gov `gsa-tts-oros-fac` org.

#### Notes on distinction between `dev` and `staging` from 2022-11-09 dev huddle

Development in early phases will primarily be done on `dev`.

While both `dev` and `staging` will have persistent data across pushes, developers should feel free to remove existing data on `dev` and start from scratch rather than struggling with difficult data migrations for data that’s purely for test purposes anyway.

Once we start working against `staging`, we’ll probably treat its data more carefully and deleting the database will require more discussion.

### Acceptance Criteria (We'll know we're done when...)

- [ ]   `dev`, `staging`, and `prod` spaces all exist within the `gsa-tts-oros-fac` OROS organization.

### Tasks

- [ ]   Create `dev`, `staging`, and `prod` in the `gsa-tts-oros-fac` org.

PR #639, opened 2022-12-07
moving apps to new cloud.gov spaces using oneflow rules

LindsayYoung:
#580 

Setting up cloud.gov:
- Creates manifests for each environment in new org
- Removes bug where we overwrite SAM_API key
- First pass at github actions that would conform to oneflow rules
- Services are created
- Adds a testing gate for deploys

ToDo:
- [x] update egress rules
- [x] put service deployer account creds in github
- [x] add redirect uri redirect_uri to login.gov
- [x] update tag rules
- [ ] remove dev route from current app in 10x so we can use it
- [x] update documentation
- [x] clean up extra deployers

Terraform will be different PR [#650 -Bret]
We can update to use environments later.

## Tasks
For dev:
```
cf create-service aws-rds medium-gp-psql fac-db
cf create-service s3 basic-public fac-public-s3
cf cups fac-key-service -p "{\"SAM_API_KEY\":\"$SAM_API_KEY\", \"DJANGO_SECRET_LOGIN_KEY\":\"$DJANGO_SECRET_LOGIN_KEY\"}"' 
cf create-service-key dev-deployer dev-deployer-key
```

For staging:
```
cf create-service aws-rds medium-gp-psql fac-db
cf create-service s3 basic-public fac-public-s3
cf cups fac-key-service -p "{\"SAM_API_KEY\":\"$SAM_API_KEY\", \"DJANGO_SECRET_LOGIN_KEY\":\"$DJANGO_SECRET_LOGIN_KEY\"}"' 
cf create-service-key staging-deployer staging-deployer-key
```

For production:
```
cf create-service aws-rds medium-gp-psql-redundant fac-db
cf create-service s3 basic-public fac-public-s3
cf cups fac-key-service -p "{\"SAM_API_KEY\":\"$SAM_API_KEY\", \"DJANGO_SECRET_LOGIN_KEY\":\"$DJANGO_SECRET_LOGIN_KEY\"}"' 
cf create-service-key production-deployer production-deployer-key
```
Note that this will be managed by terraform in the future

PR #650, opened 2022-12-08
Manage infrastructure using terraform

mogul:
Ref #582 #583 #584 

This PR introduces the `terraform/` subdirectory, which is intended to be used for managing all "non-code" configuration for our environments: Services, routes, access, security groups, etc. It generally implements the ADR described in #660.

- A `bootstrap` directory was used for setting up the remote Terraform state in an S3 service in the `production` space, as described in the README.md.
- There's a directory with a module for each managed environment/space (`dev`, `staging`, `production`, and `management` initially) per [this guidance from Truss](https://truss.works/blog/building-for-the-future-with-terraform).
  - The `management` module ~~provisions and~~ manages access to the spaces
  - The other modules are intended to manage non-code assets in each space; right now there's nothing managed by Terraform, but a commented-out DB service shows what that should look like.
- There are GitHub workflows for previewing changes by posting the `terraform plan` on PRs
- A reusable `terraform apply` workflow is here, but it's not actually called by anything in `test_and_deploy.yml` yet; I'm waiting for that to stabilize before I submit another PR to hook it in!


geekygirlsarah:
Ahhh Terraform, we meet again...

tadhg-ohiggins:
@mogul and I discussed some things we're looking out for, and came up with this:

----
Competing interests:

1.  We want different Terraform directories so that we can support e.g different database plans for `dev` versus `staging` or `production` (more storage/compute resources/persistence/etc.).
2.  We want to avoid a situation where a code change that depends on an infrastructure change goes to `dev` and look great, but only has the Terraform infrastructure rules for `dev` in it. Once pushed to `staging`, this would break because it doesn’t contain the corresponding Terraform rules for `staging` but nothing in our process would warn us of this prior. (This also applies for things looking fine in `staging` but where we forgot the rules for `production`.)

We can address both of these interest via the following Terraform setup:

*   Use Terraform modules extensively in config.
*   The environments only specify where they are different from the defaults.
*   The defaults are all in a shared module.

This is an example of how that might look:

```terraform
module "fac-dev" {
  source = "../modules/fac"
  database_plan = "micro-psql"
}
```

The above example uses the default from `../modules/fac` but overrides the value of `database_plan` to use a smaller instance for the `dev` environment.

If we want something to be different between environments, we have to explicitly allow for customization by providing an input variable—Terraform modules must specify which of their variables can be overridden.

Example of modules/fac (where we concentrate all the config details with staging/prod defaults, but allow for override through variables):

```terraform
variable "database_plan" {
  type        = string
  description = "name of the service plan name to create"
  default     = "xlarge-gp-psql-redundant"
}

module "database" {
  source = "../modules/database"

  …

  rds_plan_name    = var.database_plan
}
```

mogul:
Now rebased on #639 since we want terraform changes for a given environment to be deployed right before code changes, and @LindsayYoung's PR isn't merged yet. Will mark non-draft once 639 is in.

mogul:
FYI: I rebased this onto `main` just now so we can potentially land this around the other changes in flight (to refactor `test_and_deploy.yml` to use environments).

terraform/README.md
```diff
@@ -0,0 +1,151 @@
...
+  |- variables.tf
+  |- run.sh
+  |- teardown_creds.sh
+  |- import.sh
+- <env>/
+  |- main.tf
+  |- providers.tf
+  |- secrets.auto.tfvars
+  |- variables.tf
+- shared/
```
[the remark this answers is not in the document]
mogul: I'll be making big changes to the bootstrap process in another PR that will revamp the README entirely so I'm going to hold off on here for now.

terraform/shared/modules/base/variables.tf
```diff
@@ -0,0 +1,26 @@
+variable "cf_org_name" {
```
mogul, line 1: Add a note here indicating that defaults are production defaults, and they're exposed as variables to enable environments to specify alternate config when needed. Reference the production module as the "no config necessary" invocation.

terraform/management/README.md
```diff
@@ -0,0 +1,19 @@
+# Management environment
+
+The "management" is a "meta" configuration, in that it specifies
+configuration pertaining to spaces and users across environments. In an
+initial bring-up/contingency recovery situation, this is the first environment
+to apply.
+
+Among other things, this module...
+- configures up the spaces for dev, staging, and production
```
mogul, line 9: Not yet, due to provider bug; to be addressed in bootstrap PR, so ~~strikethrough~~ here until that's fixed.

terraform/management/README.md
```diff
@@ -0,0 +1,19 @@
...
+- configures user access to those spaces
+- configures application security groups (ASGs) for the spaces
```
mogul, line 11: Again, ~~strikethrough~~ due to the upstream provider bug

terraform/management/config.tf
```diff
@@ -0,0 +1,38 @@
...
+    "sarah.withee@gsa.gov",
+    "tadhg.ohiggins@gsa.gov",
+    "timothy.ballard@gsa.gov",
+  ]
+
+  managers = [
+    # The OrgManager user explicitly sets itself as a SpaceManager for each space
+    var.cf_user,
+
+    # Mirrors @GSA-TTS/FAC-admins
```
mogul, line 25: Copy this comment up for the developers list too.

terraform/management/main.tf
```diff
@@ -0,0 +1,31 @@
...
+# Ensure spaces exist
+resource "cloudfoundry_space" "space" {
+  for_each = toset(local.spaces)
+  name     = each.key
+  org      = data.cloudfoundry_org.org.id
+
+  # Disallow SSH access in production
+  allow_ssh = each.key != "production"
+
+  # All spaces have full public egress (for now)
```
mogul, line 20: This comment should be on the config, not the resource

terraform/management/main.tf
```diff
@@ -0,0 +1,31 @@
...
+  asgs = [for d in data.cloudfoundry_asg.asgs : d.id]
+}
+
+# Everyone has SpaceDeveloper permission in all spaces
```
mogul, line 24: Move comment up to the configuration

mogul (commented):
Comments from my review with Tadhg just now.

tadhg-ohiggins (approved):
Went over this in pairing, looks good.

PR #651, opened 2022-12-08
Create new BPMN model folder + first model

MelissaBraxton:


JeanMarie-TTS:
We should adopt the appropriate naming standards for BPMN. I think the convention is to have `bpmn` as the file extension.

JeanMarie-TTS:
@MelissaBraxton Should we also have a short but descriptive name for the folder (without spaces preferably) - just `bpmn` perhaps? We don't need to include a date or version number in the file name.

mogul:
- I don't think someone unfamiliar with BPMN would guess what they're going to find in the `BPMN` folder, so maybe call it `workflow` or `process`? 
- I'd like to see a `README.md` in the directory with some context: This is where we document the set of steps and interactions involved is managing submissions, we document them in something called BPMN (with links), they're diagrams which you can load up in any BPMN tool (but here's bpmn.io)
- I think the directory should also include a screenshot of the current rendered version of each file, but I'm not sure how to keep that up to date easily. 🤔

mogul (changes requested):
- I don't think someone unfamiliar with BPMN would guess what they're going to find in the `BPMN` folder, so maybe call it `workflow` or `process`? 
- I'd like to see a `README.md` in the directory with some context: This is where we document the set of steps and interactions involved is managing submissions, we document them in something called BPMN (with links), they're diagrams which you can load up in any BPMN tool (but here's bpmn.io)
- I think the directory should also include a screenshot of the current rendered version of each file, but I'm not sure how to keep that up to date easily. 🤔

MelissaBraxton:
Yes, I am working on the readme. This is just an initial commit to get the folder up and running.

geekygirlsarah (approved):
👍

MelissaBraxton (commented):
This initial commit will just create the folder and add one file. I'll address the other req. changes in another PR.

JeanMarie-TTS:
> * I think the directory should also include a screenshot of the current rendered version of each file, but I'm not sure how to keep that up to date easily. 🤔
@mogul I was able to see the diagram locally using a VS Code extension. Rather than including a screenshot, I suggest we include some recommendations for viewing tools in the README

mogul (approved):

PR #664, opened 2022-12-12
Update roadmap.md

pkigongo:
Added  a note explaining why both delivery options are listed, even thought we're going with Option 1. We also removed the list of Milestones from Option 1 and provided a link to those milestones instead.


JeanMarie-TTS:
@melchoyce , @mogul , can one of you please review and approve this PR?

mogul (approved):

PR #671, opened 2022-12-15
General Information JSON schema

### LINKED ISSUE #668: General Information JSON Schema
## Background

We need to create a JSON Schema to validate the General Information section of the SingleAuditChecklist. This section should include the following fields:

```
- auditee_fiscal_period_start
- auditee_fiscal_period_end
- ein
- ein_not_an_ssn_attestation
- multiple_eins_covered
- auditee_uei
- multiple_ueis_covered
- auditee_name
- auditee_address_line_1
- auditee_city
- auditee_state
- auditee_zip
- auditee_contact_name
- auditee_contact_title
- auditee_phone
- auditee_email
- user_provided_organization_type
- met_spending_threshold
- is_usa_based
- auditor_firm_name
- auditor_ein
- auditor_ein_not_an_ssn_attestation
- auditor_country
- auditor_address_line_1
- auditor_city
- auditor_state
- auditor_zip
- auditor_contact_name
- auditor_contact_title
- auditor_phone
- auditor_email
```

## Acceptance Criteria

- [ ] We have a schema (`GeneralInformation.schema.json`) that serves as the source of truth for valid general information submissions

timoballard:
Resolves #668 

- Adds `GeneralInformation.schema.json` to validate the "General Information" section of the SingleAuditChecklist
- Adds tests against parts of the schema that involve customization (e.g. regex patterns, enumerations)


backend/audit/test_schemas.py
```diff
@@ -1,20 +1,264 @@
...
+from random import choice, randrange
 
 # Simplest way to create a new copy of simple case rather than getting
 # references to things used by other tests:
 jsoncopy = lambda v: json.loads(json.dumps(v))
 
+
+# wrap the validate function to include a format checker
+def validate(instance, schema):
+    return jsonschema_validate(instance, schema, format_checker=FormatChecker())
```
timoballard, line 18: the `format_checker` bit here is necessary if you want `jsonschema` to validate type-formatted strings (e.g. `email`, `date`, etc). We may want to wrap the `validate` call in a utility function so that we don't have to remember to do this in each case

backend/audit/test_schemas.py
```diff
@@ -1,20 +1,264 @@
...
+
+    def test_simple_pass(self):
+        """
+        The simple case should be valid
+        """
+        schema = self.GENERAL_INFO_SCHEMA
+
+        validate(self.SIMPLE_CASE, schema)
+
+    def test_invalid_auditee_fiscal_period_start(self):
```
timoballard, line 75: Undecided on whether these tests add any value- in one sense, they're testing `jsonschema's` internal date validation functionality (which isn't valuable). In another sense, they're testing that we are in fact marking this field as needing to be date-formatted (which is valuable).

tadhg-ohiggins: It's probably better to keep them as an additional safety net against accidentally not marking a field as a date.

tadhg-ohiggins (approved):

PR #672, opened 2022-12-15
auto deploy 

LindsayYoung:
Uses github actions to build assets and deploy.

Thank you to @tadhg-ohiggins @geekygirlsarah  & @timoballard for lugging this over the finish line!!

We moved npm run build to its own step in test_and_deploy.yml and put that ahead of the 18f/cg-deploy-action@main actions.
We switched back to using 18f/cg-deploy-action@main to push to cloud.gov and removed the debug step of trying to do that from a Docker container.
We put python manage.py collectstatic --noinput into the Procfile.
The result is that after having deleted all the assets in the bucket earlier, the last build (with the above changes) pushed the assets to the S3 bucket.
One of the key issues was that running collectstatic anywhere outside of the CF instance meant that Django couldn’t access the S3 credentials that are stored in VCAP_SERVICES, which is why we moved collectstatic back into the Procfile.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -15,7 +15,12 @@ WORKDIR ..
 
 RUN \
   apt-get update && \
-  apt-get install -yqq apt-transport-https wget gnupg2
+  apt-get install -yqq apt-transport-https wget gnupg2  && \
+  apt-get install wget && \
+  wget -q -O - https://packages.cloudfoundry.org/debian/cli.cloudfoundry.org.key | apt-key add - && \
+  echo "deb https://packages.cloudfoundry.org/debian stable main" | tee /etc/apt/sources.list.d/cloudfoundry-cli.list && \
+  apt-get update && \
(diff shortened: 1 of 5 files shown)


tadhg-ohiggins (approved):

backend/Dockerfile
```diff
@@ -37,6 +37,8 @@ COPY . /src/
 WORKDIR /src/
 
 RUN npm install
+RUN chown -R 1001:123 "/root/.npm"
+RUN npm run build && python manage.py shell --command="import environs; env = environs.Env(); print(env.str('ENV', 'UNDEFINED').upper())" && python manage.py collectstatic
```
timoballard, line 41: the env printing here is a hangover from our debugging yesterday, can/should be removed

LindsayYoung: can do

timoballard (approved):

PR #676, opened 2022-12-16
Run collect static twice

LindsayYoung:
It is run by the buildpack and run in the Procfile.

Collectstatic needs to be run by the Procfile to get assets to s3. This adds images etc to the bucket, but not NPM built assets. Collectstatic needs to be run by the buildpack so that the Procfile doesn't timeout.


tadhg-ohiggins (approved):
One of those “ya gotta do what ya gotta do” situations.

PR #681, opened 2022-12-19
Remove collectstatic to confirm that it's really the problem

tadhg-ohiggins:
Too tired for haiku
Too many syllables? Or
Maybe just perfect.

-----

Monday-morning sanity check to confirm that yes, collectstatic really is the cause of the deployment failure.


geekygirlsarah (approved):

tadhg-ohiggins:
Closed because we worked through the actual problems in https://github.com/GSA-TTS/FAC/pull/702

PR #684, opened 2022-12-20
Add workflow to ensure PNGs are updated whenever BPMN changes

mogul:
Forgot to sync images?
No need to worry.
I'll take care of that for you.


mogul:
The removed .bpmn file didn't show up in the [list of removed files](https://github.com/GSA-TTS/FAC/actions/runs/3738442243/jobs/6344530691#step:6:2), so the logic to remove the corresponding rendered PNG and SVG file didn't run! 🤔

mogul:
Hmm, same problem with this other changed-files action. It's because if you look at the set of "changed files" since the start of the branch, add+remove==nochange. We need to check for files changed since the last time the BPMN was in sync, earlier in the PR!

Next thing to trye: Follow the [`Get all changed files using the last successful commit of the base branch > Push event` example here](https://github.com/tj-actions/changed-files#examples). (I tried this previously, but couldn't figure out an error.)

mogul:
For those who might be watching: I've just gone through and removed all the auto-generated posts above about PRs that were generated while I was working on this. That way I can demonstrate the full solution in action clearly before the PR review.

mogul:
At this point in the process, we can see that the PNG images are up to date with their BPMN source:
image

Now I'm going to push a commit that adds a new .bpmn file... Watch what happens!

mogul:
In response to a new BPMN file being added/modified, the workflow generate a missing PNG file, then made a PR for merging it into the branch and told us about it. The workflow then explicitly failed, which is intentional! This way the original branch can't be merged until the new PNG is in place.
image

Next we're going to merge in that generated PR.

mogul:
Once the PR got merged, the check is passing.
image

Now we're going to remove the test BPMN file, and check that a PR is generated which removes the corresponding PNG file.

mogul:
As expected, the check failed and we got a new PR suggested to fix the problem.
image

Next we'll merge that change in and watch the check go green again.

mogul:
And there we have it! _QED_
image

mogul:
Here's a summary of what you see happening above, based on the list of commits:
image

tadhg-ohiggins (approved):

PR #693, opened 2022-12-20
Test addition and deletion of BPMN files generates PRs with images

mogul:


mogul:
Hmmm, well, this was no more instructive than doing it on #684, will continue experimenting there.

PR #702, opened 2022-12-21
Better deploy process

tadhg-ohiggins:
Put things in their place.
Don't do things multiple times.
All as it should be.

-----

Add `.cfignore` to not put over 2,000 material icon SVG images so that Django’s `collectstatic` step doesn’t time out.

Move steps from `Procfile` to `.profile` and make those steps only run when the initial app instance is deployed.

Remove `--strategy rolling` from push arguments because the action already adds it to the command.


LindsayYoung (approved):
Paired on this 12/21/22

PR #707, opened 2022-12-22
Ly/envs take 2

LindsayYoung:
Reduces code duplication by using GitHub environment specific variables

Changes:
- Break out tests into their own file
- Creates a triggers file to just manage the triggers and sequencing
- Moves deployment to reusable script using GitHub environments

Clean up step:
- [ ] Delete global variable "CF_USERNAME" 
- [ ] Delete global variable "CF_PASSWORD"
(These reference the old 10x org app and are not needed)


tadhg-ohiggins (approved):

mogul:
I took care of removing the repo-level CF_* secrets.

PR #717, opened 2022-12-28
Add general_information JSON field to SingleAuditChecklist model

### LINKED ISSUE #488: Add `general_information` JSONField to SingleAuditChecklist model
### Background
As part of the effort to [future proof our data model](https://github.com/GSA-TTS/FAC/issues/455), we need to add a new field to the `SingleAuditChecklist` model, called `general_information`. This field will be of type `JSONField`, and will store the following data, as discussed in #455:

```
Auditee fiscal period start (matches current question 1).
Auditee fiscal period end (matches current question 1).
Auditee identification tokens and related metadata (matches current question 4).
Auditee info (name, address, etc.) (matches current question 5).
Primary auditor info (name, address, etc.) (matches current question 6).
```

We will need to retain the ability to filter `SingleAuditChecklists` based on fields internal to `general_information`. Django natively supports queries against JSON documents. For example, if we wanted to find all `SingleAuditChecklists` with a UEI of KZV2XNZZN3A8, we could do the following:
```
sacs = SingleAuditChecklist.objects.filter(general_information__auditee_uei="KZV2XNZZN3A8")
```

This query will work for `SingleAuditChecklists` with form version 2023.00 as described above. However, imagine that we create a new form version (e.g. 2023.01), in which the `auditee_uei` field is now part of the `federal_awards` section. The above query would not work as expected.

Queries against the `general_information` field, then, should be centralized in convenience functions, so that we can add support for querying against multiple form versions in a single place. We should extend the `SingleAuditChecklist` model Manager to include any necessary functions for querying `SingleAuditChecklists` based on data stored in the `general_information` field.

- [ ] The `SingleAuditChecklist` model contains a `JSONField` named `general_information`
- [ ] We are able to query for `SingleAuditChecklists` by UEI using syntax such as `SingleAuditChecklist.objects.filter_uei('KZV2XNZZN3A8')`
- [ ] There are tests around the UEI filtering functionality
- [ ] The `general_information` column uses [GeneralInformation.schema.json](#668) as a field validator

### Tasks

- [ ] Create a migration that adds the `general_information` field to the `SingleAuditChecklist` model
- [ ] Create a function `filter_uei` on `SingleAuditChecklistManager` that finds all `SingleAuditChecklists` **across form_versions** with the provided `auditee_uei` (initially, there will only be a single form_version, but this function should be written in such a way that extension to support additional form_versions is straightforward)
- [ ] Write tests around the UEI filtering functionality

timoballard:
Resolves #488 

- Adds a `general_information` `JSONField` column to the `SingleAuditChecklist` model
- Sets up a field validator for the `general_information` column to validate inputs against `GeneralInformation.schema.json`
- Updates serializers and views to use the `general_information` field
- Updates tests


tadhg-ohiggins (approved):
This looks good. I wonder if we're going to need to split `test_views.py` into multiple files.

PR #719, opened 2023-01-03
Pin version of npm v14

LindsayYoung:
## Backstory
We want to run Node v14. I updated the docker file to install Node v14 instead of v12. Pinning the versions in the package.json will cause an error on build if the wrong version is running. 

## Change
Install node 14
Pin npm 8 and node 14


heymatthenry (approved):
LGTM. If we end up leveraging node for much else as we get further along, it's possible that this could bite us, but unless & until we get to that point this should be fine.

heymatthenry (approved):

geekygirlsarah (approved):

PR #720, opened 2023-01-04
Add lighthouse

heymatthenry:
This PR adds [Lighthouse CI](https://github.com/GoogleChrome/lighthouse-ci/) as a Github action to ensure the project is getting at least some automated accessibility testing.

It runs Lighthouse against the homepage (additional pages will need to be added manually to the configuration).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,9 @@ jobs:
       ENV: TESTING
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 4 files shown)


heymatthenry:
@LindsayYoung I can get a 100% when I run it locally but for some reason it's giving me a mere 98% when running in CI. I'm fine with that for now if others are, but it's something we should chase down at some point.

LindsayYoung (approved):
This seems like a great way to get accessibility started. I can make a new PR to remove the unused Pa11y stuff separately.

PR #721, opened 2023-01-05
JSON storage ADR

### LINKED ISSUE #669: ADR for storing SingleAuditChecklist submission data as JSON
We need to write an ADR that documents our decision to store SingleAuditChecklist submission data in the database as JSON. This ADR should discuss the need to accommodate schema changes over time while maintaining the ability to accept submissions for prior schema versions, the desire for a single source of truth for submission validation, and the decision to split each of the top-level submissions sections into their own database column (`JSONField`), each with a corresponding JSON Schema document.

timoballard:
Resolves #669

### DIFF
--- a/.adr-dir
+++ b/.adr-dir
@@ -1,0 +1,1 @@
+docs/architecture/decisions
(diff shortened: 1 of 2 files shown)


LindsayYoung (approved):
Nice summary of the decision

tadhg-ohiggins (approved):

Issue #343, opened 2022-07-12
Implement "My audit submissions" page


petermarks-gsa:
## Acceptance criteria
(Acceptance criteria referring to login routing moved over to be part of #388 )

This issue only covers the page listing a user's audits, and doesn't address routing (or whether or not they should be redirected to their lone audit if they only have one).

### Frontend acceptance criteria
- [x]   The user will see a table containing the audits they're associated with.
- [x]   The table will include the following columns: **Report ID**, **Status**, **Auditee UEI**, **Fiscal Year End Date**, **Auditee Name**.
- [x]   The table will be sortable. The default sort order is **Report ID** ascending.
- [x]   The entries for **Report ID** will be links to the submissions. (URLs for submissions have not yet been implemented, but they should probably be something based on **Report ID**. Until we have a final route for that, the links should go to `/submission/<report_id>` or similar, even if those URLs aren't yet functional.)
- [x]   The header for **Auditee UEI** should have a link/tooltip/etc. for text explaining what a UEI is.
- [x]   There should be a “Start a new submission” form below the table. This consists of some text reminding the user not to duplicate already-existing submissions, a checkbox for agreeing to the terms and conditions, and a **Start a new submission** button.
- [x]   The page exhibits the appropriate behavior if the user clicks that button without checking the checkbox.
- [x]   Checking the checkbox and then clicking the button results in navigation to the first step of a new submission.

## Design
https://www.figma.com/file/1bNncVt1TcFFDhQslqFC2o/FAC-Playground?node-id=2443%3A11490

## Tasks (already broken out in https://github.com/GSA-TTS/FAC/issues/391)
* 🅜 Back end - "My audit submissions" page
  * Surfaces a list of audits with Report id, Status, Auditee UEI, Fiscal year end date
  * Assume no sorting on backend or pagination (speculating that 400 is upper limit, most way less)
  * NEEDS CLARIFICATION: are we filtering available out of the list year
*   🅛 Backend: all audits associated with current user endpoint [#410](https://github.com/GSA-TTS/FAC/issues/410).
* 🅛 Front end - "My audit submissions" page
   * Sortable Columns: Report id, Status, Auditee UEI, Fiscal year end dat

## Subtasks in scope
* https://github.com/GSA-TTS/FAC/issues/391
* https://github.com/GSA-TTS/FAC/issues/424
* https://github.com/GSA-TTS/FAC/issues/410

## Related, but separate stories
[...]

PR #728, opened 2023-01-10
Jm/my audit submissions

JeanMarie-TTS:
# First django implementation of MySubmissions page #343

Completed:
- Add a view in audit app
- Add a html template in audit app
- Refactor home and base html pages to reduce duplication
- Link from Home page to MySubmissions page
- Add a logout option in header.html
- Update config.urls to get to this page

Deferred for later:
- [ ] Update link to Start A New Submission when the relevant page is available.  
- [ ] Make additional changes based on other later issues related to this


heymatthenry:
Probably out of scope for this ticket, but the session data should include the user's name so we should be able to use that rather than this ID.

image

heymatthenry (approved):
LGTM. There are some tweaks to be made to the signout button but I think they're out of scope for this ticket so I'll make a new one for that.

PR #735, opened 2023-01-13
Change STATICFILES_STORAGE setting for non-prod to support django tests

JeanMarie-TTS:
Was getting the following error before this change when running tests using django.test.Client:
ValueError: Missing staticfiles manifest entry for 'compiled/scss/main-post.css'


backend/config/settings.py
```diff
@@ -177,13 +177,13 @@
     DEBUG = env.bool("DJANGO_DEBUG", default=True)
     STATIC_URL = "/static/"
     # Whitenoise for serving static files
-    STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
```
geekygirlsarah, line 180: The `CompressedManifestStaticFilesStorage` adds a MD5 manifest to each file to ensure that if it's changed it's reloaded rather than reused from cache. It seems like this would be a thing we want (and not using the non-manifest version). Does this make sense? Can you explain why we should turn this off?

The error of "Missing staticfiles manifest entry for 'compiled/scss/main-post.css'" also sounds like it might be a `npm run build:esbuild` issue, which I think was fixed from @heymatthenry's PR #732 yesterday. Did you test that first before doing this?

LindsayYoung: @geekygirlsarah you make a good point about this being helpful for local development.

Also noting  from another conversation in slack: We are going to make another ticket to We would probably want to configure Django storages for manifest storage. If we did that, will have better parity for local and deployed environments.

JeanMarie-TTS: Matt's fix does not resolve the problem that I'm seeing. I reverted my changes and tested again with Matt's fix but got the same error. I made my changes on top of Matt's fix, and the tests run fine.

JeanMarie-TTS:
Issue resolved in a different PR by changing docker settings

PR #739, opened 2023-01-19
Migrate Pre-SAC form to new single Django app

### LINKED ISSUE #638: Migrate Pre-SAC to Django app
Breakout issue of #563 (and subsequently #556)

### Background

The FAC-Frontend submitted data for the three steps of the Pre-SAC checklist to Django REST Framework. We need to migrate that over to the Django report_submission app.

- [ ]  We have a working Django application that can accept Pre-SAC form data and either send it via the existing DRF endpoints or enter it into the DB some other way, whichever is easier.
- [ ] No new CSS or JavaScript has been written, as that needs to be covered by the new SASS build pipeline.
- [x] You need to be authenticated to access the steps

### Tasks

- [x] Figure out the pre-SAC flow and how it would work in the Django app
- [x] Stand up step 1's form from the pre-SAC
- [x] Stand up step 2's form from the pre-SAC
- [x] Stand up step 3's form from the pre-SAC
- [ ] Get step 1 data to the DB
- [ ] Get step 2 data to the DB
- [ ] Get step 3 data to the DB

geekygirlsarah:
Fixes #638 

This merges frontend and backend of the Pre-SAC to work in the new single Django app.

What works:
* The flow through all the steps
* Saving data to the user profile along the way
* Creating the SAC database object in the end

What does not work or is in other issues:
* Resolving frontend error conditions (like server timeouts) as they weren't in the original frontend
* Stopping and resuming a pre-SAC and pulling previously saved info as it wasn't in the original frontend/backend
* Directing to the `/audit/<report_id>` page as that endpoint doesn't exist yet

This PR should be merged+squashed when it's done to reduce commit history.


tadhg-ohiggins:
Note that we're going to squash + merge this one, barring any last-minute appeals to save the individual commits.

tadhg-ohiggins (commented):
This looks good, but we need to add tests; if we think the frontend will be covered by Cypress, we can leave that for now, but we do need Python tests.

tadhg-ohiggins:
My [last commit]() includes the following:

*   Add end-to-end test for pre-SAC backend.
*   Add one rather large test that goes through all three steps.
*   Add some smaller tests.
*   Alter the `report_submissions/views.py` code to use if clauses after validating input rather than looking for non-existent exceptions.
*   Change the link on the `my_submissions` page to link to the old API JSON view of a SAC, so that we can at least see its contents easily during manual testing.
*   Remove the CSRF token from the data that gets stored in `general_information`.

The major remaining problem is that the backend code and the frontend code both do the work of moving the user through the process; the backend code sends redirects and the frontend code changes `window.location.href`. In-browser, the frontend code controls what the user sees, although both requests are made, such that at the moment it appears that every time a user submits one of these pages, two GET requests are sent to the server.

We should not have redundant code doing the same thing on the frontend and backend.

In my view the server-side redirects are what we should rely on, but rewiring the JavaScript to enable the forms and add/remove various other details was not something I had time to do.

Because the JavaScript is what currently “wins”, the Python views expect JSON as the payload; changing to more traditional form submission might require having them handle form payloads instead, which also might require tweaking the tests.

The tests in their current form aren’t ideal, but test coverage for `report_submission` is over 98% and I think refactoring them can probably wait.

Manual end-to-end testing of form submission works (at least locally…) and the tests pass (locally and in GitHub).

tadhg-ohiggins:
As noted in Slack, I think that we should fix the issue of the frontend and backend both trying to control the flow, and do so by having the backend handle it. However, if that proves to be too difficult to complete by EOD Monday, I think we should merge the working version even with that flaw, and create new issues to handle the required cleanup.

geekygirlsarah:
Filed issue #757 based off some testing found on this. With one approval and no objections on Slack, I'm merging.

PR #753, opened 2023-01-24
update login instructions for local dev

LindsayYoung:
Just noticed we did not update this

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -158,7 +158,7 @@ fac ()
 
 Let's use this workflow to create a `superuser` in our development environment so we can access the Admin interface!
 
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -158,15 +158,21 @@ fac ()
 
 Let's use this workflow to create a `superuser` in our development environment so we can access the Admin interface!
 
+First create an account in the Login.gov test environment, The best way to do that is to run [http://localhost:8000](http://localhost:8000) click on the log in link from your app running locally. (It needs to be localhost and not http://0.0.0.0:8000 to work with how we configured out Login.gov test account.)
```
timoballard, line 161:
```suggestion
First create an account in the Login.gov test environment, The best way to do that is to run [http://localhost:8000](http://localhost:8000) click on the log in link from your app running locally. (It needs to be localhost and not http://0.0.0.0:8000 to work with how we configured our Login.gov test account.) 
```

timoballard (approved):
one small typo, otherwise lgtm!

geekygirlsarah (approved):
With Tim's typo suggestion it looks good to me too!

PR #758, opened 2023-01-25
Config the zap scan 

LindsayYoung:
-I (do not return failure on warning)


geekygirlsarah (approved):
`I` see this change and approve it! `U` can merge it in now.

Issue #654, opened 2022-12-08
Implement My Audit Submissions design updates


melchoyce:
Updated version of #343 and #424.

Related: #537 

## Background

Since the initial implementation of the My Audit Submissions design, we've done two rounds of usability testing that have led to both content and design updates. We'll want to update the live website with the updated design.

![image](https://user-images.githubusercontent.com/23194720/206579010-eed70312-16e0-4c5a-9a3e-139c208d4b24.png)

## Acceptance Criteria (We'll know we're done when...)

- [x] The My Audit Submissions page has been updated with the most recent design

## Tasks

- [ ] TBD

JeanMarie-TTS:
@melchoyce ,  Should we not be designing this page for a time when Census FAC will be decommissioned? Should we be spending cycles to make sure that we link to relevant page on Census FAC?

melchoyce:
Hey! I was under the impression that we'd get the submission flow launched before the data dissemination flow, which is why this page includes a link to https://facdissem.census.gov/—has that plan changed?

mogul:
To demo: Link on the date rather than report ID, column order

PR #761, opened 2023-01-27
Add test cases , fix column ordering

JeanMarie-TTS:
Fixes a comment in #343 

Test cases to ensure login is required for views to be accessed.
Reorder columns and implement link on Date - https://www.figma.com/file/1bNncVt1TcFFDhQslqFC2o/FAC-Playground?node-id=4954%3A24137
Implement a placeholder view to edit a submission.
Place link to the temporary vies in the table.

PR #764, opened 2023-01-31
Design updates for audit submissions page

jasnakai:
Related to #654 which is a breakout of #343 

This PR concerns the design updates for the "My audit submissions" page.

What this PR covers:
- Updates to:
  - header text (See Image 1)
  - table caption (See Image 1)
  - column order (See Image 1)
  - "Start a new submission" header to "Create a new audit" (See Image 2)
  - Create a new audit text (See Image 2)
- Added link to old FAC site to table caption. (See Image 1)

(Image 1)
image

(Image 2)
image

Not covered in this PR:
- Links to edit submissions that are on the Fiscal period end dates points to a un-finalized URL.
  - These URLs need to be updated when the destination is created.
- Table data needs to be properly formatted.
  - Status (i.e. "in_progress" should be "In Progress")
  - Fiscal period end dates should be in MM/DD/YYYY format
- The Entity name data returns "None". This is not correct.

### DIFF
--- a/backend/audit/templates/audit/my_submissions.html
+++ b/backend/audit/templates/audit/my_submissions.html
@@ -3,7 +3,7 @@
 
 {% block content%}
 <div class="grid-container">
-    <h1>My audit submissions</h1>
+    <h1>Create or modify an audit</h1>
     {% if data.len == 0 %}
     <p>
         No submissions found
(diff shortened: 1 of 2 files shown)


JeanMarie-TTS:
@heymatthenry , @jasnakai We should merge PR #761 before this one since there are some table changes there as well.

jasnakai:
Sure. I think those changes are redundant so it shouldn't cause any major merge conflicts. That is why I was asking for a link to the AC for #761 to check if they were redundant.

Issue #746, opened 2023-01-23
Normalize data models


LindsayYoung:
Breakout issue of Data Migration Epic #723

### Background
We will want to optimize the data structure so that queries can be more performant.  It should also be more approachable for development.

Depends on #742 and #745

### Acceptance Criteria (We'll know we're done when...)
- [x]  We have a reasonably normalized data structure

LindsayYoung:
See progress here: https://github.com/GSA-TTS/FAC/pull/754

LindsayYoung:
We will always be open to improvements, but initial work is merged

Issue #745, opened 2023-01-23
Link the data between the models


LindsayYoung:
Breakout issue of Data Migration Epic #723

### Background
The initial Data schema has the relational keys from the census system, but we want to link the records with foreign keys, many to many look ups, etc. 

This depends on  there being data to link and test with. The 22 downloads  in  #744 should be sufficent.

### Acceptance Criteria (We'll know we're done when...)
- [x]  The models are linked appropriately, so you can stitch all the data together

LindsayYoung:
This ended up being easy, all other tables point to the gen table

Issue #742, opened 2023-01-23
Initial data structure


LindsayYoung:
Breakout issue of Data Migration Epic #723

### Task
Create simple data models with readable names and documentation based on the structure of the downloads and the download data key

### Acceptance Criteria (We'll know we're done when...)
- [x] There are models and fields for all the public data elements
- [x] The names are readable
- [x] The fields are documented so that  we can optimize the data structure without loosing track of the data source

LindsayYoung:
This was an intermeiatestep in the data migration epic, you can see the moels here:
LY/data exploration [WIP] #754

We discused this at eng huddle an considered this closed.

Issue #747, opened 2023-01-23
Modify the script that we can load all the public data


LindsayYoung:
Breakout issue of Data Migration Epic #723

### Background

Depends on new schema #746

### Acceptance Criteria (We'll know we're done when...)

- [x] Modify the public data load script so that it can load all the public data
- [x] Modify the public data load script so that it creates relationships as it loads

LindsayYoung:
I am going to close this ticket since the script is in use, I am going to make a new ticket to track loading.

PR #768, opened 2023-02-03
Ly/public data upload

LindsayYoung:
This is a work in progress and will still undergo changes. Adding it as a draft here for visibility.

This as a upload script that can handle 2022 uploads you can run it with` manage.py public_data_loader`. Read the documentation included in the PR for details 

Here are the [normalized models](https://github.com/GSA-TTS/FAC/blob/2cac4549e28874dfc04045e5ec4daa9875db1075/backend/data_distro/models.py)  I am working on.

This fulfills:
-  https://github.com/GSA-TTS/FAC/issues/742 The first pass at models reflected the  Census's data documentation to create a good way for us to tell where the data comes from. 
- Link the data between the models https://github.com/GSA-TTS/FAC/issues/745
- Normalize data models https://github.com/GSA-TTS/FAC/issues/746 

Note:
- This replaces https://github.com/GSA-TTS/FAC/pull/754. 
- The CORS update is to remove the /static path from the domain, It works as-is when deployed, but this should be correct and it was causing some validation errors that was keeping my task from running
- I am adding 2 type ignores
   - There is an empty list of dependencies for out initial migration, it's supposed to be empty so it doesn't need types
   - It is having trouble finding stubs for the parser for datutils, I am just going to ignore it because there isn't a problem with our code.

## Considerations for review:
- I think adding the task manifest would be helpful as a resource, but we can take it out
- This adds Pandas to the main dependencies. That seems kind of heavy, since this is not going to be run often, but we will need them for the tests.

# Data loading

We have a script to load the public data.

Download data from https://facdissem.census.gov/PublicDataDownloads.aspx
Then unzip the files and place the them in data_distro/data_to_load/

Load them locally with: `manage.py public_data_loader`

It's a lot of data, so you might just want to run it against one of the cloud.gov instances. You can do that with task here is [an example](https://github.com/GSA-TTS/FAC/blob/main/backend/manifests/task-manifest-staging.yaml)

We are doing an initial data load in February 2022. There will be another round of data upload when we get the private data from Census and a mapping to the PDF files of reports. We will need to update data before launch to grab any records that have been added since the initial download work.

The upload script should be flexible enough to handle changes in the models that may happen between data loads.

To refresh the data loading script:
1) Look at the `make_table_structure` command. Make sure things like new_fields are up to date.
2) Rerun the `make_table_structure` management command to produce `new_upload_mapping.json`.
3) Check the outputs to make sure that things make sense. You will need to do a manual check on the upload mapping, If a data column needs to go to more than one model, that needs to be addressed in the upload script. For example, AUDITOR_EIN has the same info reported on two of the upload tables, so you only need to upload it once. Then, replace `mappings/upload_mappings.json` with the new mapping.
4) Check the `public_data_loader` script. Renaming should mostly be handled by `upload_mapping` but changing relationships need to be accounted for in the script. Make sure the relationships that are in the `add_relational` output of `make_table_structure` are added in the script.
5) Run the script. Each phase should tell you the number of expected objects, for tables like general that should be 1:1 relationship for auditors, auditees and things that can be consolidated because of foreign to and many to many relationships, there will be fewer objects.


LindsayYoung:
I am going to continue to load data I am using #747 for that work, which can come next.

tadhg-ohiggins (approved):
Discussed and went over with Lindsay, this is still a work in progress but looks fine to merge as it's mostly distinct.

We will want to look at fac-dev once this builds and deploys to see if the settings.py changes fix the manifest 500 we were getting.

PR #769, opened 2023-02-03
Optionally disable OIDC authentication

### LINKED ISSUE #765: Create test authentication backend
### Background
In order to unblock #762, we will introduce a test authentication backend, which creates (if it doesn't already exist) a test user in the database and automatically authenticates as that user.

### Acceptance Criteria

- [ ] When the `DISABLE_AUTH` environment variable is true, the FAC app uses `FACTestAuthenticationBackend` as its default authentication backend
- [ ] When the `DISABLE_AUTH` environment variable is true, the FAC app establishes an authenticated session without any input from the frontend

timoballard:
Resolves #765 

When the `DISABLE_AUTH` environment variable is set to true, the OIDC authentication flows are disabled, and incoming requests are automatically authenticated as a test user.

### DIFF
--- a/backend/config/settings.py
+++ b/backend/config/settings.py
@@ -319,3 +319,14 @@ OIDC_PROVIDERS = {
 LOGIN_URL = f"{env_base_url}/openid/login/"
 
 USER_PROMOTION_COMMANDS_ENABLED = environment in ["LOCAL", "TESTING", "UNDEFINED"]
+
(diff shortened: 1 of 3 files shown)


heymatthenry (approved):
LGTM! But probably worth having someone who knows more about Django than I do have 👀 on it

tadhg-ohiggins (approved):

PR #774, opened 2023-02-06
Excel storage ADR

### LINKED ISSUE #773: Excel file storage ADR
We need an ADR summarizing our decision to store Excel files on the backend instead of doing bi-directional translation between our data models and Excel.

timoballard:
Resolves #773 

Co-authored with @tadhg-ohiggins

### DIFF
--- a/docs/architecture/decisions/0015-store-uploaded-excel-files.md
+++ b/docs/architecture/decisions/0015-store-uploaded-excel-files.md
@@ -1,0 +1,60 @@
+# 15. Store uploaded Excel files
+
+Date: 2023-02-06
+
+## Status
+
+Accepted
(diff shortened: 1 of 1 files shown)


mogul (approved):
All looks fine to me, assuming we can take on the ClamAV implementation as incremental development.

timoballard:
@mogul - yes, agreed. @tadhg-ohiggins and I took a first pass at writing out the tickets that will come out of this, and we have the ClamAV bit slotted in as its own ticket, which only gets layered in after the basic upload plumbing is in place and the simple file validation checks (file size, name, extension, etc) are implemented. We can go into those in detail during backlog grooming as well

PR #800, opened 2023-02-17
Basic Excel validation

### LINKED ISSUE #779: Implement basic Excel file validation
Breakout of #772 

### Background

When an Excel file is uploaded, we need to perform a set of basic validation checks before allowing the file to be stored in S3. This ticket will focus on validating the user-provided filename, the file extension and content type, and the file size. Infection scanning and content validation will be covered in separate tickets.

These validations should most likely be implemented as field validators on the (ExcelFile)[#777]'s `FileField`. [See a similar implementation for the USDOJ-CRT Portal](https://github.com/usdoj-crt/crt-portal/blob/e96d4df9d4bd3c1245a31825ac9cf9dee515a912/crt_portal/cts_forms/models.py#L472).

### Acceptance Criteria
- [x] When a file is uploaded that has an invalid filename (it cannot be slugified and/or it has no extension), the upload is rejected and a descriptive error message is returned
- [x] When a file is uploaded with a file extension other than xls or xlsx, the upload is rejected and a descriptive error message is returned
- [x] When a file is uploaded with a [MIME type other than those for xls/xlsx files](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types), the upload is rejected and a descriptive error message is returned
- [x] When a file is uploaded with a file size that exceeds the maximum allowed file size, the upload is rejected and a descriptive error message is returned.
- [x] There are tests covering the above scenarios

timoballard:
Resolves #779


tadhg-ohiggins (approved):
Looks good!

PR #810, opened 2023-02-17
Make first draft of  sequence diagram

JeanMarie-TTS:
Took a first cut at creating a sequence diagram. Need collaborators and reviewers to get this to completion.
Also need help creating a png file as something is broken on my laptop.


tadhg-ohiggins:
Both auditee and auditor are likely to make changes to the Excel files, so both of those should reflect pulling down the Excel files and then uploading them. The auditors do not only submit findings, but a lot of the other info as well. Who submits what info appears to depend a lot on the individual user(s).

Certification must be done both by someone from the auditor and someone from the auditee; this diagram makes it seem as though only the auditee performs that action, which doesn't match my understanding.

`s/Agebcy/Agency` throughout.

For the agency users, I'm not clear on the direction of the arrows; if the fourth user in the user list is the agency user, shouldn't access start from them rather than from the agency? And in either case, shouldn't line one (“accesses FAC”) and line seven (“returns API results”), as well as some of the others, end up going to the same place?

JeanMarie-TTS:
@tadhg-ohiggins , the typo caused the agency access to be rendered incorrectly. 
My assumption is that the auditor makes submissions on behalf of a grantee and that it should suffice to show the grantee actions for this case.
Is there a generic term to denote what the auditor submits? I used 'Findings', but we need something broader.
I need to understand better what the auditor certifies.
Will make another commit now so you can comment on my changes.

docs/architecture/diagrams/src/FAC_Sequence.puml
```diff
@@ -0,0 +1,61 @@
...
+
+autonumber 
+== Submission ==
+Grantee -> FAC : Grantee Accesses FAC
+FAC -> LoginGov : Redirects for authentication
+LoginGov -> FAC : Redirects to FAC
+Grantee -> FAC : Fetches Submission
+DB -> FAC : Fetches previously saved structured data
+S3 -> FAC : Fetches previously saved xlsx docs
+FAC -> Grantee : Mahed previous submission available for editing
```
tadhg-ohiggins, line 22: s/Mahed/Makes

docs/architecture/diagrams/src/FAC_Sequence.puml
```diff
@@ -0,0 +1,61 @@
...
+Database "S3" as S3
+
+autonumber 
+== Submission ==
+Grantee -> FAC : Grantee Accesses FAC
+FAC -> LoginGov : Redirects for authentication
+LoginGov -> FAC : Redirects to FAC
+Grantee -> FAC : Fetches Submission
+DB -> FAC : Fetches previously saved structured data
+S3 -> FAC : Fetches previously saved xlsx docs
```
tadhg-ohiggins, line 21: “docs” here should be “documents”, or “documents” elsewhere should be “docs”, for consistency.

tadhg-ohiggins (changes requested):

tadhg-ohiggins (approved):

PR #814, opened 2023-02-22
Excel file infection scanning

### LINKED ISSUE #780: Implement Excel file infection scanning
Breakout of #772 

### Background
Before ingesting and storing uploaded Excel files, we need to perform an infection scan using an antivirus tool. As outlined in a recent [ADR](https://github.com/GSA-TTS/FAC/blob/main/docs/architecture/decisions/0010-file-uploads.md), we will use [ClamAV](https://www.clamav.net/) to perform these scans.

We can refer to a similar setup for the USDOJ-CRT Portal where applicable:
- [Local development configuration](https://github.com/usdoj-crt/crt-portal/blob/e96d4df9d4bd3c1245a31825ac9cf9dee515a912/docker-compose.yml#L56-L62)
- [Cloud.gov configuration](https://github.com/usdoj-crt/crt-portal/blob/e96d4df9d4bd3c1245a31825ac9cf9dee515a912/manifest_dev.yaml#L23-L32)

#### Notes
When adding ClamAV to the USDOJ-CRT Portal, we needed to significantly increase our memory allocation, as ClamAV relies on very large in-memory databases of virus definitions to do its work

### Acceptance Criteria

- [x] The local development configuration is updated to use the [clamav-rest](https://hub.docker.com/r/ajilaag/clamav-rest) docker image as a service
- [x] All cloud.gov environment configurations are updated to use  [clamav-rest](https://hub.docker.com/r/ajilaag/clamav-rest) as a bound service
- [x] There is an additional field validator for the [ExcelFile](#777) model that performs an infection scan using ClamAV
- [x] If an infection is found, the upload is rejected and a descriptive error message is returned to the user.
- [x] If the ClamAV service is unavailable or returns an internal error, the upload is rejected and the user is instructed to try again at a later time
- [x] There are tests covering the above scenarios

timoballard:
Resolves #780 

https://user-images.githubusercontent.com/1425377/220705245-836a06f4-2d19-43e0-8632-f5b9e518408c.mov

### DIFF
--- a/backend/audit/test_validators.py
+++ b/backend/audit/test_validators.py
@@ -539,6 +539,7 @@ class ExcelFileInfectionValidatorTests(TestCase):
 
     @patch("requests.post")
     def test_av_service_unavailable(self, mock_post):
+        """If the AV service is unavailable, the file should not pass validation"""
         mock_post.side_effect = requests.exceptions.ConnectionError(
             "service unavailable"
         )
@@ -548,6 +549,7 @@ class ExcelFileInfectionValidatorTests(TestCase):
 
     @patch("audit.validators._scan_file")
     def test_validation_fails_on_av_service_error(self, mock_scan_file):
+        """If the AV service returns an internal error, the file should not pass validation"""
         mock_scan_file.return_value = MockHttpResponse(500, "error text")
 
         with self.assertRaises(ValidationError):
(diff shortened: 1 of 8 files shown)


tadhg-ohiggins:
@mogul Can you coordinate with Tim on setting ClamAV up with Terraform?

mogul:
☝️That last commit oughta do it. (Bonus: By using [auto-magical port 61443](https://docs.cloudfoundry.org/concepts/understand-cf-networking.html#securing-traffic), traffic to the ClamAV REST API will happen over TLS.)

mogul (approved):

tadhg-ohiggins (approved):

mogul:
Huh, I'm not sure this deployed right... The ClamAV app is deployed and the internal route is in place, but it's running zero instances. 🤔 
image

Investigating...

mogul:
Still not clear why it deployed zero instances, but there was another bug that's fixed here:
https://github.com/GSA-TTS/FAC/pull/820

PR #822, opened 2023-02-27
ADR for decision to switch away from using OSCAL/Trestle

tadhg-ohiggins:
So we must postpone
the glorious future of
automated docs

-----

ADR for deciding to not use OSCAL/Trestle for compliance documentation. Supersedes [ADR 9](https://github.com/GSA-TTS/FAC/blob/main/docs/architecture/decisions/0009-compliance-documentation.md).


LindsayYoung (approved):
👍

PR #827, opened 2023-02-28
Set REQUESTS_CA_BUNDLE in cg manifests

timoballard:
Relates to #780 

[Sets `REQUESTS_CA_BUNDLE` to allow validation of cloud.gov-provided certs](https://cloud.gov/docs/management/container-to-container/#addressing-certificate-validation-errors)


mogul (approved):

mogul (commented):
If this isn't something you actually expect to be different in different environments (meaning the value will always be the same), it might be better to just set it in `.profile`.

Issue #556, opened 2022-10-27
Figure out how to transfer frontend to Django


tadhg-ohiggins:
Related to #553 

### Background

As part of shifting to a more traditional Django web application, we need to move the frontend over to Django. This is a placeholder issue for trying out and iterating on different approaches for this.

mogul:
@tadhg-ohiggins is going to schedule a meeting to fill in a list of other issues as they come up.

PR #829, opened 2023-03-01
General Information form

jasnakai:
Breakout issue of https://github.com/GSA-TTS/FAC/issues/556

This issue covers standing up very basic versions the General Information form in Django. This includes some hack to generate the pre-SAC information in a way that makes it possible to start the General Information entry process.

What this PR covers:
- We have a working Django application that can accept General Information form data, validate it against the JSON schema, and update the relevant SingleAuditChecklist in the database
- No new CSS or JavaScript has been written, as that needs to be covered by the new SASS build pipeline.

Not covered in this PR:
- Figure out the pre-SAC flow we need to work around needing to port the whole existing pre-SAC process over.
  - Handled in previous ticket.
- Front end error handling and display for form submission/validation errors. Placeholder ticket for that work is [here.](https://github.com/GSA-TTS/FAC/issues/937)


JeanMarie-TTS:
@jasnakai , there was a decision to implement templates centrally rather than within django apps (dev huddle about a month ago). I believe all you need to is:
`mv report_submission/templates/report_submission templates`

tadhg-ohiggins (approved):

Issue #751, opened 2023-01-23
Link existing data to PDFs via URL


LindsayYoung:
Breakout issue of Data Migration Epic #723 

### Background

Get PDFs and add them to buckets. 

We can use the filename to generate a url point to the bucket, we would want to incase the bucket name changes. We also don't want the url in the data because it will be different in prod, stage, and dev. We may want to revisit this when we look at #748 (We may want to pregenerate production uls for speeder downloads.)

Linking the PDF file names to  reports can be done by looking at the name of the file - {DBKEY}{AUDITYEAR}.pdf

### Acceptance Criteria (We'll know we're done when...)

- [ ]  We have public PDFs in predictable urls in our public bucket
- [ ]  We have private PDFs in predictable urls in our private bucket

### Tasks

- [ ] Sort PDFs into public and private buckets
- [x] Create private bucket for private PDFs
- [ ] Add public PDFs to the public bucket the structure could be something like `/collections/year/month/<file name>.pdf` (Buckets don't like a lot of sequentially numbered file names one bucket, so trying to break them up into chunks that will be performant.)
- [ ]  Add private PDFs to the private bucket in a file structure that mimics the public PDFs

Reach goal:
- [ ]  Add some test PDFs for local development

jadudm:
We have, at this point, a single private S3 bucket. We are going to route everything through the app.

E.g. 

`https://app.fac.gov/pdf/...`

will link to a file, and we'll determine in the app/do a redirect if appropriate.

PR #840, opened 2023-03-07
LY/data-load-improvements

LindsayYoung:
- adding support for UEI table upload (only 2022)
- loosen cpa validation
- rename one boolean field list item
- loading to new db (Note: remove before merge)
- ads an array field for the urls of report pdfs


tadhg-ohiggins (approved):

backend/data_distro/migrations/0028_general_pdf_urls.py
```diff
@@ -0,0 +1,26 @@
...
+        ("data_distro", "0027_null_for_auditee_auditee_contact"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="general",
+            name="pdf_urls",
+            field=django.contrib.postgres.fields.ArrayField(
+                base_field=models.CharField(
+                    max_length=400,
```
jadudm, line 18: I suppose it's an easy change later, but do we want to set a higher max length (e.g. 1024) for the URL? Max object key lengths in S3 are 1024 characters, and I think URLs are a max of 2048 (in general)?

jadudm: Or, I might misunderstand this field. In which case, disregard.

jadudm (approved):
I left a comment, but feel the overall commit can go ahead with any answer (change or disregard) to the one question I had.

backend/data_distro/migrations/0028_general_pdf_urls.py
LindsayYoung: My proposal is the that PDF urls should be {AWS bucket}/collections/year/month/<file number>.pdf

My guess for that our PDF URLS would would be about 150 characters  (80 char for the bucket, 25 for the path, ~15 for the file name) so 400 gives a lot of future flexavility. But realistically,  if we are making URLS longer than 400 chars we are not giving them very good names (IMHO.)  It may be better to proxy them through the app to have shorter permalinks at a .gov address that don't change if we change buckets, but that would be a later improvement.

here is the PDF ticket that is from https://github.com/GSA-TTS/FAC/issues/751

PR #850, opened 2023-03-14
Configure e2e tests to run in GH actions

heymatthenry:
What it says on the tin. This PR configures the end-to-end tests to run in Github Actions as part of our CI toolchain.


geekygirlsarah:
I think it looks good but I have no idea on why it's timing out on check 1. I'll approve if we can figure this out next week!

heymatthenry:
@geekygirlsarah I just re-ran the suite and it's green now. One of the bummers about Cypress running in CI is that it does sometimes give heisenbug failures. In a perfect world, we'd also set up an S3 bucket somewhere to keep artifacts like videos of test runs to see what actually happened in cases like this.

Issue #622, opened 2022-11-30
Decide whether Pa11y is required along with Lighthouse


tadhg-ohiggins:
### Background

As part of our app unification we need to move our Pa11y setup to run in Docker in our new configuration.

### Acceptance Criteria (We'll know we're done when...)

- [ ] 
- [ ] 

### Tasks

- [ ] Write an ADR
- [ ]

pkigongo:
Since we have Lighthouse, not sure if we need Pa11y

jadudm:
Per conversation, we're going to keep both (as they are not true sub/supersets of each other for coverage), and document that in an ADR. Assigning to me.

jadudm:
Again, I don't know if this is ADR-level. We decided to keep two different checkers, because they cover most of the same things, but each catch slightly different things in practice. If we decide we need an ADR, we can come back.

PR #862, opened 2023-03-16
Bolster automated accessibillity testing

heymatthenry:
This PR adds more accessibility testing to the app. More specifically, it:

- Ensures `lighthouse-ci` runs against all existing pages
- Adds `pa11y-ci` to the CI pipeline
- Fixes a handful of accessibility issues identified in the preceding steps

This also probably moots #622

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,7 +80,7 @@ jobs:
         working-directory: ./backend
         run: npm run check-all
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     env:
       ENV: TESTING
       SAM_API_KEY: ${{ secrets.SAM_API_KEY }}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,5 +113,7 @@ jobs:
         run: |
           npm install -g @lhci/cli@0.8.x
           lhci autorun
+      - name: run pa11y
+        run: npx pa11y-ci
   validate-terraform:
      uses: ./.github/workflows/terraform-lint.yml
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,6 +114,8 @@ jobs:
           npm install -g @lhci/cli@0.8.x
           lhci autorun
       - name: run pa11y
-        run: npx pa11y-ci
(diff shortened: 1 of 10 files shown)


heymatthenry:
Setting this back to draft so I can figure out what's going on with the flaky specs failing in CI but not locally

backend/templates/base.html
```diff
@@ -9,8 +9,14 @@
             href="{% static 'compiled/scss/main-post.css' %}"
             media="all"
             type="text/css"/>
+        <title>Federal Audit Clearinghouse</title>
+
         {% block metatags %}
-            <title>{{ post.title }}</title>
+            <!--
```
LindsayYoung, line 15: We could create a title block so it there could be a default, but we can overwrite it on child pages

heymatthenry:
@LindsayYoung I added a title block to the base template as well as commands to run the a11y tests manually and docs about about to do so.

docs/testing.md
```diff
@@ -61,6 +61,13 @@ We use a combination of [Lighthouse-ci](https://github.com/GoogleChrome/lighthou
 
 Accessibility tests are executed as part of our CI/CD pipeline on each PR to the `main` branch, commit to the `main` branch, and PR into the `prod` branch.
 
+To run Lighthouse or pa11y locally, just run:
```
LindsayYoung, line 64: I don't think this runs in the docker container because the config file is above the /backend directory. For now, I think that shouldn't block the pr.  You can accept this if that's right, if could you help me trouble shoot locally?
```suggestion
To run Lighthouse or pa11y locally, install the dependencies on your machine and outside the docker container, run the following commands.
```

LindsayYoung (approved):

.github/workflows/test.yml
```diff
@@ -90,8 +90,8 @@ jobs:
       ALLOWED_HOSTS: '0.0.0.0 127.0.0.1 localhost'
       DISABLE_AUTH: False
```
LindsayYoung, line 91:
```suggestion
      DISABLE_AUTH: True
```

heymatthenry:
The app has moved forward enough that this PR should just be closed. #1206 and #1207 capture the work that should go into getting this all working.

PR #864, opened 2023-03-17
Story template includes exit criteria checklists

mogul:


pkigongo (approved):
Approved. Thanks, Bret

Issue #778, opened 2023-02-07
Implement Excel file storage


timoballard:
Breakout of #772 

### Background

We need to configure our Django app to use our private S3 bucket as its default storage mechanism.

The work for this ticket should focus solely on setting up the file upload and storage pipeline. File validation and scanning will be covered in a separate ticket.

### Acceptance Criteria

- [x] The app is configured to use our private S3 bucket as its default file storage mechanism
- [x] The local development configuration is updated to use `localstack` to emulate S3
- [ ] The local development documentation is updated accordingly
- [ ] When an ExcelFile is added via Django admin, the file is stored in the private S3 bucket and not the local filesystem

pkigongo:
@geekygirlsarah - Please leave artifacts in comments to this issue. Thanks

geekygirlsarah:
> geekygirlsarah - Please leave artifacts in comments to this issue. Thanks

I put them in the PR since that's where I'm used to putting them on past teams. Copying here for reference.


<img width="893" alt="image" src="https://user-images.githubusercontent.com/2601974/228935453-90055c09-404a-4aad-abc3-e49c3ea1de2a.png">

<img width="587" alt="image" src="https://user-images.githubusercontent.com/2601974/228935655-8d528f75-86b0-4393-8b7d-1856ded4d4a8.png">

<img width="1464" alt="image" src="https://user-images.githubusercontent.com/2601974/228936003-8d90cf2f-23ac-4a8a-8782-e2315a7de503.png">

PR #866, opened 2023-03-17
Update docs for LocalStack, linting, and a header correction

geekygirlsarah:
Minor development documentation updates. Partly related to #778 but found other things along the way.


docs/development.md
```diff
@@ -142,6 +143,14 @@ python -m pip install --upgrade pip
 pip install pip-tools
 ```
 
+### LocalStack
+
+LocalStack is an AWS emulator. We use it to emulate S3 for storing files that users upload.
+
+It will set itself up correctly if you run the app with `docker-compse up`.
```
timoballard, line 150: typo on "compose"

timoballard (approved):
👍

docs/development.md
geekygirlsarah: Fixed!

Issue #733, opened 2023-01-12
Tweaks to sign out button


heymatthenry:
#728 will implement the sign-out button, but the button will need some additional work that is out of scope for the ticket associated with that PR (#343): 

![Screenshot showing signout button with a long identifier rather than a user's name](https://user-images.githubusercontent.com/6290/211869295-4b7bfc63-beb0-4cb9-9201-ce7235e5ffd5.png)

In particular, the button needs to show the user's name, and the button's design and copy should match the mockups. 

[:eyes: [Figma](https://www.figma.com/file/1bNncVt1TcFFDhQslqFC2o/FAC-Playground?node-id=2443%3A11490)]

The design was already implemented in the FAC-Frontend app so it should be straightforward to move the [HTML](https://github.com/GSA-TTS/FAC-Frontend/blob/99fc9ba204f005d233ff943145a1fc3cb3429ab5/src/_includes/header.njk#L81) and [CSS](https://github.com/GSA-TTS/FAC-Frontend/blob/99fc9ba204f005d233ff943145a1fc3cb3429ab5/src/scss/_header.scss#L40) over.

## Acceptance criteria 

- [x] The sign-out button displays the user's name
- [x] The button copy matches the design
- [x] Visual design is consistent with the Figma mockup

mogul:
@jperson1 want to drop a screenshot of the fixed version here for easy reference during our upcoming sprint review?

jperson1:
Now, when first/last name is present:
<img width="346" alt="image" src="https://user-images.githubusercontent.com/91098850/228595268-f96b790a-557b-49c2-8ede-b96db1151bc2.png">

When first/last name is _not_ present, email displays:
<img width="403" alt="image" src="https://user-images.githubusercontent.com/91098850/228594714-03cbcae9-316e-41b5-b83a-daaaa8392824.png">

Compared against current Figma designs:
<img width="731" alt="image" src="https://user-images.githubusercontent.com/91098850/228595554-d8019ac1-aaec-434b-9ee2-570b7e2354c1.png">

jadudm:
This was done a long time ago, and demoed.

PR #876, opened 2023-03-22
Sign-out button tweaks

jperson1:
PR for issue #733 - tweaks to the sign out button.

Sign out button style should match [the Figma designs](https://www.figma.com/file/1bNncVt1TcFFDhQslqFC2o/FAC-Playground?node-id=2443%3A11490).

Button will display "Sign out" followed by the users first and last name, if it is present. Otherwise, it displays their email.

Also includes djlint automatic changes on a few HTML files.

Examples:
image
image

Edits: Further details and example images

### DIFF
--- a/backend/static/scss/_header.scss
+++ b/backend/static/scss/_header.scss
@@ -21,7 +21,7 @@
 .sign-in,
 .sign-out {
   background-color: color('primary-darker');
-  padding: 0.5em 2.5em 0.5em 1em;
+  padding: 0.5em 1.5em 1.5em 1em;
 }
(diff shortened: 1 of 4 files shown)


jperson1:
> I think this is pretty close. In comparing my local to Figma, it looks like we could add a little padding between the icon and the text. I think that would be nice if it is an easy fix, if it is a bigger lift I would defer to the team on how to prioritize.

I've added a right margin to the icon - see the example images above!

LindsayYoung (approved):
Thanks!

PR #886, opened 2023-03-24
Enable build pipeline caching

geekygirlsarah:
_Note: I won't be around to merge this... when it's good and everyone's happy, merge it in!_

This adds caching of different steps in the build process to try to save time off of the build, test, and verification processes.

* Add npm caching, which speeds up installs for frontend and lint actions
* Add pip caching, which does the same
* Add Docker layer caching, which speeds up builds of the main app container
* Moves the pull of localstack, postgres, and clamav to separate the step out a bit better so those aren't cached (they'd pull faster from Docker Hub)
* Replace `make docker-test` back with its instructions to not repeat steps, but also to force steps that need caching to be cached*
* Add npm and pip caching to the deploy process, using the same caches as the PRs were built on

This seems to have shaved ~5 minutes off of the test process, and frontend-testing and linting both run in under a minute.

(The same kind of Docker optimizations can be done to the `end-to-end-test.yml` file to speed up its Docker build, since it uses the same images/layers that testing does. Adding that in should shave minutes off of it as well, whenever it's turned back on.)

* It might be possible to put the command back, but this would need to be tested more and I ran out of time to really do it since this was added after I started try # 1 of this PR.

### DIFF
--- a/.github/workflows/deploy-apps.yml
+++ b/.github/workflows/deploy-apps.yml
@@ -17,10 +17,31 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Restore npm cache
+        uses: actions/cache@v3
+        id: cache-npm
+        with:
+          path: ~/.npm
+          key: fac-build-npm-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            fac-build-npm-
+            fac-build-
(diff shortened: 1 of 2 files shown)


LindsayYoung (approved):
It's great! I used it when I was trying to fix some unrelated deploy issues this morning. Thanks again!

PR #893, opened 2023-03-27
.yaml -> .yml

tadhg-ohiggins:
No a in yaml
Because of consistency
Filenames can be hard

-----

Rename manifest files for staging to be `.yml`. instead of `.yaml` so as not to confuse our deployment scripts.


LindsayYoung (approved):
thanks!!

PR #896, opened 2023-03-27
Grant Product Lead the Maintainer role for fac-team

mogul:


mogul:
@tadhg-ohiggins I added a check to the offboarding template after your review. ;)

PR #899, opened 2023-03-27
Deployment: try using cloud-gov/cg-cli-tools instead of 18f/cg-deploy-action

tadhg-ohiggins:
Mystery failure,
So try a different action?
Maybe worth a try.

-----

https://github.com/GSA-TTS/FAC/actions/runs/4536362781/jobs/7994723814 kept failing with
```
dpkg-deb: error: 'cf-cli_amd64.deb' is not a Debian format archive
dpkg: error processing archive cf-cli_amd64.deb (--install):
 dpkg-deb --control subprocess returned error exit status 2
Errors were encountered while processing:
 cf-cli_amd64.deb
Error: Process completed with exit code 1.
```

So we're trying switching to a more recent action.

### DIFF
--- a/.github/workflows/deploy-apps.yml
+++ b/.github/workflows/deploy-apps.yml
@@ -28,7 +28,7 @@ jobs:
         working-directory: ./backend
         run: npm run build
       - name: Update service keys
-        uses: 18f/cg-deploy-action@main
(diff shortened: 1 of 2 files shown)


LindsayYoung:
replaced by #901

PR #910, opened 2023-03-30
Use tasklists in issue templates

mogul:
We want to encourage the use of task lists from the time that story issues are created so that it's easy to break ACs and tasks out into sub-issues as needed without losing hierarchy.

(Note the preview of these changes won't properly render the tasklist in the PR context, but I verified the markup in an existing issue to ensure it's correct.)

### DIFF
--- a/.github/ISSUE_TEMPLATE/custom.md
+++ b/.github/ISSUE_TEMPLATE/custom.md
@@ -1,22 +1,14 @@
 ---
 name: Basic task tracking issue
-about: Starting point for new issues tracking project tasks
+about: Starting point for new issues tracking general tasks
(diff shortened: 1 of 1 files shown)


jadudm:
Unfortunately, the iterations we did on the story ticket moved past this PR, but the custom.md changes still make sense. Thoughts?

mogul:
I dropped the commit that modified the story template, rebased on main, and did a force push to this branch to limit it to just the non-story-template changes.

jadudm:
LGTM. Thank you.

jadudm (approved):
We did the iteration on the story template in another PR. LGTM.

PR #923, opened 2023-03-31
Tadhg/dump context

tadhg-ohiggins:
test context printout


.github/workflows/triggers.yml
jadudm, line 1: If this kind of debugging/dump behavior is desired, then possibly having a `debug/` branch/trigger might make more sense going forward than a name-based trigger. Approving so as to move this out of the PR queue.

PR #935, opened 2023-04-03
Jadudm/improve dev docs

jadudm:
The changes in this PR are as follows:

1. The dev docs are improved, so that someone can work straight through them and end up with a local docker stack that works and passes tests.
2. A new requirement to the dev requirements was added, as the web container would not run correctly with out it. I alphabetized the requirements lists while I was at it.
3. Notes about local development were moved into a separate file, as this is not a well-supported method of development on the team. If we want to keep those directions alive, we can maintain them in their own file.
4. Added an ignore for Census .txt files in the `data_distro` folder, because... otherwise, they want to be committed, and they should not be.


docs/development.md
```diff
@@ -99,162 +99,174 @@ An application and database are configured in [../backend/docker-compose.yml](..
...
 
-We use environment variables to configure much of how Django operates, at a minimum you'll need to configure the uri of your local database.
 
-Set a `DATABASE_URL` environment variable with the uri of your local database
-    *  `postgresql://[userspec@][hostspec][/dbname]`
+```shell
+docker-compose run web python manage.py public_data_loader -y 20
+```
 
+which will load the data from 2020 into your database. This is slow. Grab a cup of coffee, sit back, and watch the blinkenlights.
```
LindsayYoung, line 193:
```suggestion
See full documentation for loading data and keeping script up to date in [data_loading.md](https://github.com/GSA-TTS/FAC/blob/main/docs/data_loading.md). 
```

docs/development.md
```diff
@@ -99,162 +99,174 @@ An application and database are configured in [../backend/docker-compose.yml](..
...
 ```
 
-As a convenience, you can create an alias in your shell following this or a similar pattern
-```shell
-fac ()
-{
-  docker compose run web python manage.py ${@}
-}
-```
+Now, you can open [http://localhost:8000/admin](http://localhost:8000/admin) in your browser.
```
LindsayYoung, line 212:
```suggestion
Now, you can open [http://localhost:8000/admin](http://localhost:8000/admin) in your browser. (Use local host and not 0.0.0.0, to work with local login.gov auth.)
```

LindsayYoung (changes requested):
I left a few extra hints as suggestions

jadudm (commented):
Suggestions accepted.

docs/development.md
jadudm: I think I resolved this.

jadudm (commented):
All changes reviewed, accepted.

jadudm:
I literally have no idea if I have accepted or resolved the changes @LindsayYoung suggested. I tried to.

LindsayYoung (approved):
I see my suggestions

PR #942, opened 2023-04-04
#854 Added hassandeme.mamasambo to the list of developers

sambodeme:
[#854]()

## Description
This is just to add myself to the list of developers.


jadudm (approved):
Does what it says on the tin. Thank you!

Issue #563, opened 2022-11-01
First pass at General Information form in Django app


tadhg-ohiggins:
Breakout issue of #556

### Background

The current frontend submits data to Django REST Framework endpoints; we think that having the Django version do the same makes sense, but we need to explore that to see if there are hidden obstacles.

This issue covers standing up very basic versions the General Information form in Django. This includes some hack to generate the pre-SAC information in a way that makes it possible to start the General Information entry process.

### Acceptance Criteria (We'll know we're done when...)

- [ ]   We have a working Django application that can accept General Information form data and either send it via the existing DRF endpoints or enter it into the DB some other way, whichever is easier.
- [ ]    No new CSS or JavaScript has been written, as that needs to be covered by the new SASS build pipeline.

### Tasks

- [x]   Figure out the pre-SAC flow we need to work around needing to port the whole existing pre-SAC process over.
- [x]   Stand up a form that can accept the General Information data and get it into the DB.

geekygirlsarah:
@jasnakai What I've learned from chats:
* We should probably put templates/HTMLs in `/backend/report_submission/templates/*`
* Views would also go under `/backend/report_submission/views.py`
* Probably we'd want to take out the API calls and just move them over to Django, with submission via POST to Django, and returning HTML instead of an API call then just GETting to a new page 
* We need to keep CSS/JS, but can delete some JS if needed
* Views will pass over data to the templates to render (as opposed to being filled in via API calls)

I'm happy to show you any of this tomorrow if you want before/after our big 1/2 day meeting.

pkigongo:
Waiting on #638

melchoyce:
Latest mockup:

![image](https://user-images.githubusercontent.com/2846578/214883919-613852c0-54f4-4c09-93bf-0bbaadc419f9.png)
https://www.figma.com/file/1bNncVt1TcFFDhQslqFC2o/FAC-Playground?node-id=4954%3A24912&t=ILmPWtlO7zE5ea1b-4

The side nav here is for within this page, since it's so long, and some of the screens after it change depending on the type of audit you're filing.

Related: #185

pkigongo:
Per Jason "Need to pair with backend dev to get the general info form to submit."

JeanMarie-TTS:
[...]

PR #947, opened 2023-04-05
Pre-SAC serialization bugfix

timoballard:
Relates to #563 

#829 implemented the General Information form, but in so doing introduced a serialization issue for two of the pre-SAC boolean fields (`met_spending_threshold` and `is_usa_based`). These were being stored in the `UserProfile` as strings, which caused a downstream failure when the general information JSON document is checked against the JSON schema, which expects them to be boolean types.

### DIFF
--- a/backend/api/views.py
+++ b/backend/api/views.py
@@ -48,7 +48,7 @@ def eligibility_check(user, data):
         next_step = reverse("api-auditee-info")
 
         # Store step 0 data in profile, overwriting any pre-existing.
-        user.profile.entry_form_data = data
+        user.profile.entry_form_data = serializer.data
         user.profile.save()
         return {"eligible": True, "next": next_step}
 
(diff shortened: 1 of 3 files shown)


backend/report_submission/forms.py
```diff
@@ -22,7 +22,6 @@ class GeneralInformationForm(forms.Form):
     auditee_contact_title = forms.CharField()
     auditee_phone = forms.CharField()
     auditee_email = forms.CharField()
-    user_provided_organization_type = forms.CharField()
```
timoballard, line 25: this field is not actually submitted as part of the general information form, so we're removing it from the expected fields here

backend/api/views.py
```diff
@@ -48,7 +48,7 @@ def eligibility_check(user, data):
         next_step = reverse("api-auditee-info")
 
         # Store step 0 data in profile, overwriting any pre-existing.
-        user.profile.entry_form_data = data
+        user.profile.entry_form_data = serializer.data
```
timoballard, line 51: we should store the serialized data instead of the raw data, as the serialized data will have done the appropriate type conversions (e.g. `"True" --> True`)

tadhg-ohiggins (approved):

PR #952, opened 2023-04-05
Add .local/ to .gitignore

tadhg-ohiggins:
We track many things,
But we don’t need everything.
So, add `.local`.

-----

Add `.local/` to `.gitignore` so that we have a place to put files that we want to keep locally but don’t want to check in and don’t want to track one-by-one in `.gitignore`.


danswick (approved):
Looks good!

Issue #919, opened 2023-03-30
Exiting audit creation


jadudm:
# Exiting audit creation

When creating an audit (the Pre-SAC workflow), an auditor or auditee may decide they do not want to create an audit at this time. They should be taken back to the homepage when they decide to *cancel*. 

Breakout issue of https://github.com/GSA-TTS/FAC/issues/395

Replaces issue https://github.com/GSA-TTS/FAC/issues/430

# Acceptance Criteria

We use [behavior-driven development](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications) wherever possible.

[comment]: # "ACs should be clearly demoable/verifiable whenever possible. Try specifying them using behavior driven development."

[comment]: # "**Blurb** : **Given** **and** **when** **then**"

### Scenario: I am an auditee starting the submission process.

 **Given** that I have started the audit process, **when** I click *cancel*, ...
```[tasklist]
### then...

- [x] A modal alert is thrown indicating I will loose all progress (OK/I want to keep going)
- [x] I am returned to the "My Audit Submission Page" after I click *cancel*.
- [x] All state is cleared from the browser for the Pre-SAC workflow.
- [x] I remain logged in.
- [x] Content has revised the "cancel button" language and those changes are reflected in the implementation.
```

@jperson1 is the issue shepherd.

# Background

#430 was the less-descriptive, long-standing ticket for this work.

---

# Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))

There are no security concerns to exiting the Pre-SAC workflow.

# Sketch

[Notes or a checklist reflecting our understanding of the selected approach]

- [ ] Design designs all the things
- [ ] Engineering engineers all the things

# Definition of Done

### If not likely to be important in the next quarter...
- [ ] Archived from the board

### Otherwise...

- [ ] Has a clear story statement
- [ ] Design or Engineering accepts that it belongs in their respective backlog

## Design Backlog

- [ ] Has clearly stated/testable acceptance criteria
- [ ] Meets the design Definition of Ready [citation needed]

## Design In Progress

- [ ] Meets the design Definition of Done [citation needed]

## Design Review Needed

- [ ] Necessary outside review/sign-off was provided

## Design Done

- [ ] Presented in a sprint review
- [ ] Includes screenshots or references to artifacts

### If no engineering is necessary
- [ ] Tagged with the sprint where it was finished
- [ ] Archived

## Engineering Backlog

[...]

PR #953, opened 2023-04-05
Exit Audit Creation

jperson1:
PR for issue https://github.com/GSA-TTS/FAC/issues/919 - Exiting audit creation

Implements functionality for the cancel button in the audit submission process, as well as tests to confirm it works. Cancelling brings up an "Are you sure?" modal, and choosing "OK" returns you to the home screen. Opinions on phrasing/placement/looks welcome. 

Also includes djlint changes on a few HTML files.

Edit: 
The cancel button is now in its own `cancel-button.html`, which lives under `backend/report_submission/templates`. 

Examples:
image
image


backend/report_submission/templates/report_submission/step-1.html
```diff
@@ -149,14 +153,50 @@
...
+                           aria-controls="modal-cancel"
+                           data-open-modal>Cancel</a>
                     </li>
                     <li class="usa-button-group__item">
                         <button class="usa-button" id="continue">Continue</button>
                     </li>
                 </ul>
             </fieldset>
         </form>
+        <div class="usa-modal flex-justify-start"
```
sambodeme, line 166: This modal is used at least at three places (step-1, step-2, step-3). I would suggest to put that in a fragment instead and  and use it like  {% include 'report_submission/your_modal_fragment_name.html' %} wherever it needed to avoid code duplication.

sambodeme (commented):
Was able to test (manually/automated test) this with no issue, great!

backend/report_submission/templates/report_submission/step-1.html
jperson1: I had initially intended to do this, and then... didn't? I don't know. It's done now, though!

sambodeme (approved):

LindsayYoung (approved):
confirmed locally that this is working

PR #968, opened 2023-04-10
Using Jsonnet for schema

jadudm:
Draft ADR for using JSonnet for schema validation.

### DIFF
--- a/docs/architecture/decisions/0018-jsonnet-schema.md
+++ b/docs/architecture/decisions/0018-jsonnet-schema.md
@@ -30,6 +30,10 @@ This also provides a layer of validation on our schema; a valid Jsonnet file wil
 
 We will author our schema (which validate forms and spreadsheets in the project) in Jsonnet.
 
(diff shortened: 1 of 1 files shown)


jadudm:
Will do. I think... hm. My inclination is to *not* track the build products in version control. At the moment we are, only because we don't have the Jsonnet versions building with our automation. 

Any concerns about leaving the resulting schema out of version control? Or, should they be static enough that we want to track them in version control, even though they are (technically) build products?

jadudm:
Noting for revision:

In conversation, we agreed we should reorganize that part of the tree a  bit, and keep JSON outputs in the tree.

1. Have a `src`, `schema`, and possibly... `test-files` subfolders.
2. Keep the JSON outputs from the sonnet files.
3. Create a Makefile

The reason being: they're critical docs. They *are* the validation, basically. And, someday, we might people at them for external use.

To be revised.

jadudm:
@tadhg-ohiggins , changes made as noted.

PR #971, opened 2023-04-10
Require all fields in XLSX worksheets

jadudm:
Currently, sheets have fields that can be optionally left blank. This ADR suggests all fields have positive values that must be entered (e.g. "N/A" vs. leaving a field blank.)

For discussion.

### DIFF
--- a/docs/product/decisions/0008-sfsac-all-fields.md
+++ b/docs/product/decisions/0008-sfsac-all-fields.md
@@ -1,0 +1,32 @@
+# 1. Require all fields in SF-SAC XLSX uploads
+
+Introduced: 2023-04-10
+
+## Status
+For consideration
+
+## Context
+
+Much of the SF-SAC was, and will continue to be, submitted as spreadsheets.
+
+However, users are not required to complete all of the fields in a spreadsheet row. This creates opportunities for user error, and therefore for data validation problems.
+
+For example, consider that the SF-SAC "Federal Awards" template had a column that asks if the row is a load or loan guarantee. If the user answers `Y`, then they are required to complete additional columns. However, if they answer `N`, they can (optionally) leave the subsequent columns blank. Or, they could say `N/A`. (Different conditional columns also have different behavior in this regard, which is confusing.)
+
+## Decision
+
+Every column in every row should require a value.
+
+In the above example, if the user answers `Y`, then the next few columns are obviously required (e.g. "loan amount remaining.") If they answer `N`, however, they should be *required* to enter `N/A`, `0`, or some other appropriate value in the related, remaining columns. In this regard, we are asking for auditors and auditees to positively confirm, without question, their answer to the relevant question. 
+
+The problem with empty columns in a row is that we cannot tell the difference between a user who has *chosen* to leave the column blank vs. a user who *forgot* to enter a value. By requiring a positive action on every field, we increate the quality of the data in the FAC, and help users make fewer errors.
+
+This does not change the reporting burden, as the fields already exist and had to be completed. 
+
+## Consequences
+
+Documentation will need to be updated, but it will become simpler: every field has values that **must** be present. There are no longer any **should** values.
+
+This change makes some things explicit for users that may have been tacit in the past. 
+
+Validation rules need to be refined, but this will be a simplification of existing rules.


docs/product/decisions/0008-sfsac-all-fields.md
```diff
@@ -0,0 +1,32 @@
+# 1. Require all fields in SF-SAC XLSX uploads
+
+Introduced: 2023-04-10
+
+## Status
+For consideration
```
mogul, line 6:
```suggestion
Approved
```

If this makes it through the PR process and gets merged, then at that point by definition it's approved, and we wouldn't want to have to make another PR just to change this line. ;)

docs/product/decisions/0008-sfsac-all-fields.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Context
+
+Much of the SF-SAC was, and will continue to be, submitted as spreadsheets.
+
+However, users are not required to complete all of the fields in a spreadsheet row. This creates opportunities for user error, and therefore for data validation problems.
+
+For example, consider that the SF-SAC "Federal Awards" template had a column that asks if the row is a load or loan guarantee. If the user answers `Y`, then they are required to complete additional columns. However, if they answer `N`, they can (optionally) leave the subsequent columns blank. Or, they could say `N/A`. (Different conditional columns also have different behavior in this regard, which is confusing.)
```
mogul, line 14:
```suggestion
For example, consider that the SF-SAC "Federal Awards" template had a column that asks if the row is a loan or loan guarantee. If the user answers `Y`, then they are required to complete additional columns. However, if they answer `N`, they can (optionally) leave the subsequent columns blank. Or, they could say `N/A`. (Different conditional columns also have different behavior in this regard, which is confusing.)
```

mogul (changes requested):

jadudm:
It does represent a shift in thinking. Let's play with this.

I hear you saying "people may submit a sheet that needs more work, with the assumption someone else will download it, do things to it, and re-submit." This suggests the FAC is a kind of **single-slot, mostly inadequate case management system**: it is how auditors and auditees share files with each-other as they work on a submission. We have never claimed the FAC is a case management system for end-users. Furthermore, we don't have a notion of validation that considers any submission "in-progress." We have one set of validation rules, and those are applied at the time of upload.  When a spreadsheet comes in, we assume it is done, and we pull it apart (and validate it) as such. There is no "later" validation that is more stringent.

I'm suggesting that the FAC is a **submission system**. People submit form data, upload files, and certify that their work is accurate to the best of their ability, because they are submitting audits on hundreds of thousands to billions of dollars. If they want to collaborate on those documents before submission, they have tools (email, various file sharing services like GDrive, etc.) to do so. We're not trying to replace those tools.

So, I'm saying lets act like a submission system and not a kinda-but-not-really case management system. When data hits our system, we expect all the data to be complete and correct. Empty fields are not "an empty string, empty array, or null," but instead "N/A" or the correct value (for example). If we tighten the rules on our fields/forms, then we can assure more correct data, and give better instructions to the users preparing the documents. As a result, we increase the quality of the data coming into the system, because the rules are simpler and clearer. 

To your example: a user who leaves a field blank (because we allow it) and forgets to fill it later will still submit the FAC with the missing field, because we have no subsequent validation that is more stringent. Hence, if we only have one point of validation (when any given form page or spreadsheet is saved or uploaded), then we need to apply the most stringent/data-protecting validations we can *at that point*, and the user cannot submit that portion of work until they submit work that is complete and correct.

tadhg-ohiggins:
I'm all for switching to a submission system instead of a case management system.

> a user who leaves a field blank (because we allow it) and forgets to fill it later will still submit the FAC with the missing field, because we have no subsequent validation that is more stringent.

We had always planned to have not one, but _two_ sets of validation rules, and by that plan they would not be able to submit the blank fields.

jadudm:
I'm digging deeper into things with Census around this, and there may be more external conversation required.

We probably will need 

1. Per-sheet validation and 
2. Cross-sheet validation (end)

That will be different than *more* and *less* stringent validations. 

So, I'm working to move us to submission vs. case management, and continuing to work to simplify validation process. But, because the Uniform Guidance has so many conditionals, it's hard. We'll see where we get; this is worth pushing on.

jadudm:
Closing this PR as it will get picked up in #1190.

PR #991, opened 2023-04-12
Adding a script to migrate public PDFs to the public bucket

LindsayYoung:
- Creates the scaffolding to upload PDFs to public and private buckets, once we have a mapping, we will be able to update so that we pull the records we need.
- Upload and add to model framework for PDF workflows
- Add documentation on how to modify the script once we have the data we need

### DIFF
--- a/backend/config/settings.py
+++ b/backend/config/settings.py
@@ -323,6 +323,14 @@ else:
     # Will not be enabled in cloud environments
     DISABLE_AUTH = False
 
+# Remove once all Census data has been migrated
+# Add these as env vars, look at the bucket for values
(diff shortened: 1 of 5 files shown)


jadudm:
@danswick , is there any reason we should not merge this in? (Or, given your work in this space, would you mind giving it a look? I can merge things through after a thumbs up from you.)

jadudm:
Actually, I'm going to merge this. The work is part of the data migration, and should be in `main`. We're continuing to iterate on it regardless.

jadudm:
Oh. I can't. OK.

danswick:
I'll go ahead and merge, @jadudm. I created a brief clean-up ticket under the data migration epic so we can track some of the temporary env vars and the like. https://github.com/GSA-TTS/FAC/issues/1047

PR #1001, opened 2023-04-14
Docs: fix loging.gov typo in BPMN diagrams

tadhg-ohiggins:
Spelling can be hard;
I N so often followed
by G—but not here.

-----

`loging.gov` -> `login.gov` in BPMN source files.


tadhg-ohiggins:
Closing because we need to check that the [BPMN checks fix](https://github.com/GSA-TTS/FAC/pull/1002) worked.

PR #1002, opened 2023-04-14
Restore BPMN checks

mogul:
We want to ignore _most_ documentation changes, but BPMN source changes should trigger diagram updates, for example. 

This commit flips our trigger from using `paths-ignore:` to `paths:` to enable more flexible specification of exceptions. [See docs on this pattern here.](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths)

### DIFF
--- a/.github/workflows/triggers-PR.yml
+++ b/.github/workflows/triggers-PR.yml
@@ -5,8 +5,10 @@ on:
     branches:
       - main
       - prod
-    paths-ignore:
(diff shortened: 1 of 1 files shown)


mogul:
@LindsayYoung JFYI

tadhg-ohiggins (approved):

PR #1005, opened 2023-04-14
Hotfix: Update production manifest to fix deploy v1.20230412

asteel-gsa:
**Description:**
Deployment [v1.20230412](https://github.com/GSA-TTS/FAC/actions/runs/4681028221) failed due to manifest error
```
Pushing with manifest file: backend/manifests/production/manifest-fac-production.yml
Expected to find variables: DB_URI
FAILED
```
**File impacted:** 
[manifest-fac-production.yml](https://github.com/GSA-TTS/FAC/blob/4b597ab25f1fa42110f42971747af52f641ec81d/backend/manifests/production/manifest-fac-production.yml#L23-L49) where [deploy-apps.yml](https://github.com/GSA-TTS/FAC/blob/e0639a98b610feb25070a6604c0145fc45a293a2/.github/workflows/deploy-apps.yml#L64-L71) line 71 does not pass `--var DB_URI=${{ secrets.DB_URI }}` and is contained within [manifest-postgrest-production.yml](https://github.com/GSA-TTS/FAC/blob/main/backend/manifests/production/manifest-postgrest-production.yml) and passed in [deploy-apps.yml](https://github.com/GSA-TTS/FAC/blob/e0639a98b610feb25070a6604c0145fc45a293a2/.github/workflows/deploy-apps.yml#L79) line 79

**Resolution:**
Remove the yml from prod manifest, as the postgrest and swagger yml are handled in their own manifests, and picked up with the workflow.

**Pre-Merge Checks:**
- [X] Draft PR has been created
- [X]  Github Actions pass successfully

@jperson1 @JeanMarie-TTS


JeanMarie-TTS:
1.  We have three trigger files - one for each space. Can we not combine them into a single one?
2. The trigger for staging has the following line, which refers to the dev space:
 `url: "https://fac-dev.app.cloud.gov/"`
3. When I accessed the app in fac-prod, it took me directly to a login.gov page instead of taking me to the FAC home page.

The above van be addressed in this PR if these are small changes. Otherwise, we should create separate tickets for the unresolved items.

tadhg-ohiggins:
@JeanMarie-TTS I don't we should address any of those in this PR, since it is a hotfix targeting `prod`.

On the specific items you mention:

1. Merging trigger files: Given the amount of time we've spent troubleshooting these files, I would argue against spending any time combining three trigger files into one, and I suspect there are details around that which would make it trickier than it might appear.
2. Staging trigger: Unless I am mistaken, the intent here is for the scan of the dev site to occur before pushing to the staging site as a way of verifying that the deployed code passes the scan. So I think https://github.com/GSA-TTS/FAC/blob/prod/.github/workflows/triggers-staging.yml#L18 may be corect.
3. Login redirect in production: It's possible this is intentional since the site isn't really up yet. We might want to check on what we think should happen there.

tadhg-ohiggins (approved):

mogul:
The issue here was not per-environment triggers, but rather [per-environment manifests](https://github.com/GSA-TTS/FAC/tree/main/backend/manifests). The manifest originally included all three apps but was later refactored into three separate manifest files for `dev` and `staging`. In prod, the same refactoring happened, but the original file did not have the outdated entries removed.

mogul:
If you'd like to refactor the manifest files to make them shared across environments, you may want to know that [manifests can be parameterized using the notation described here, with an environment-specific vars file filling in the values](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#variable-substitution). This would eliminate copied code and make environment-specific differences explicit.

mogul:
Here's an example of [a single parameterized manifest file](https://github.com/GSA/catalog.data.gov/blob/main/manifest.yml) with separate vars files with valuies for the [`development`](https://github.com/GSA/catalog.data.gov/blob/main/vars.development.yml), [`staging`](https://github.com/GSA/catalog.data.gov/blob/main/vars.staging.yml), and [`prod`](https://github.com/GSA/catalog.data.gov/blob/main/vars.prod.yml) environments.

Issue #928, opened 2023-03-31
SF-SAC submission


jadudm:
# At a glance

[comment]: # "Begin with a short summary of this ticket, so its intent can be understood at a glance."

Auditors and auditees work together to complete the SF-SAC.

# Acceptance Criteria

We use [behavior-driven development](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications) wherever possible.

[comment]: # "ACs should be clearly demoable/verifiable whenever possible. Try specifying them using behavior driven development."
[comment]: # "**Given** **when** **then**"
[comment]: # "Given: the initial context at the beginning of the scenario"
[comment]: # "when: the event that triggers the scenario"
[comment]: # "then: the expected outcome(s)"
[comment]: # "Repeat scenarios as needed, or repeat behaviors and lists within a scenario as needed."

#### Key
* submitter = auditee or auditor

### Scenario: Submitters want to submit their SF-SAC and associated files.

**Given**  that a submitter is ready to start the SF-SAC
**when**  they have completed the Pre-SAC checklist
...

```[tasklist]
### then...
- [x] the submitter can complete the SF-SAC general information page
- [ ] https://github.com/GSA-TTS/FAC/issues/945
- [ ] https://github.com/GSA-TTS/FAC/issues/989
- [ ] https://github.com/GSA-TTS/FAC/issues/977
- [ ] https://github.com/GSA-TTS/FAC/issues/992
- [ ] https://github.com/GSA-TTS/FAC/issues/993
- [ ] https://github.com/GSA-TTS/FAC/issues/965
- [ ] https://github.com/GSA-TTS/FAC/issues/994
- [ ] https://github.com/GSA-TTS/FAC/issues/995
- [ ] https://github.com/GSA-TTS/FAC/issues/996
```

```[tasklist]
### E2E Testing
- [ ] https://github.com/GSA-TTS/FAC/issues/1055
- [x] Audit Findings
- [x] Audit Findings Text
- [x] CAP (Corrective Action Plan)
```

### Shepherd

[comment]: # "@ mention shepherds as we move across the board."

* Design shepherd: 
* Engineering shepherd: 

# Background

[comment]: # "Any helpful contextual notes or links to artifacts/evidence, if needed"

# Security Considerations

Required per [CM-4](https://nvd.nist.gov/800-53/Rev4/control/CM-4).

[...]

PR #1016, opened 2023-04-17
Upload Page Template

jperson1:
Copy paste is bad
Eight files is less than fun
Templates are good news

-----

This PR covers "the ___ page is authored in the frontend" for all of the upload pages covered by #928 (SF-SAC submission)

The proposal is:
1. Have a single `upload-page.html` handle all the upload pages, with the differences between handled by passed context from the view.
2. Have a single `upload-page.js` that can handle different excel validation based on the passed context from the view.
3. If any page is determined to be significantly different from the others, it can be fairly easily broken into its own files. 

Caveats to this PR:
- The upload validation runs against the federal awards schema, regardless of which upload page you're on. All the schemas don't exist yet. 
- Tests are not complete - to be implemented as another part of #928

Next steps:
1. Change excel validation endpoint to allow for more than just federal awards upload. 
  a. Alternatively, implement several endpoints for each type of upload validation. Each has its benefits. 
2. If a file is already stored for this audit, have it show in the front end and (potentially) allow it to be downloaded.


jadudm (changes requested):
You can double-check me, but I think, after I submit the Federal Awards sheet, I get directed to ... Audit Findings. This looks like we jump around the list of submissions on the LH side of the page.

I'd recommend adjusting the layout on the page so that the sheets we upload match the order that we upload files in. Or, if we are going to redirect from an upload page to another, then it would be good to make sure the ordering on the left matches the order we sequence them in?

Hm. Also, nothing happens for me when I click:

* Notes to SEFA
* Audit information
* Finalize

But, I think Finalize is "out of scope" for this particular bit of work.

I think the ordering is a change request, and I'm uncertain whether the other pages should "work" with this PR? I think they should?

jadudm:
I'm going to merge this to `main`, as I believe this is going to make it easier as we bring the rest of #928 together. Thank you for the revisions/changes; LGTM, but we can expect some (possible) breakage this week as everything comes together.

Issue #1017, opened 2023-04-17
Update manifests to be standardized across the environments.


asteel-gsa:
### Description:
Upon removing some of the code in #1005 that removed postgrest and swagger from the fac-prod manifest, there was a discussion about the usage of the multiple manifests across the environments.

### Fix
[deploy-apps.yml](https://github.com/GSA-TTS/FAC/blob/main/.github/workflows/deploy-apps.yml) would contain a new passthrough `--var` for each of the [deploys](https://github.com/GSA-TTS/FAC/blob/cb9d5983205fd5b45fe98048959dd9ec0d2269d3/.github/workflows/deploy-apps.yml#L64-L87) in the context of: 
`--var ENVIRONMENT=${{ env.space }}`

In turn, that would accomplish the following.
- Parameterized and Abstracted code
- Turn 9 (12 after new relic) into 3 (4 after new relic)
- Minimize drift across the 3 environments, having `deploy-apps.yml` calling 1 file, instead of 3 for a deployment in each environment.

This change would turn [manifests](https://github.com/GSA-TTS/FAC/tree/main/backend/manifests) into non-environment dependent manifests, creating the following tree:
```
backend/manifests/ -
| - manifest-fac.yml
| - manifest-postgrest.yml
| - manifest-swagger.yml
| - manifest-newrelic.yml
```
instead of:
```
backend/manifests/ -
| -/{{env}}
| | - manifest-fac-{{env}}.yml
| | - manifest-postgrest-{{env}}.yml
| | - manifest-swagger-{{env}}.yml
| | - manifest-newrelic-{{env}}.yml
```

The code changes would create blocks like this:
```
    env:
      ENV: DEVELOPMENT
      DJANGO_BASE_URL: https://fac-((ENVIRONMENT)).app.cloud.gov
      ALLOWED_HOSTS: fac-((ENVIRONMENT)).app.cloud.gov
      AV_SCAN_URL: https://fac-av-((ENVIRONMENT)).apps.internal:61443/scan
```
instead of:
```
    env:
      ENV: DEVELOPMENT
      DJANGO_BASE_URL: https://fac-dev.app.cloud.gov
      ALLOWED_HOSTS: fac-dev.app.cloud.gov
      AV_SCAN_URL: https://fac-av-dev.apps.internal:61443/scan
```

### Tasks:
```[tasklist]
- [x] Utilize the new manifest structure for swagger
- [x] Utilize the new manifest structure for postgrest
[...]

PR #1025, opened 2023-04-19
Standardize manifests and utilize a vars file

asteel-gsa:
**Ticket:**  #1017 
**Mock Deployment with Vars File:** [Swagger Deploy](https://github.com/GSA-TTS/FAC/actions/runs/4743412135/jobs/8422931940)

**Files to Review:**
- [deploy-apps L72](https://github.com/GSA-TTS/FAC/blob/f22ba3908b90664173aa163aa58fc5b9c93623fe/.github/workflows/deploy-apps.yml#L72) | [deploy-apps L80](https://github.com/GSA-TTS/FAC/blob/f22ba3908b90664173aa163aa58fc5b9c93623fe/.github/workflows/deploy-apps.yml#L80) | [deploy-apps L89](https://github.com/GSA-TTS/FAC/blob/f22ba3908b90664173aa163aa58fc5b9c93623fe/.github/workflows/deploy-apps.yml#L89)
- [Vars Files](https://github.com/GSA-TTS/FAC/tree/asteel/standardize-manifests/backend/manifests/vars)
- [fac](https://github.com/GSA-TTS/FAC/blob/asteel/standardize-manifests/backend/manifests/manifest-fac.yml) | [postgrest](https://github.com/GSA-TTS/FAC/blob/asteel/standardize-manifests/backend/manifests/manifest-postgrest.yml) | [swagger](https://github.com/GSA-TTS/FAC/blob/asteel/standardize-manifests/backend/manifests/manifest-swagger.yml)

**To-Do:**
- Refactor postgrest & swagger into terraform layer, instead of manifest layer cc: @mogul

### Pre-Merge Checklist:
```[tasklist]
- [X] Code has been reviewed before PR is opened
- [X] Draft PR has been created
- [X] Pre-Merge tests complete
```


backend/manifests/manifest-postgrest.yml
```diff
@@ -0,0 +1,15 @@
...
+applications:
+  - name: ((pgrst_app_name))
+    instances: ((pgrst_instances))
+    disk_quota: ((pgrst_disk_quota))
+    memory: ((pgrst_mem_amount))
+    timeout: 180
+    docker:
+      image: ((pgrst_image))
+    routes:
+      - route: fac-((env_name))-postgrest.app.cloud.gov
```
jadudm, line 11: Should this also be ((app_name))?  (It shows up a few places... that is, if we set the app name, should we also use that app name in the URL?)

asteel-gsa: @jadudm if we did that we should standardize across the fac app as well, but keep in mind, per Bret, the next step after this will be entirely removing postgrest and swagger from these manifests, and handling them in terraform

asteel-gsa: I believe your proposed change is this, correct?:

`manifest-postgrest.yml`
```
    routes:
      - route: fac-((env_name))-postgrest.app.cloud.gov
```
 
 `manifest-swagger.yml`
```
    routes:
      - route: fac-((env_name))-((swagger_app_name)).app.cloud.gov
    env:
      API_URL: https://fac-((env_name))-((pgrst_app_name)).app.cloud.gov
      SWAGGER_JSON_URL: https://fac-((env_name))-((pgrst_app_name)).app.cloud.gov/
```

I don't see an issue with it if we want to go that abstract

jadudm: Ah. If this is going to go away soon with the move to Terraform, then don't worry about it either way. I was just pattern matching as I read through the PR.

mogul: Personally I think we should only parameterize things that will change per-environment. Otherwise we're abstracting a "constant" and it just makes the manifest harder to read.

(Counterpoint: Making the app name fully abstract enables multiple parallel deployments in a given space without any collisions, eg if multiple people are working on a branch.)

backend/manifests/vars/vars-dev.yml
```diff
@@ -0,0 +1,21 @@
+# gsa-fac manifest
+app_name: gsa-fac
+mem_amount: 512M
+cf_env_name: DEVELOPMENT
+env_name: dev
+service_name: dev
```
mogul, line 6: These three variables feel both confusing and redundant... Their values are nearly identical per-environment, and the name of the variable doesn't really help us understand how they're used.

It's not a blocker for this PR, but we should think about how we can further standardize so that the `env_name` is all that's needed across app names, routes, and env vars.

mogul (approved):

backend/manifests/vars/vars-dev.yml
asteel-gsa: cf_env_name will always be caps, and the service_name is used for long, lowercase names.

There is a gap in the original manifests, where for prod, the following `PRODUCTION` `prod` and `production` are used. and then in dev, it is `DEVELOPMENT` and `dev`

the values coming in from inputs.environment are `dev, staging, production`

PR #1028, opened 2023-04-19
Move path back 1 level to properly find procfile in backend/

asteel-gsa:
Deployment: [Deploy into dev](https://github.com/GSA-TTS/FAC/actions/runs/4747141266/jobs/8431659080)

Issue arose from `path: ../../` in manifest-fac.yml needing to be `path: ../` to properly pick up the procfile


jadudm (approved):
LGTM; did a pair, and this now points everything to the right place in the tree.

PR #1033, opened 2023-04-20
Remove obsolete manifests

mogul:
These are no longer needed now that these two PRs are in: 
https://github.com/GSA-TTS/FAC/pull/1025
https://github.com/GSA-TTS/FAC/pull/1028


asteel-gsa:
Was going to leave this task until we get a tagged deployment into production, just so we ensure that the new rework moves through the stack as expected. 

But if we want to remove now, LGTM

asteel-gsa (approved):

Issue #1015, opened 2023-04-17
Set up and configure egress proxy services


JeanMarie-TTS:
# At a glance

[comment]: # "Begin with a short summary so intent can be understood at a glance."

[comment]: # "In order to: some objective or value to be achieved"
[comment]: # "as a: stakeholder"
[comment]: # "I want: some new feature"

**In order to** restrict app egress while enabling connections that are explicitly allowed for operation (meeting the intent of NIST control SC-7)
**As a** FAC system architect
**I want** a proxy in a public-egress space that will only proxy connections to configured domains.

# Acceptance Criteria

We use [DRY](https://docs.behat.org/en/latest/user_guide/writing_scenarios.html#backgrounds) [behavior-driven development](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications) wherever possible.

[comment]: # "ACs should be clearly demoable/verifiable whenever possible."
[comment]: # "Given: the initial context at the beginning of the scenario"
[comment]: # "when: the event that triggers the scenario"
[comment]: # "then: the expected outcome(s)"
[comment]: # "Repeat scenarios as needed, or repeat behaviors and lists within a scenario as needed."

[comment]: # "The scenario should be a short, plain language description."
[comment]: # "Feeling repetative? Apply the DRY (Don't Repeat Yourself) principle!"

### Scenario: The egress proxy is deployed and properly configured

**Given** I am logged into cloud.gov and targeting the `gsa-tts-oros-fac` organization
**When** I run `cf spaces`
...

[comment]: # "Each task should be a verifiable outcome"
```[tasklist]
### then...
- [x] I see that there are `[dev|staging|production]-egress` spaces
```

**Given** I am logged into cloud.gov 
**And** and I have run `cf t -o gsa-tts-oros-fac -s [dev|staging|production]-egress`
**When** I run `cf app egress ; cf network-policies; cf env egress | grep PROXY`
...

[comment]: # "Each task should be a verifiable outcome"
```[tasklist]
### then...
- [x] I see that the `egress` app is running
- [x] I see a route on the `apps.internal` domain is mapped to the app
- [x] I see a network policy allows connections from FAC applications in the corresponding space to port 61443
- [x] I see that the expected allow/deny configuration and a random username/password in the egress environment variables
```

**Given** I am logged into cloud.gov 
**And** and I have run `cf t -o gsa-tts-oros-fac -s [dev|staging|production]`
**When** I run `cf service egress-creds`
...

[...]

PR #1052, opened 2023-04-27
Add egress proxy Terraform configuration

mogul:
Related to #1015

Once this is merged:

- The egress service will be available in the `dev` space.

A subsequent PR will:
1. add the app logic to have the client app(s) make use of the proxy
2. make the egress service available in staging+production (uncomment inclusion of the `env-egress` shared module)


mogul:
Just realized that this doesn't include any GitHub Action mojo to preview and deploy changes to the `-egress` environments... Moving back to draft state until that's in here, but please start code review now!

mogul:
This is again ready for review! The most recent commit is likely to be overwhelming, so let's go over it together. 

We will also probably need to work together post-merge... Terraform is going to try to create the deployer services and service-keys and fail because they were already created manually. We will need to import the existing GUIDs so that Terraform can take ownership of them.

terraform/shared/modules/egress-proxy/prepare-proxy.sh
```diff
@@ -0,0 +1,24 @@
...
+
+popdir=$(pwd)
+
+# Portable construct so this will work everywhere
+# https://unix.stackexchange.com/a/84980
+tmpdir=$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
+cd "$tmpdir"
+
+# Grab a copy of the main branch
+curl -s -L https://github.com/GSA-TTS/cg-egress-proxy/archive/refs/heads/main.zip --output main.zip
```
mogul, line 14: We should be grabbing a particular release tag (which implies that the repository is making tags/releases).

terraform/shared/modules/egress-proxy/test/validate.sh
```diff
@@ -0,0 +1,24 @@
...
+cf env ${APPNAME} | grep -q PROXY_DENY
+cf env ${APPNAME} | grep -q PROXY_USERNAME
+cf env ${APPNAME} | grep -q PROXY_PASSWORD
+
+# Check that a network policy with appropriate values exists 
+# TODO: Improve this to use just one robust regexp (or use "cf curl" and "jq").
+cf network-policies | grep test | grep ${APPNAME} | grep 61443 | grep "$ORGNAME" | grep -q "$SPACENAME" 
+
+# Check that a user-provided service with the egress credentials exists
+cf service ${APPNAME}-creds > /dev/null 2>&1
```
mogul, line 24: Add a `cf ssh` with a curl to test that the allow/deny acls are working

mogul (commented):
A bunch of stuff we noticed when walking through together... Fixes incoming!

terraform/shared/modules/egress-proxy/test/validate.sh
mogul: I added a `cf ssh` test that uses `curl` and I can see that it's giving the results I expect. However, it's a little squirrely to test the output is what's expected because the return value is hard to extract over cf ssh. I'm too tired to work through that now so I'll need someone to help with it in the morning!

mogul:
I pushed fixes for everything we found in our review today. However, two problems remain:

1. The -egress spaces don't exist as of this PR; an apply for `management` following this PR being merged will create them. So terraform plan fails for `dev`, because the space it tries to read isn't there!
  - I'll cherry-pick the commit that creates the -egress spaces into a separate PR that we can merge on its own.
2. It looks like the terraform actions include `unzip`, but not `zip`! So the script that prepares the proxy .zip files are failing.
  - I'm not sure if this is just a matter of `apt install zip` before running terraform in the GitHub workflow, or if something explicit must be done with the dflook GitHub Actions that we're using. Needs investigation and maybe a pair in the morning.

mogul:
> The -egress spaces don't exist as of this PR; an apply for `management` following this PR being merged will create them. So terraform plan fails for `dev`, because the space it tries to read isn't there!
> * I'll cherry-pick the commit that creates the -egress spaces into a separate PR that we can merge on its own.

Done in #1060... 👀 over there please for the merge/apply!

mogul:
Oh unless we want to resolve this first:
> I added a `cf ssh` test that uses `curl` and I can see that it's giving the results I expect. However, it's a little squirrely to test the output is what's expected because the return value is hard to extract over cf ssh. I'm too tired to work through that now so I'll need someone to help with it in the morning!

mogul:
> > I added a `cf ssh` test that uses `curl` and I can see that it's giving the results I expect. However, it's a little squirrely to test the output is what's expected because the return value is hard to extract over cf ssh. I'm too tired to work through that now so I'll need someone to help with it in the morning!

I resolved this, and created the `-egress` spaces directly while I address the root cause that's blocking #1060 from working. We're now getting [a good plan for `dev`](https://github.com/GSA-TTS/FAC/pull/1052#issuecomment-1524746135), so this is ready for a final review, merge, and deploy!

JeanMarie-TTS (approved):

PR #1060, opened 2023-04-28
Create the "-egress" spaces where the egress proxies will live

mogul:
Related to #1015, and a necessary precursor to PR #1052. 

We specify not only that the -egress spaces should exist, but that a corresponding -deployer account should have access to deploy in them too.

(I've renamed the existing `*-deployer` services in each space to `*-deployer-handpropped` so they won't be in the way when Terraform creates the ones we specify here.)

Once this is merged:

- There will be new `-deployer` services and `-deployer-key` service keys in each of the `[dev|staging|production|management]` spaces

We must then manually:

1. Query the `-deployer-key` service keys and update the `CF_USERNAME` and `CF_PASSWORD` secrets in each of [our GitHub environments](https://github.com/GSA-TTS/FAC/settings/environments)
2. Delete the now-obsolete `-deployer-handpropped` services and service-keys from each space


terraform/management/main.tf
```diff
@@ -8,22 +8,55 @@ data "cloudfoundry_asg" "asgs" {
...
+  for_each         = toset(local.spaces)
+  name             = "${each.key}-deployer-key"
+  service_instance = cloudfoundry_service_instance.space-deployer[each.key].id
+}
+
+resource "cloudfoundry_space_users" "egress-deployer" {
+  for_each   = toset(local.spaces)
+  space      = cloudfoundry_space.space["${each.key}-egress"].id
+  developers = [cloudfoundry_service_key.space-deployer-key[each.key].credentials.username]
+}
```
asteel-gsa: Needs a \n at the end of file for code cleanliness.

asteel-gsa: resolved

asteel-gsa:
@JeanMarie-TTS approved, please wait for the PR tests to complete and then merge to resolve deployment on `<env>-deployer`

asteel-gsa (approved):

PR #1061, opened 2023-04-28
Remove spurious dependency on deployer service

mogul:
The FAC app doesn't need to manipulate other apps or services in the space on cloud.gov, so it shouldn't be bound to the deployer credential service instance.


asteel-gsa (approved):
Per discussions in slack and on egress-proxy related information, approved.

PR #1072, opened 2023-05-02
Add sketch section

mogul:
Prep a place where we can easily add implementation tasks that are below the level of an AC. (Also fixes a Markdown typo.)

### DIFF
--- a/.github/ISSUE_TEMPLATE/story-issue-template.md
+++ b/.github/ISSUE_TEMPLATE/story-issue-template.md
@@ -15,7 +15,7 @@ assignees: ''
 [comment]: # "I want: some new feature"
 
 **In order to** 
-**as a **
+**as a**
 **I want**
 
 # Acceptance Criteria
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/story-issue-template.md
```diff
@@ -62,6 +62,14 @@ Required per [CM-4](https://nvd.nist.gov/800-53/Rev4/control/CM-4).
 [comment]: # "Please do not remove this section without care."
 [comment]: # "Note any security concerns that might be implicated in the change. 'None' is OK, but we must be explicit here."
 
+# Sketch
+
+We're thinking we'll try implementing A and then changing B.
+
+```[tasklist]
```
danswick, line 69: What do you think about adding a short comment to describe what a "sketch" is for?

```suggestion
We're thinking we'll try implementing A and then changing B.

[comment]: # "Broad list of implementation steps that aren't as well-defined or 'demoable' as ACs"
```[tasklist]
```

danswick (commented):
Added a suggestion for you to consider, @mogul!

mogul:
> I'm not sure how this is different from "Background," but if this is useful to others, lets put it in the template.

"Background" is context for the work that's not obvious from the story statement or ACs, for example:
> "We have to do this because XYZ, and our user research/recordings supporting that are available at ABC. We tried doing these other things before, but those weren't sufficient because LMN"

"Sketch" is notes for implementers, for example:
> Since this is going to require XYZ, we're going to pay down the tech debt on ABC. So the first thing to do is clean up that migration, fix the tests, and then LMN. Also, don't forget:
> - [ ] Get Laura to look at the copy
> - [ ] Update the docs if we ended up changing behavior from what's expected

The point in doing this is so that (1) people have a place to park thoughts that aren't specifically business-/user-relevant, so that (2) it's easier for them or (more importantly) **other people** to pick up the work later.

PR #1076, opened 2023-05-03
Remove deployment of postgrest and swagger from the GitHub Action

mogul:
These apps are managed via Terraform now


asteel-gsa (approved):
Approved - Managed via terraform now

Issue #725, opened 2023-01-09
Epic: Obtain an ATO for FAC


JeanMarie-TTS:
```[tasklist]
### Must/required for assessment
- [x] Initiate ATO meeting with Ryan Palmer
- [x] Have an ISSO assigned
- [x] Have a folder created
- [x] Complete FIPS 199 template
- [x] Complete DIA Template
- [x] Get PTA approved
- [x] Create architecture doc
- [ ] https://github.com/GSA-TTS/FAC/issues/1006
- [ ] #828
- [ ] #834
- [ ] https://github.com/GSA-TTS/FAC/issues/1031
- [ ] https://github.com/GSA-TTS/FAC/issues/1014
- [ ] https://github.com/GSA-TTS/FAC/issues/1015
- [ ] https://github.com/GSA-TTS/FAC/issues/1064
- [ ] https://github.com/GSA-TTS/FAC/issues/1071
- [ ] https://github.com/GSA-TTS/FAC/issues/1161
- [ ] https://github.com/GSA-TTS/FAC/issues/1027
- [ ] https://github.com/GSA-TTS/FAC/issues/1394
- [ ] https://github.com/GSA-TTS/FAC/issues/1022
- [x] Remove public_networks_egress ASG from spaces
- [x] Fill in data transfer agreement for Census/GSA transfer
- [ ] Provision OracleDB for ETL
- [x] Update boundary diagrams to include ETL/Oracle DB
- [x] Update boundary diagrams to include ETL/S3 bucket
- [ ] https://github.com/GSA-TTS/FAC/issues/1070
- [ ] https://github.com/GSA-TTS/FAC/issues/1522
- [ ] https://github.com/GSA-TTS/FAC/issues/1397
- [ ] https://github.com/GSA-TTS/FAC/issues/1677
- [ ] https://github.com/GSA-TTS/FAC/issues/1020
- [ ] https://github.com/GSA-TTS/FAC/issues/1019
- [ ] https://github.com/GSA-TTS/FAC/issues/1521
- [ ] https://github.com/GSA-TTS/FAC/issues/1679
- [ ] https://github.com/GSA-TTS/FAC/issues/1634
- [ ] https://github.com/GSA-TTS/FAC/issues/1648
- [ ] https://github.com/GSA-TTS/FAC/issues/1649
- [ ] https://github.com/GSA-TTS/FAC/issues/1676
```

JeanMarie-TTS:
[...]

PR #1077, opened 2023-05-03
Configure New Relic

mogul:


mogul:
I put the `NEW_RELIC_LICENSE_KEY` into the GitHub repository-level secrets. Let's review and merge this, and leave further customization of the `newrelic.ini` to future PRs!

backend/.profile
```diff
@@ -1,11 +1,20 @@
+#!/bin/bash
+
 export REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
 
+# Grab the New Relic license key from the newrelic-creds user-provided service instance
```
mogul, line 5: @asteel-gsa I noticed that [you don't even have to have a `newrelic.ini` file, only these two env vars set](https://docs.newrelic.com/install/python/?python-non-web=non-web-no&python-docker=docker-yes&python-framework=unsupported-asgi-framework&python-server=gunicorn#pass-in-some-basic-configurations-7). So that's all I do here now. We can introduce the `newrelic.ini` file as warranted when we start customizing other stuff.

JeanMarie-TTS:
You can run `make lint` locally to catch linting errors. You have have all the packages in place first - `make compile`, `ptenv activate`, etc.
We should probably create a `make docker-lint`

mogul:
> We should probably create a `make docker-lint`

It might help me. I have a hard time investing effort in installing and configuring a bunch of toolchain pieces locally that I only occasionally use! Also it would help keep everyone developing using the same versions of tools, and allow us to control updates explicitly. [The devContainer spec is for addressing exactly these problems](https://containers.dev/), but I haven't ever had time to invest in making one for any of my projects!

asteel-gsa:
@mogul There is something we may want to consider, and we should probably have a discussion about it with @ChrisB-16. We should assess the "sensitivity" level of the data being ingested via new relic, and we should probably consider adding the `newrelic.ini` at some point, or if not, then we should `export` an evironment var to set [high security mode](https://github.com/GSA-TTS/FAC/blob/00911bbd46f0a9af296bf6daf11e3acee7a6e585/backend/newrelic.ini#L75-L85) to `true`. Documentation is provided [here](https://docs.newrelic.com/docs/subscriptions/high-security) and can be utilized with the `NEW_RELIC_HIGH_SECURITY` var. 

Something to noodle on for data security and submission.

mogul:
That looks like something we'd definitely want to turn on for staging and prod. Devs might want it off for their local work and the dev environment though, so we should look into how hard it would be to have a separate account set up just for that. 

Can you add a checklist item to the "Could" section of #928?

mogul:
Got it, it's #725. (The tasklists don't show up in GitHub Mobile so I wasn't sure if I had the right one.)

mogul:
This PR should be ready to go btw, please review @JeanMarie-TTS!

PR #1080, opened 2023-05-03
Use github.base_ref directly on PRs

mogul:
In the PR trigger context (which is all this file handles) we can get exactly what we want from the github context without going through the event. We've seen the trigger end up false in the latter case, so we're hoping this will work better.


danswick (approved):
looks good 👍

PR #1081, opened 2023-05-04
Configure apps use https_proxy for outbound connections

mogul:
Keeping this in draft until we can uncomment in the proxy for clamav


mogul:
> Keeping this in draft until we can uncomment in the proxy for clamav

Done now, this should be ready to go!

mogul:
Waiting on [the GitHub incident](https://www.githubstatus.com/incidents/c2jg911dtkjb), which is preventing me from pushing the tag for the fix upstream...

mogul:
OK, _now_ we're ready... If the plan output looks weighty, it's because I did a ton of refactoring in this PR and a bunch of resource names changed, and I also updated the version of the modules we're using.

JeanMarie-TTS (approved):

PR #1086, opened 2023-05-04
Ensure DEBUG is always set, with default False

mogul:
[This was causing problems for DAP.](https://github.com/GSA-TTS/FAC/blob/7a24c74747b026e5e5635e2a75a9d07dbef72997/backend/cms/templatetags/fac_tags.py#L10)


asteel-gsa (approved):
Discussed during huddle.

DEBUG was not being set, and needed this to be added so we can make DAP work, and resolve the 500 on the app.

PR #1090, opened 2023-05-05
Get docker images into GH

JeanMarie-TTS:


JeanMarie-TTS:
This PR fails when it runs in a branch but is likely to work when run in main. Tried it on another public repo called JeanMarie-TTS/DocScan.

asteel-gsa:
Action that shows passing run: [DocScan Action](https://github.com/JeanMarie-TTS/DocScan/actions/runs/4939444665)

.github/workflows/download-image.yml
```diff
@@ -0,0 +1,59 @@
+---
+name: Docker Images
+# on:
+#   schedule:
+#     - cron: "*/15 * * * *"
+#   workflow_dispatch:
+on:
+  push:
```
mogul, line 8: What's the idea behind this trigger?

.github/workflows/download-image.yml
```diff
@@ -0,0 +1,59 @@
...
+    paths:
+      - '.github/workflows/**'
+jobs:
+  docker-update:
```
mogul, line 12:
```suggestion
  vet-images:
```

.github/workflows/download-image.yml
```diff
@@ -0,0 +1,59 @@
+---
+name: Docker Images
```
mogul, line 2:
```suggestion
name: Scan Images
```

mogul, line 2:
```suggestion
name: Scan Images
```

.github/workflows/download-image.yml
```diff
@@ -0,0 +1,59 @@
+---
```
mogul, line 1: The filename should be more descriptive of what it does, eg "scan-images.yml"

.github/workflows/download-image.yml
```diff
@@ -0,0 +1,59 @@
...
+    paths:
+      - '.github/workflows/**'
+jobs:
+  docker-update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    env:
+      GCR_PAT: ${{ secrets.GITHUB_TOKEN }}
```
mogul, line 18: Is there a reason to put this in the environment, rather than just referring to `${{ secrets.GITHUB_TOKEN }}` in line 52 below? If not then I'd do that, so that the variable assignment/usage doesn't distract from what's happening.

.github/workflows/download-image.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: grab trivy container
+        run: docker pull aquasec/trivy:latest
+      - name: Scan PostGREST
+        run: docker run  aquasec/trivy:latest image postgrest/postgrest:latest
+      - name: Scan swagger
+        run: docker run  aquasec/trivy:latest image swaggerapi/swagger-ui:latest
+      - name: Scan ClamAV
+        run: docker run  aquasec/trivy:latest image  --scanners vuln clamav/clamav:latest
+
```
mogul, line 38: We're scanning, but pushing the images regardless of what the scan found... There should be some kind of conditional here to prevent tagging+pushing if the images have findings that we don't accept, right?

.github/workflows/download-image.yml
```diff
@@ -0,0 +1,59 @@
...
+    permissions:
+      contents: read
+      packages: write
+    env:
+      GCR_PAT: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Pull PostGREST Docker Image
```
mogul, line 23: Consider using a matrix strategy here to reduce repetition (eg specify the list of images to scan just once at the top rather than repeating the steps for each below).

mogul (changes requested):
I'm mostly OK with merging this if the naming changes are done, but I'm confused about how the scan results will be acted upon; is that something we can resolve before merging? Or does it happen in an immediate followup?

JeanMarie-TTS:
> Action that shows passing run: [DocScan Action](https://github.com/JeanMarie-TTS/DocScan/actions/runs/4939444665)

The change works only on the default branch. It should run successfully in our repo when the PR is merged. There are no dependent jobs, so there won't be a problem even if this particular job fails in the future.

.github/workflows/download-image.yml
mogul: (Note that this is a good reason to use the [Trivy GitHub action](https://github.com/marketplace/actions/aqua-security-trivy) instead of running the Docker image ourselves... We can set severity and the exit code will be used to halt/proceed, and the results are ready to include in GitHub's native facility for alerting us to problems.)

.github/workflows/download-image.yml
JeanMarie-TTS: Fixed

.github/workflows/download-image.yml
JeanMarie-TTS: Was planning to look into this after I get it working.

> I'm mostly OK with merging this if the naming changes are done, but I'm confused about how the scan results will be acted upon; is that something we can resolve before merging? Or does it happen in an immediate followup?

I've addressed the name changes. Will have a follow up PR to make sure that we can use the  scanned images in other jobs and TF, that other jobs are not blocked if scan fails, etc.

mogul (approved):

PR #1092, opened 2023-05-08
Create bug-template.yaml

jadudm:
Creating a bug report template. This has already been asked for as part of the API pilot. Borrowed the template from USWDS.


JeanMarie-TTS (approved):
LGTM for now. We should get started with this now and tweak later as needed.

engrzakaug1989 (approved):

PR #1104, opened 2023-05-08
add login sandbox to egress allow list

timoballard:


terraform/shared/modules/env/env.tf
```diff
@@ -25,6 +25,9 @@ module "egress-proxy" {
       # because it only allows subdomain wildcards in `*.` as a prefix. So this wildcard is a little broader than
       # we would prefer, but realistically the tightest domain mask we can specify given our current solution.
       "*.newrelic.com:443",
+
+      # Login.gov sanbox
```
mogul, line 29: typo in comment, but I don't care. Tadhg though...? Don't let Tadhg see this.

mogul (approved):

PR #1105, opened 2023-05-09
Switch to CAPI v3 "rolling" app deployment strategy

mogul:
The "blue-green" deployment strategy predates CAPI v3, when the built-in "rolling" strategy became available. The blue-green method deploys an entirely new app _with its own unique GUID_ before taking down the older app. This is a problem because without the same GUID, things start to break in subtle ways. For example, a Terraform-created `cloudfoundry_network_policy` resource might refer to the old GUID and therefore the network policy would simply evaporate when the old app is deleted... That appears to be what's happening here.

The CF terraform provider [provides a way to cope with this](https://registry.terraform.io/providers/cloudfoundry-community/cloudfoundry/latest/docs/resources/app#update-resource-using-blue-green-app-id), but it's not necessary if we just use the built-in "rolling" deployment strategy, which keeps the same GUID.

### DIFF
--- a/terraform/shared/modules/egress-proxy/egress-proxy.tf
+++ b/terraform/shared/modules/egress-proxy/egress-proxy.tf
@@ -61,7 +61,7 @@ resource "cloudfoundry_app" "egress_app" {
   command          = "./caddy run --config Caddyfile"
   memory           = var.egress_memory
   instances        = var.https_proxy_instances
-  strategy         = "blue-green"
+  strategy         = "rolling"
 
   routes {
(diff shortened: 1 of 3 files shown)


mogul:
Perhaps you're wondering why we were using the `blue-green` strategy for deployment in the first place...? It's because [the docs were not updated to note `rolling` as an option when the underlying resource started using CAPI V3](https://github.com/cloudfoundry-community/terraform-provider-cloudfoundry/pull/482)!

jadudm (approved):
Subtle, and good catch. THANK YOU.

PR #1107, opened 2023-05-09
Point Swagger at public API

mogul:
It turns out Swagger does everything that it does client-side, and the URLs we provide here are only used to "pre-configure" the Swagger UI to point at a particular URL. Since there's no need to worry about the Swagger app making outbound requests directly, it can use the public API URL.


engrzakaug1989 (approved):
lgtm

jadudm (approved):
Looks good, and should fix Swagger.

PR #1120, opened 2023-05-13
Jadudm/xlsx gen spike

sambodeme:
## Description

- Updated code to generate xlsx files from xlsx definition templates.
- Updated test cases to use new xlsx templates.
- Updated json schema definitions.


sambodeme:
@jadudm This is good to go with respect to what we dicussed this morning.

eazapata23 (approved):
Tested locally and test for happy/bad passed for submitting upload with the new templates.

sambodeme:
@timoballard I combined the two PRs into this one.

PR #1132, opened 2023-05-18
Add FEDRAMP endpoint to .profile

asteel-gsa:
Per the guidance given by New Relic Support, since our account is FEDRAMP compliant, they asked us to add this host before attempting to escalate to engineering. 

Ref: [Fedramp Compliant Endpoints](https://docs.newrelic.com/docs/security/security-privacy/compliance/fedramp-compliant-endpoints/)


mogul (approved):
🤞😓🤞

PR #1139, opened 2023-05-19
Add `audit_type` to General Information

tadhg-ohiggins:
So audits have types
And we should really store them—
As should the schema.

-----

Actually store audit type, and other changes.

Making audit type work is the most significant fix for General Information, and that's what the edits to the schema file are for. Also has some edits taking advantage of Jsonnet and moving some things into `Base.libsonnet`.

Also makes the HTML output, via the Django template, actually reflect some of the other values that were being stored but not being shown.


backend/schemas/sources/Base.libsonnet
```diff
@@ -18,6 +18,9 @@ local Types = {
   array: { type: 'array' },
   number: { type: 'number' },
   NULL: { type: Const.NULL },
+  boolean: { type: 'boolean' },
```
jadudm, line 21: We'll pick this up in a future consolidation of schemas (we have them running around the entire codebase)... I've got a "Compound" type for compound types. These could, perhaps, be called "atomic" types. However, this is not worth changing now, as we need to refactor all of the schemas and get them in one place in the tree. We'll pick that up in a future sprint *real soon now*.

backend/schemas/sources/Base.libsonnet
```diff
@@ -102,6 +105,77 @@ local Enum = {
     ],
     title: 'EmptyString_Null',
   },
+  UnitedStatesStateAbbr: Types.string {
```
jadudm, line 108: This one we should probably ticket as a bug, or fix now... does this include all of the territories? I don't see PR, for example.

backend/schemas/sources/GeneralInformation.schema.jsonnet
```diff
@@ -1,148 +1,58 @@
...
-            "WI",
-            "WY"
-         ],
+      },
+      "State": Base.Enum.UnitedStatesStateAbbr {
          "title": "State",
-         "type": "string"
       },
-      "UEI": {
+      "UEI": Types.string_or_null {
```
jadudm, line 17: We'll catch this in a refactor; I have a more specific type, and we should probably not use a compound type as a base type. It buries optionality in a bad way. That is, if we want a field to be UEI_or_null, we should build that compound type, but not have the base type of UEI be optionally a string or null.

jadudm (approved):
There's things in our schema that I want to fix, but I want to story them out as a single piece of work. I noted a few things here, and will come back around to story them. Lets move this commit in, and pick up all the schema work in one place next sprint.

backend/schemas/sources/Base.libsonnet
tadhg-ohiggins: It does not, and I currently can't recall whether that's because it's how Census did it or because at some point we hurried through adding this list.

PR #1151, opened 2023-05-23
Use docker images from ghcr

JeanMarie-TTS:


.github/workflows/terraform-plan-env.yml
```diff
@@ -10,9 +10,32 @@ on:
   
 
 jobs:
+  fetch-versions:
```
mogul, line 13: Lines 13-34 look like they are just a copy of `workflows/fetch-versions.yml... This should be removed since you're calling that workflow and referring to its outputs below!

.github/workflows/terraform-plan-env.yml
```diff
@@ -23,6 +46,7 @@ jobs:
       TERRAFORM_PRE_RUN: |
           apt-get update
           apt-get install -y zip
+      TF_VAR_postgrest: ${{ needs.fetch-versions.outputs.postgrest }}
```
mogul, line 49: Noice! I didn't know you could refer to outputs this way. You should also be doing this in `terraform-apply-env.yml`.

terraform/dev/dev.tf
```diff
@@ -2,6 +2,7 @@ module "dev" {
   source                = "../shared/modules/env"
   cf_space_name         = "dev"
   new_relic_license_key = var.new_relic_license_key
+  postgrest             = var.postgrest
```
mogul, line 5:
```suggestion
  postgrest_image             = var.postgrest_image
```

terraform/dev/variables.tf
```diff
@@ -13,3 +13,8 @@ variable "new_relic_license_key" {
   type        = string
   description = "the license key to use when setting up the New Relic agent"
 }
+
+variable "postgrest" {
```
mogul, line 17:
```suggestion
variable "postgrest_image" {
```

terraform/dev/dev.tf
mogul: You'll need to add this to the `staging.tf` and `production.tf` files too.

terraform/shared/modules/env/postgrest.tf
```diff
@@ -1,5 +1,6 @@
 locals {
   postgrest_name = "postgrest"
+  postgrest_image = "ghcr.io/gsa-tts/fac/postgrest:${var.postgrest}"
```
mogul, line 3:
```suggestion
```

terraform/shared/modules/env/postgrest.tf
```diff
@@ -16,7 +17,7 @@ resource "cloudfoundry_service_key" "postgrest" {
 resource "cloudfoundry_app" "postgrest" {
   name         = local.postgrest_name
   space        = data.cloudfoundry_space.apps.id
-  docker_image = "postgrest/postgrest:v10.1.2"
+  docker_image = local.postgrest_image
```
mogul, line 20:
```suggestion
  docker_image = var.postgrest_image
```

terraform/shared/modules/env/variables.tf
```diff
@@ -66,3 +66,8 @@ variable "new_relic_license_key" {
   type        = string
   description = "the license key to use when setting up the New Relic agent"
 }
+
+variable "postgrest" {
```
mogul, line 70:
```suggestion
variable "postgrest_image" {
```

.github/workflows/terraform-plan-env.yml
```diff
@@ -31,11 +55,15 @@ jobs:
...
         # on the PR branch triggers a re-plan.
+      - name: show versions
+        run: |
+          echo "TF_VAR_postgrest=$TF_VAR_postgrest"
       - name: terraform plan ( ${{ inputs.environment }} )
         uses: dflook/terraform-plan@v1
         with:
           path: terraform/${{ inputs.environment }}
           label: ${{ inputs.environment }}
+
```
mogul, line 66:
```suggestion
```

mogul (changes requested):
Other than the copy-paste of `fetch-versions.yml` content into `terraform-plan-env.yml` this looks pretty good! 

- I made suggestions to just pass in the whole image specifier rather than hardcoding the registry location in the module and passing in the version, then constructing it. 

- If you take those suggestions, you'll want to modify the place where you set `TF_VAR_` in the `-plan-env` and `-apply-env` workflows to do the construction there.
- If you don't, please do at least rename the variables in terraform to indicate what they are, eg `postgrest_version`.

terraform/shared/modules/env/env.tf
```diff
@@ -45,7 +47,7 @@ module "clamav" {
 
   cf_org_name   = var.cf_org_name
   cf_space_name = var.cf_space_name
-  clamav_image  = "ajilaag/clamav-rest:20230228"
+  clamav_image  = local.clamav_image
```
mogul, line 50:
```suggestion
  clamav_image  = var.clamav_image
```

terraform/shared/modules/env/env.tf
```diff
@@ -1,5 +1,7 @@
 locals {
-  clam_name = "fac-av-${var.cf_space_name}"
+  clam_name    = "fac-av-${var.cf_space_name}"
+  clamav_image = var.clamav_image
```
mogul, line 3:
```suggestion
```

.github/workflows/terraform-plan-env.yml
```diff
@@ -31,6 +60,10 @@ jobs:
         # The post includes a link to the workflow run where the plan was
         # generated, and the same comment will be updated each time a new commit
         # on the PR branch triggers a re-plan.
+      - name: show versions
+        run: |
+          echo "TF_VAR_postgrest=$TF_VAR_postgrest_image"
+          echo "TF_VAR_clamav=$TF_VAR_clamav_image"
```
mogul, line 66:
```suggestion
          echo "TF_VAR_postgrest_image=$TF_VAR_postgrest_image"
          echo "TF_VAR_clamav_image=$TF_VAR_clamav_image"
```

mogul (changes requested):
Looking pretty good... In addition to the cleanup and bugfix here, I'd like to see the `fetch-versions` job factored back out!

terraform/shared/modules/env/env.tf
mogul: No need to indirect through a `local.` here, can just refer to the variable.

.github/workflows/terraform-plan-env.yml
mogul: This one's a bug... Terraform won't like supplying a variable value for a variable name that's not declared, and omitting a value for a variable where a value is required.

JeanMarie-TTS:
> Looking pretty good... In addition to the cleanup and bugfix here, I'd like to see the `fetch-versions` job factored back out!

Created #1157 as a follow-up

mogul (approved):

PR #1152, opened 2023-05-23
Paths/files/imports consolidation

tadhg-ohiggins:
Paths, imports, and files
Rearranged to make more sense—
At least, I hope so.

-----

Have everything derive from `BASE_DIR`, which is a `Path`, eliminating a bunch of casting and uses of `os.path.join`.

Move `audit` fixture/schema files into `data_fixtures` and `schemas`.

Eliminate some stray uses of `open` instead of `Path.read_text` along the way, and bits and pieces of other cleanup.


backend/config/settings.py
```diff
@@ -361,11 +362,12 @@
...
 SAM_API_KEY = secret("SAM_API_KEY")
 
-SCHEMAS_DIR = os.path.join("audit", "schemas")
-
-SECTION_SCHEMA_DIR = os.path.join("schemas", "output", "sections")
-
-XLSX_TEMPLATE_DIR = os.path.join("schemas", "output", "excel", "json")
+# Data/schema directories
+DATA_FIXTURES = BASE_DIR / "data_fixtures"
+SCHEMAS_DIR = BASE_DIR / "schemas" / "output" / "audit"
```
sambodeme, line 367: This could be renamed to reflect what we intend to store in that folder (SOMETHING_DIR). To me SCHEMAS_DIR implies  BASE_DIR / "schemas"

sambodeme (approved):

Issue #1071, opened 2023-05-02
Create and deploy an SMTP egress-proxy


mogul:
# At a glance

[comment]: # "Begin with a short summary so intent can be understood at a glance."

[comment]: # "In order to: some objective or value to be achieved"
[comment]: # "as a: stakeholder"
[comment]: # "I want: some new feature"

**In order to** enable FAC apps to send mail from a restricted space
**as a** FAC system architect
**I want** an SMTP proxy configured for apps to use that will only allow mail to our upstream SMTP provider.

# Acceptance Criteria

We use [DRY](https://docs.behat.org/en/latest/user_guide/writing_scenarios.html#backgrounds) [behavior-driven development](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications) wherever possible.

[comment]: # "ACs should be clearly demoable/verifiable whenever possible."
[comment]: # "Given: the initial context at the beginning of the scenario"
[comment]: # "when: the event that triggers the scenario"
[comment]: # "then: the expected outcome(s)"
[comment]: # "Repeat scenarios as needed, or repeat behaviors and lists within a scenario as needed."

[comment]: # "The scenario should be a short, plain language description."
[comment]: # "Feeling repetative? Apply the DRY (Don't Repeat Yourself) principle!"

### Scenario: 

**Given** an `smtp-proxy` app is deployed in the `-egress` space
**And** a network policy enables traffic from `gsa-fac` to the `smtp-proxy` app on port 8080
**And** a USPI containing SMTP creds for the proxy is bound to the `gsa-fac` app
**And** I have `cf ssh`'d into the `gsa-fac` app
**And** I have extracted the SMTP creds from VCAP_SERVICES
**when** I send mail using the extracted credentials
...

[comment]: # "Each task should be a verifiable outcome"
```[tasklist]
### then...
- [x] I can see the mail is delivered to the target address
```

### Shepherd

[comment]: # "@ mention shepherds as we move across the board."

* Design shepherd: N/A
* Engineering shepherd: Bret

# Background

[comment]: # "Any helpful contextual notes or links to artifacts/evidence, if needed"

# Security Considerations

Required per [CM-4](https://nvd.nist.gov/800-53/Rev4/control/CM-4).

[comment]: # "Our SSP says 'The team ensures security implications are considered as part of the agile requirements refinement process by including a section in the issue template used as a basis for new work.'"
[comment]: # "Please do not remove this section without care."
[...]

PR #1156, opened 2023-05-24
Implement SMTP egress proxy

mogul:
Relates to #1071.


JeanMarie-TTS:
It would be good to add some test scripts to show that we can send emails via the GSA SMTP service  and that we cannot do so via some other SMTP service provider.

jadudm:
Does "it would be good to..." mean:

1. I think someone/the ticket originator should do this as part of this ticket/PR,
2. I think we should ticket it separately, OR
3. I think this is a good idea, and I'll do it?

JeanMarie-TTS:
"It would be good" means this is a recommendation to the PR owner. They can choose to implement as part of the PR, say this does not make sense and why, create a ticket for later implementation, etc.

mogul:
> It would be good to add some test scripts to show that we can send emails via the GSA SMTP service and that we cannot do so via some other SMTP service provider.

Unlike the `https-proxy`, the `stream-proxy` only ever passes bytes between the client and [the specific preconfigured upstream host+port](https://github.com/GSA-TTS/FAC/pull/1156/files#diff-4fc62bb315d8ddecd4d3a66f85038573ad8c2fa3811467c8c8072b1ffc87c7c3R20) (the "stream")... There's no logic to filter requests based on a requested destination, because the client can't even specify a destination! 

So it wouldn't make sense to add a test specifically for the GSA SMTP service.

mogul:
Let me know if y'all want a walkthrough of this before approving/merging!

asteel-gsa:
> Let me know if y'all want a walkthrough of this before approving/merging!

Happy to do a walkthrough if you have time in the next hour, and can approve.

JeanMarie-TTS:
LGTM - Will leave the approval to Alex after the walkthrough.

JeanMarie-TTS:
> > It would be good to add some test scripts to show that we can send emails via the GSA SMTP service and that we cannot do so via some other SMTP service provider.
> 
> Unlike the `https-proxy`, the `stream-proxy` only ever passes bytes between the client and [the specific preconfigured upstream host+port](https://github.com/GSA-TTS/FAC/pull/1156/files#diff-4fc62bb315d8ddecd4d3a66f85038573ad8c2fa3811467c8c8072b1ffc87c7c3R20) (the "stream")... There's no logic to filter requests based on a requested destination, because the client can't even specify a destination!
> 
> So it wouldn't make sense to add a test specifically for the GSA SMTP service.

Good for now. But I think we should create a test to send an email as a follow yo ticket.. This will give us the confidence that the integration is working. We should not be having paths with no test coverage.

asteel-gsa (approved):

mogul:
Note from walkthrough with Alex: I'm going to add a Mermaid diagram showing how the test works, since it's not obvious that the client app is acting both as a client and as the proxy's upstream. (Essentially, the client app is a mock.)

mogul:
> > So it wouldn't make sense to add a test specifically for the GSA SMTP service.
> 
> Good for now. But I think we should create a test to send an email as a follow yo ticket.. This will give us the confidence that the integration is working. We should not be having paths with no test coverage.

If you want to be able to prove that the _application_ sends mails, and that they will be delivered to the right location when they're submitted to the G Suite SMTP Relay, that's pretty out of scope for what the proxy actually does... The sole function of the proxy is to enable controlling/funneling egress traffic. 

What you're describing is more of an integration test. I was originally going to do that with mailhog as the mock, but once I started working on it I realized that I was testing a lot more than what the module actually does. I do think there's a call for mailhog to be integrated into the `docker-compose` stack with tests there for the application sending mails as expected, but that's a whole `COULD` issue.

mogul:
(I invite you to contemplate what kind of controlled test setup would be necessary to send a mail through the G Suite SMTP Relay _and_ verify that it was delivered, even without the proxy in the loop. 😜)

mogul:
> I'm going to add a Mermaid diagram showing how the test works, since it's not obvious that the client app is acting both as a client and as the proxy's upstream. (Essentially, the client app is a mock.)

[I added that diagram.](https://github.com/GSA-TTS/FAC/pull/1156/files?short_path=4f09979#diff-4f099791113d6c4fb834e76b77963083079a5f276e5ca43f0047bd7f41490dfe) I also reordered the commits to make them easier to follow in sequence.

PR #1163, opened 2023-05-25
Sambodeme/consolidate field types

sambodeme:
## Description

Using function to introduce compound type.


backend/schemas/source/excel/templates/federal-awards-expended-template.jsonnet
```diff
@@ -37,6 +37,7 @@ local open_ranges_defns = [
   [Sheets.open_range, {}, 'Cluster Name', 'cluster_name'],
   [Sheets.open_range, {}, 'If State Cluster, Enter State Cluster Name', 'state_cluster_name'],
   [Sheets.open_range, {}, 'If Other Cluster, Enter Other Cluster Name', 'other_cluster_name'],
+  // 20230525 HDMS FIXME: A formula to auto calculate federal_program_total in the excel is missing (see instructions in census template)!!!
```
sambodeme, line 40: There is an instruction related to this field in census FederalAwardsExpended.xlsx template but I am not sure how we are going to achive this.

tadhg-ohiggins (commented):
The quotation mark/whitespace changes in `GeneralInformation.schema.jsonnet` make it difficult to review; were there any meaningful changes in that file or are they purely formatting changes?

Also, I think you may have checked in files with Windows line endings by mistake, as unless I enable the “ignore whitespace” functionality on GitHub, a number of files show as having changed every line.

backend/schemas/source/base/Base.libsonnet
```diff
@@ -1,465 +1,453 @@
...
-      'submitted',
-      'received',
-      'available',
-    ],
-    title: 'SubmissionStatus',
-  },
-  ALNPrefixes: Types.string {
-    description: 'Valid two-digit program numbers; part of the CFDA/ALN',
-    enum: [
-      '1',
```
tadhg-ohiggins, line 236: Did you mean to remove these? I thought they were going to be two-digit with leading zeroes to match formulas like [this one](https://github.com/GSA-TTS/FAC/blob/c04acb04167ce86b44514e6f686739d283bace63/backend/schemas/output/excel/json/federal-awards-expended-template.json#L35). Or am I missing a change elsewhere covering that?

tadhg-ohiggins:
> > The quotation mark/whitespace changes in `GeneralInformation.schema.jsonnet` make it difficult to review; were there any meaningful changes in that file or are they purely formatting changes?
> > Also, I think you may have checked in files with Windows line endings by mistake, as unless I enable the “ignore whitespace” functionality on GitHub, a number of files show as having changed every line.
> 
> No meaningful change here. I just used jsonnetfmt cmd to make sure all jsonnet files have the same format. Good catch on the end of line change (I just pushed an update)

That's a good idea, but we should put formatting changes like that one in separate PRs so that it's clear no logic changes are involved.

backend/schemas/source/excel/templates/federal-awards-expended-template.jsonnet
jadudm: I think this has been ticketed separately at this point, so we can track it/address it. Leaving this here for now is fine as a reminder. We've decided (in conversation) there's more than one way to solve this, and none of them look scary/look like blockers. We'll leave the FIXME as a reminder in the codebase.

jadudm (approved):
Again, thank you, and looks great. This is the "other half" of the schema cleanup, and it looks great. Again, putting us forward as we look at additional improvements/changes.

PR #1164, opened 2023-05-25
Add dissemination API validator and schemas

danswick:
I picked up @jadudm's experimentation branch, added schemas for the remaining views, and made some small changes to the validator's logic. This is still pretty rough, but gives us a starting point to work from. 

How to test locally:
1. Fire up the dockerized app with `docker compose up`.
2. Pick a view to validate and construct your command. For example:
    ```bash
    docker compose run web python manage.py validate_api --view vw_auditor
    ````
    a. You might want to write the validation results to a text file. There's an example, `validate_log.txt` in this draft PR.
3. All of the views (except `revision`) should be valid according to the schemas in this draft PR. You might try changing a schema around to see what kinds of failures are triggered.

@eazapata23 I left you some wayfinding comments inline since you'll be working on validation over the next couple of weeks.


backend/data_distro/api_tests/fac.py
```diff
@@ -0,0 +1,107 @@
+import os
+import requests
+import typing
+from typing import List, Set, Dict, Tuple
+
+URI="https://api.data.gov/TEST/audit-clearinghouse/v0/dissemination"
+if os.getenv("API_GOV_URL") is not None:
+    URI = os.getenv("API_GOV_URL")
```
danswick, line 8: @eazapata23 these env variables will need to exist on your local system.

backend/data_distro/api_tests/inputs/auditee_0.json
```diff
@@ -0,0 +1,36 @@
+
```
danswick, line 1: I don't think we need any of these `{view}_0.json` files. I believe they were mostly there for setup and I just followed the pattern since it seemed easiest!

backend/data_distro/api_tests/schema/APITestBase.libsonnet
```diff
@@ -0,0 +1,116 @@
...
+local phone_regex = '[1-9]{1}[0-9]{9}+';
+local e164_regex = '^\\+[0-9]{1,3}[ ]*[0-9]{2,3}[ ]*[0-9]{2,3}[ ]*[0-9]{4}|^\\+[0-9]{1,3}[ ]*[0-9]{1,14}([ ]*[0-9]{1,13})?|^\\([0-9]{3}\\)[ ]*[0-9]{3}[ ]*[0-9]{4}?';
+local email_regex = '^[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+(?:\\.[a-zA-Z0-9!#$%&\'*+/=?^_`{|}~-]+)*@(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\\.)+[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$';
+
+// Extend the Base.Types object.
+// Lets us transparently use the additional types in the
+// API validations. Might want to move some of these back to the base.
+// Also suggests we need to rethink/reorganize all of the validation/schema things.
+Base {
+  Types: Base.Types {
```
danswick, line 63: @eazapata23 see the list of custom types below. I missed most of these while quickly setting up the schemas.

backend/data_distro/api_tests/schema/APITestBase.libsonnet
```diff
@@ -0,0 +1,116 @@
...
+    },
+    ZIP: type_zipcode,
+    DBKEY: Base.Types.string {
+      pattern: REGEX_DBKEY,
+    },
+    IntegerArray: {
+      type: 'array',
+      items: Types.integer
+    },
+    IntegerArrayOrNull: {
```
danswick, line 104: I added this, but don't feel good about it.

backend/data_distro/api_tests/schema/Makefile
```diff
@@ -0,0 +1,6 @@
+all:
```
danswick, line 1: Once we move everything into the centralized schema directory, we rely on [that makefile](https://github.com/GSA-TTS/FAC/blob/main/backend/schemas/Makefile) instead. This one isn't needed for anything anyway.

backend/data_distro/api_tests/schema/auditor.jsonnet
```diff
@@ -0,0 +1,26 @@
+local Base = import 'APITestBase.libsonnet';
+local Types = Base.Types;
+
+Types.object {
+  properties: {
+    id: Types.integer,
+    cpa_phone: Types.integer, // should this be a string?
```
danswick, line 7: @eazapata23 there are custom types for a lot of these "contact info" fields that will need to be added and verified.

backend/data_distro/api_tests/schema/passthrough.jsonnet
```diff
@@ -0,0 +1,15 @@
+local Base = import 'APITestBase.libsonnet';
+local Types = Base.Types;
+
+Types.object {
+  properties: {
+    id: Types.integer,
+    passthrough_name: Types.string,
+    passthrough_id: Types.StringOrNull, // This doesn't seem like something that should be optional, but there are nulls in the data
```
danswick, line 8: There are a few cases like this, where it doesn't seem like `null` should be possible, but there are multiple `null`s in the data.

backend/data_distro/api_tests/schema/revision.jsonnet
```diff
@@ -0,0 +1,32 @@
+local Base = import 'APITestBase.libsonnet';
+local Types = Base.Types;
+
+// LOTS of optional fields here.
```
danswick, line 4: @eazapata23 this schema is probably the worst offender, but there are a _lot_ of optional fields throughout. If you can, sync with @jadudm or some folks working on the XLS side about these. My understanding is that the new submission flow will make lots of fields mandatory.

backend/data_distro/api_tests/schema/validate_log.txt
```diff
@@ -0,0 +1,1456 @@
+{'verbosity': 1, 'settings': None, 'pythonpath': None, 'traceback': True, 'no_color': False, 'force_color': False, 'skip_checks': False, 'view': 'vw_findings', 'year': 2022, 'start': 0, 'end': 100, 'step': None}
```
danswick, line 1: @eazapata23 I left this file in here to give you an idea about what you should see coming back from the validator. The first line should always give you a readout of the arguments you're passing to the validator.

backend/data_distro/api_tests/schema/validate_log.txt
```diff
@@ -0,0 +1,1456 @@
...
+
```
danswick, line 2: 👇 

Because we validate API responses one-by-one, the validation error messages are organized by `'id'`. It will print the object id, then the object itself, then an errors section with an ordered list of validation error messages.

backend/data_distro/api_tests/schema/validate_log.txt
```diff
@@ -0,0 +1,1456 @@
...
+------
+1. None is not of type 'string'
+
+Failed validating 'type' in schema['properties']['prior_finding_ref_numbers']:
+    {'type': 'string'}
+
+On instance['prior_finding_ref_numbers']:
+    None
+
+2. None is not of type 'string'
```
danswick, line 931: Here's an example of an object that had two different types fail.

backend/data_distro/api_tests/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+        for k, v in api_object.items():    
+            if k in _transforms[api_view]:
+                obj[k] = _transforms[api_view][k](v)
+            else:
+                obj[k] = v
+    else:
+        obj = api_object
+    
+    validator = jsonschema.Draft7Validator(get_schema(api_view))
+    errors = validator.iter_errors(obj)
```
danswick, line 64: `validator.iter_errors()` returns an iterable with all of the validation errors for a given "validation run."

backend/data_distro/api_tests/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+            raise e
+        schema_cache[api_view] = json.loads(json_str)    
+    schema = schema_cache[api_view]
+    return schema
+
+def validate_api_object(api_view, api_object):
+    validation_error = None
+    # This is not what we should do.
+    obj = {}
+    if api_view in _transforms:
```
danswick, line 54: I added this check to get the validator running (it was failing on everything except `vw_auditee`, but this is really a temporary workaround.

backend/data_distro/management/commands/validate_api.py
```diff
@@ -0,0 +1,88 @@
...
+{oe_object}
+
+------
+ERRORS
+------
+{messages}
+"""
+
+            print(content.format( oe_id=oe_id, oe_object=oe["object"], messages=messages))
+        return f"failure_count: {failure_count}"
```
danswick, line 88: I made a bunch of changes above so we could print multiple validation errors for a given object. Please excuse and/or issue stern corrections for any poor python etiquette.

backend/data_distro/mappings/FAC_data_dict.csv
danswick, line 1: I don't think this should have been deleted. This may have happened when I caught @jadudm's branch up to `main`.

backend/data_distro/api_tests/schema/passthrough.jsonnet
jadudm: I think, with this (and the question about custom types above) that we might ticket out separately. That is, the schema refactor/cleanup team have a number of things that can/should be refined/improved. However, *as a first increment*, getting all the schemas in one place, and... *enbettered*... is good.

I would say before closing out the ticket, we should either 1) start a new one, or 2) tag on to a ticket with the other schema refactor team, and 3) add checklist items for exactly these kinds of questions. E.g. "Revisit all fields that contain null, and decide whether `null` should be allowed" or similar. It pushes the question down the road a bit, but it also lifts it up into view, and forces us to address it.

In other words, I think these schemas do more good *in the tree* and ready to use *in an incremental state* than halting forward progress to solve/answer these questions.

backend/data_distro/api_tests/schema/revision.jsonnet
jadudm: Again, I'd be happy with a ticket that breaks every one of these out as a checklist where we have to come back and revisit the schema. How we reconcile new data (that might not be optional) vs. old data (which might have `null`s) might require a schema that says "if the year field is 2022, or 2021, or..., then this field *might* be `null`" whereas something checking 2023 might be different. 

That said, exactly how we chose to tackle that is not something I want to solve now. I *do* want us to have a ticket that explicitly lists out each schema, and forces us to revisit them and start answering these "harder" questions. So, @danswick is not wrong... we want to address this. However, I'd rather move this PR through and into the tree, and ticket it for a future sprint/increment.

backend/data_distro/management/commands/validate_api.py
jadudm: That `validate_api` script was written by a non-pythonista, so... :shrug:. For now, it's a good thing that we're doing it.

danswick:
Priorities have changed since I opened this PR, so I'm closing here.

Issue #1140, opened 2023-05-19
Refactor schemas in-tree and in-situ


jadudm:
### Tasks

First set of related work:

```[tasklist]
- [x] Get all of our schema-related work into one place in the tree. #1150
- [x] Create base, atomic types (e.g. `string`, `integer`) to build off of. #1163/#1165
- [x] Create compound types (e.g. `string_or_null`) for use. See note number 2. #1163/#1165
- [x] Do not use union types that have `NULL` as a component as base types anywhere (e.g. do not build objects off of `string_or_null`). #1163/#1165
- [x] Integrate refined checks for complex data (e.g. use new/revised/refined UEI check). #1163/#1165
```

The following tasks are more appropriately tackled once we have a solid base from the first set:

```[tasklist]
- [ ] Look at fields that might have standards, and move to standards. Note number 3. #1180 
- [ ] Update random tests to include some targeted tests. For example, we have some random tests for phone numbers, etc., against our regexes. I think we should instead try and do a few targeted edge tests. The fuzz tests are fun (and can be kept), but they're not as useful as checking the boundaries (as best we can) on our regexes (for example). #1181 
- [ ] Walk every field and verify where `null` is, or is not, allowed/appropriate. See note  number 1. #1182 
- [ ] Make sure all lists are complete/correct (e.g. state abbreviations currently seem to exclude the territories?) #1180
```

Notes:

1. I would prefer to never allow null. I would prefer types that were of the shape `UEI_or_NA`. That is, I would like to move to a world where every field *always* has an input. There may be exceptions... for example, the second address line in a two-line address. However, at that point, we should probably assert that *at least one* of the two address fields **must** have a string value. Further, do address fields have a minimum length? What is the shortest valid entry in an address field? (*How many angels can dance on the head of a pin*, while we're at it.) Although pedantic, I think I would rather people submit helpdesk tickets where we discover, in practice, that our typing is *too robust*, and we have to relax it, than to have typing that is *too lax*, and we end up (quietly) with data that we validate, but wish we hadn't.
2. We can automate this in the library. I'm OK with creating a list of types that we want `_or_null`, writing a small function to map over that list, and it returns an object that creates the compound types for all of them. In fact, that would be a good idea. That said, I'd also like us to minimize use of `_or_null` types throughout the schemas.
[...]

PR #1165, opened 2023-05-26
Jsonnet type improvements

tadhg-ohiggins:
One or many types—
Have to handle either case
`flatMap` for the win.

-----

*   Make `compound_type` deal with either string or array values
*   Add a function for adding the types from an array of objects-with-type-properties to an object, and make it also handle string or array `type` values.
*   Move away from basing types on `_or_null` constructs and instead use functions to add in `Types.NULL`.
*   Make UEI and EIN in General Information validation use types defined in `Base.libsonnet`.
*   Move Zip/Phone validation rules to one location.


tadhg-ohiggins:
@sambodeme I think that once you change the line endings in `Base.libsonnet` in your branch back to `LF`, the diff from this branch should be easy to read.

backend/audit/test_schemas.py
```diff
@@ -138,30 +138,33 @@ def test_invalid_uei(self):
...
-            with_punc,  # contains a non-alphanum char
-            with_ioIO,  # contains one of i, o, I, O
+        with_numlike = good_uei[:idx] + choice("ioIO") + good_uei[idx + 1 :]
+
+        bad_ueis_and_messages = [
+            (too_short, f"'{too_short}' is too short"),
+            (too_long, f"'{too_long}' is too long"),
+            (zero_start, "does not match"),
+            (with_punc, "does not match"),
+            (with_numlike, "does not match"),
```
jadudm, line 157: While we're here, and before we merge, it would be good to add some tests that make sure we never have 9 digits in a row. We could do it randomly, or 

1. One check for a UEI that starts with 9 digits
2. One check for a UEI containing 9 digits in the middle
3. One check for a UEI that ends with 9 digits

Those are all invalid, as a UEI should never contain something that could be confused for an EIN or SSN. Hence, the rule about not having 9 digits in a row.

I *think* our regexs catch this. However, I'm not 100% certain. I'd feel better knowing we handled those three cases, as the first and third represent the "edges," so to speak.

backend/schemas/source/base/Base.libsonnet
```diff
@@ -445,8 +447,8 @@ local SchemaBase = Types.object {
   Atoms: Atoms,
   Meta: Meta,
   Enum: Enum,
-  Compound: Compound + {
-    UEI: type_uei
+  Compound: Compound {
+    UEI: type_uei,
```
jadudm, line 451: Why is `type_uei` here and on line 426 of Base.libsonnet?

backend/schemas/source/base/Base.libsonnet
```diff
@@ -129,75 +127,75 @@ local Enum = {
...
-        'UT',
-        'VT',
-        'VA',
-        'WA',
-        'WV',
-        'WI',
-        'WY'
-     ],
+    enum: [
+      'AL',
```
sambodeme, line 131: Unrelated to this PR but I came across this yesterday and would like your thoughts @tadhg-ohiggins @jadudm. Should not we add DC and U.S. territories to this list ?

tadhg-ohiggins: We've discussed it, and the issue is that we don't know what the actual full list of codes to accept is just yet. I think of this as part of the “Look at fields that might have standards, and move to standards.” item of the second chunk of #1140, which should probably be its own ticket at this point.

sambodeme (approved):

jadudm (approved):
This looks great to me. It's an excellent cleanup increment, and sets us up for next steps/refinement. Thank you.

Before we close the parent ticket, I'd like (for me, or if one of you get to it sooner, you) create a new ticket for the "next steps" things that we thought would be good to come around to (e.g. states/territories list check, etc.) so that we begin a checklist of things we know we want to revisit (while it is fresh in our minds).

Issue #1170, opened 2023-05-30
Epic: Search


jadudm:
# At a glance

[comment]: # "Begin with a short summary so intent can be understood at a glance."

[comment]: # "In order to: some objective or value to be achieved"
[comment]: # "as a: stakeholder"
[comment]: # "I want: some new feature"

**In order to** search for previously completed audits
**as a** user of the FAC 
**I want** to be able to search for audits.

# Acceptance Criteria

ARO = Audit Resolution Official (Federal employee at an agency)

### Scenario: An ARO is looking for  prior audits 

**Given**  that I have a UEI in hand 
**when** I search for that UEI
...

[comment]: # "Each task should be a verifiable outcome"
```[tasklist]
### then... 
- [ ] https://github.com/GSA-TTS/FAC/issues/1231
- [ ] https://github.com/GSA-TTS/FAC/issues/1233
```

Some of the design/UX work is detailed in #1148. We're closing that ticket for now, but also know we need to refine the above items; 1148 will be reflected in that work. 

### Scenario: A back-end engineer is supporting search implementation

The backend integration will work via communication with the Postgrest API—not using internal Django access to the database.

```[tasklist]
### then
- [ ] as a backend engineer, implement a proxy URL for the front-end to query.
- [ ] as a backend engineer, connect the proxy URL to the Postgrest server. (Essentially `djangoappurl.gov/api/query/postgreststuff` -> `postgrestserviceurl.whatever/postgreststuff`)
- [ ] as a backend engineer, work with the front-end engineer to implement new queries as needed to support the front-end implementation.
- [ ] as a backend engineer, work with the FE engineers/design to present the Search page
- [ ] as a backend engineer, work with the FE engineer to implement a view(s) for supporting the query process
```

### Scenario: backend/frontend engineers are trying to make search "just work"

```[tasklist]
### then...
- [ ] the frontend bundles up the search and filters as an object to send to the backend; this is a "plain Django form submission"
- [ ] the backend pulls this object apart, and does the query(s) that are necessary against PostgREST to build up a result set
- [ ] results are passed back to the FE for display. Pagination... exists within PostgREST, therefore should be implemented transparently to leverage that functionality. (Track pagination in the FE; it should be passed as part of the query object)
- [ ] fe/be and design must design a result set object that allows for population of result set cards in the UI
```

See also #1149 for some conversation that pertains to the above items (in particular, w.r.t. testing and how Django will interact with the FE and PostgREST.)

[...]

PR #1225, opened 2023-06-07
Super Simple Search

jperson1:
# Simple Search - #1170

## Changes
`/search` is a new route, under a new app, with some functionality.

1. `/search` is the search home page, linked to in the header. It contains the simple search functionality with a few fields. Form works, but currently does nothing but redirect to itself. 
   - UEI or EIN
   - ALN/CFDA
   - Fiscal Year
   - Acceptance date

2. `/search/table?query` gets and displays search results from the API. This string works **one-to-one** with the API, with two exceptions. 
   - This means a request made to
   `http://localhost:8000/search/vw_auditee?limit=20&auditee_city=eq.ARLINGTON` 
   should display the exact same results as we would get in JSON by requesting from 
   `https://api.data.gov/TEST/audit-clearinghouse/v0/dissemination/vw_auditee?limit=20&auditee_city=eq.ARLINGTON`
   - The exceptions: visiting pages made with `/vw_general` or `/vw_federal_award` will trigger an extra query. This extra query cross-references the previous results with the `/vw_auditee` table, so as to gain access to things like `auditee_city`
      - This means that requests to other endpoints will not display correctly on the front end, which is fine. It is receiving the data, just not the kind it wants to show.
      - This will be changed in the future on a layer further back than Django, where a simple "LEFT JOIN" can solve the problem. It will also make sorting easier, yay!

3. There exists a session-level variable that can tell whether you've accepted the Uniform Guidance Acknowledgement. The modal that controls this variable is available on the bottom of the search results. In the future, it will pop up in front of users once, and because of this variable, will not be seen again. 

## How to test
1. Follow the instructions to get an api key [here](https://github.com/GSA-TTS/FAC/discussions/1066). Once approved, define it as `API_GOV_KEY` wherever you keep your environment variables. Also, set `API_GOV_URL` to `"https://api.data.gov/TEST/audit-clearinghouse/v0/dissemination"`. Eventually, this will not be needed, as Django will ping the local database directly. For now, we use the online API. 
2. Swap to this branch and `docker compose up`. It shouldn't require a rebuild or anything (unless you haven't in a while, in which case the postgreSQL update will be in the way).
3. Check out [http://localhost:8000/search/](http://localhost:8000/search/). Here are some searches you can try:
    - Sort by "UEI", and enter `HWG4ML2D9EL6`. One result. [Link](http://localhost:8000/search/vw_auditee?uei_list=cd.%7B%20HWG4ML2D9EL6%20%7D&limit=100)
    - Sort by "EIN", and enter `730785251`. One result. [Link](http://localhost:8000/search/vw_auditee?ein_list=cd.%7B%20730785251%20%7D&limit=100)
    - Sort by "Audit Year", and enter `2022`. One hundred different results. [Link](http://localhost:8000/search/vw_general?audit_year=eq.2022&limit=100)
    - Sort by "ALN", and enter `93.600`. One hundred results, with many repeats. [Link](http://localhost:8000/search/vw_federal_award?agency_cfda=eq.93.600&limit=100)
4. Also try some other things, like manually changing the "limit" value in the query. It shouldn't let you go above 1000, because that did give some response time errors. 

## Things to be done for this PR (probably/maybe):
1. GET AN API KEY FOR STAGING/DEV TO USE BEFORE MERGING
2. Fix formatting on things, finalize language.
3. Include the search component on the results page.
4. Show currently used filters in the results.
5. Make the modal pop up on pageload, if the user's not seen it before.

Edit: Updated search filters. 

## Screenshots
image
image
image


jperson1:
Breaking changes added in. Frontend is more accurate to the desired design. Searching just redirects top itself.

Breaking things here is okay, since we're swapping into a new workflow and coming back to this later (perhaps even post-MVP).

JeanMarie-TTS (changes requested):
Is the django view trying to access postGREST by connecting to api.data.gov? It should be connecting directly to postGREST. Search should work even if api.data.gov is unreachable.

jadudm:
This ticket hasn't been touched in a while.

We did decide, at some point, that we would talk to PGrest
container-to-container instead of going out through data.gov. So, yes.

tadhg-ohiggins:
Is this PR still relevant or has it been shelved?

PR #1247, opened 2023-06-12
Jmm/data model

JeanMarie-TTS:
High level dissemination data model. The relationships are asserted to be true and not managed through django relationships.
Tweaks to system architecture diagram included here.


sambodeme (commented):
I thought we were moving away from this data model (e.g., entities like FederalAward would not exist in the new model). Maybe I am missing something. Is this the new model we're aiming for?

JeanMarie-TTS:
@sambodeme This is the dissemination model. We decided to not normalize the model by having entities like Auditee and Auditor. We still need to have General, FederalAward, Findin, FindingText, etc.

sambodeme:
> @sambodeme This is the dissemination model. We decided to not normalize the model by having entities like Auditee and Auditor. We still need to have General, FederalAward, Findin, FindingText, etc.

Ok, thanks. Sounds like I registered something different. For some reason, I thought the plan was to distribute the data as it is, meaning that for the MVP API, the views would mirror the table structures from Census that users are familiar with. And in subsequent iterations, we would implement modelisation.

jadudm (approved):
Looks good.

gsa-suk:
I noticed these messages just now....
Looking through the ERD from the tables created by the app and comparing the puml file, these are the following differences I see:

- one to many relationship from FederalAward to Finding (not the other way)
- one to many relationship from General to Findings_text (not from Finding to Findings_text)
- one to many relationship from General to Cap_text (not from Finding to Cap_text)
- many to optional (diamond) relationship from General to Revision

Is the puml file a representation of what the app should be doing?

JeanMarie-TTS:
@jadudm We have an ERD in the tt/1146 branch.  This is still work in progress. This model represents the data model that G-FAC will use for dissemination, it's not a model of what we get in the CSV files or even of the C-FAC data model.
The django model in backend/dissemination will be an implementation of this ERD. The ETL in G_FAC will populate this model.

PR #1264, opened 2023-06-15
We're not USWDS

tadhg-ohiggins:
Since we are the FAC
Our bug templates should say so.
The perils of paste.

-----

Remove reference to USWDS in our bug report GitHub issue template.

Feel free to make this ticket a key battleground in various struggles about whether or not we should use the definite article, spell out the full acronym, specify “GSA”, etc.


jadudm:
Thank you. I have been meaning to fix that.

jadudm (approved):
LGTM. I introduced this bug. Thank you for the fix.

PR #1266, opened 2023-06-15
Upload Page Row Errors Display

jperson1:
# Upload Page Row Errors Display

## Changes:

- On XLSX file upload, schema validation errors send back a list of issues in the format (Row, Column, Filed, Help text, Link). These issues are now displayed as a table in the info box below the file input. 
- The link doesn't go anywhere, so it's currently omitted.
- The audit home page (`/audit/submission-progress`) has all SF-SAC links unlocked. 
- Completing the general information form redirects to the audit home page instead of the federal awards upload page.

## How to test:

Switch to this branch and run normally. Create an audit, if you have none saved locally. Navigate to an upload page, and attempt to upload a spreadsheet. As XLSX is still somewhat under development, you may see things like "Ensure you have uploaded the correct template, or contact an administrator." This happens in cases where a whole field is missing, such as if "auditee_uei" is not included in the template you're using. Once you have everything right (except some row errors), you may see something like this:

image

That's what we're looking for.


backend/static/js/upload-page.js
```diff
@@ -36,12 +36,45 @@ function handleUploadErrors(error) {
...
+      <td class="text-center">${row_array[1]}</p>
+      <td>${row_array[2]}</p>
+    </tr>`;
+  }
+  const validationTable = `<p>Error on validation. Check the following cells for errors, and re-upload. 
+  Common errors include incorrect data types or missing information.</p>
+  <table class="usa-table usa-table--striped">
+    <thead>
+      <tr>
+        <th scope="col">Row</th>
```
sambodeme, line 68: @jperson1  Looks like Row and Column are in reverse order.

jadudm (approved):
I had a chance to walk the code live, and this looks good on review. We will be revisiting some of this based on user feedback, so this is a good first step to get us into user testing.

PR #1270, opened 2023-06-15
Jadudm/add vscode debugging support

jadudm:
This may/should fix the local stack issues based on work w/ Alex, Tim yesterday.


backend/docker-compose.yml
```diff
@@ -30,16 +30,18 @@ services:
...
     ports:
       - "8000:8000"
+      - "5678:5678"
     volumes:
       - .:/src
       - /src/node_modules
       - /src/staticfiles
   clamav-rest:
-    image: ghcr.io/gsa-tts/fac/clamav:latest
+    image: ghcr.io/gsa-tts/clamav-rest/clamav:latest
```
timoballard, line 44: flagging this for @asteel-gsa - I _think_ the issue with the `/fac/` image was resolved, so we don't need to do this anymore

asteel-gsa: That is correct yes. As we have it set now, these 2 images are correctly parallel.
GSA-TTS/clamav-rest will be built and published `12:30a EST | 9:30p PST | 4:00a UTC` on sunday
GSA-TTS/FAC will be built and published `4:00a EST | 1:30a PST | 8:00a UTC` on sunday

Per Jadud - `That way, if upstream breaks, we might see it coming for the production env.` with the upstream being GSA-TTS/Clamav-Rest.

The benefit to the FAC one, is that it is trivy scanned. So if a vuln sneaks in, it wont actually publish to FAC because the trivy scan will fail when we publish to FAC packages, so the `/fac./` image will *theoretically* always be safe 

@timoballard @jadudm

mogul: Shouldn't this be `ghcr.io/gsa-tts/fac/clamav:latest` in order to make sure we're using the scanned one whenever we can?

asteel-gsa: set back to fac package and merged main into branch

PR #1273, opened 2023-06-16
Refactor form section name handling

tadhg-ohiggins:
Python has tuples.
And named tuples are better.
So let’s use them here.

-----

I have to do some things with form sections as part of making Excel files have singular names, and am separating this bit of refactoring for that out.

This makes importing and handling the name of the sections easier.

It also changes the URL paths for the sections (that is, the sections under `/audit/excel/`) to bring them into line with the casing of our URLs.

### DIFF
--- a/backend/audit/fixtures/excel.py
+++ b/backend/audit/fixtures/excel.py
@@ -1,3 +1,4 @@
+from collections import namedtuple
 from django.conf import settings
 
 SHEETS_DIR = settings.XLSX_TEMPLATE_SHEET_DIR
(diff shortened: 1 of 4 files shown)


sambodeme:
@tadhg-ohiggins I am getting 5 failures when running test on this branch. It does not seem that it is related to this change but I am looking into it.

tadhg-ohiggins:
@sambodeme could you paste the errors you're getting here or in Slack?

sambodeme:
I just posted that in slack, will pick the conversation when I get back

sambodeme (approved):
Looks neat!

PR #1276, opened 2023-06-16
Sambodeme/update etl error message

sambodeme:
## Description

Updated code to utilize the help field within the template definition for generating error messages that are more user-friendly.


jadudm (approved):
Looks good. It seems like most of the work was in the test updates, and the code to send content to the frontend was minimal in the end. Thank you.

PR #1288, opened 2023-06-19
Make Excel filenames use report ID and section name, and nothing else

tadhg-ohiggins:
You can’t trust users.
Why do we let them name files?
Let’s do it ourselves.

-----

Excel files are now named with `report_id` and the name of the form section they’re uploaded to. The original filenames are discarded.

Since I added a new field for `ExcelFile`, `form_section`, this PR includes the Django two-step of a migration with a default value for the new field, and then a migration removing any default value (since we don’t actually want one).

@timoballard I think that this approach to saving the files should be basically the same for PDFs; let me know if you think it should be different:

```python
def save(self, *args, **kwargs):
    report_id = SingleAuditChecklist.objects.get(id=self.sac.id).report_id
    self.filename = f"{report_id}--{self.form_section}.xlsx"
    super(ExcelFile, self).save(*args, **kwargs)
```


jadudm (approved):
Looks good. Replaces a fragile piece of user input -- the filename -- with a name we generate.

PR #1310, opened 2023-06-22
Set transition dates using parallel arrays

JeanMarie-TTS:


tadhg-ohiggins:
What is the purpose of tracking these events? Is it for logging? Is it so that the UI can retrieve the latest date and current status of submissions? I’m not seeing a need for performance optimization on the field, which makes me think we should use another `JSONField`, probably with `DjangoJSONEncoder` to handle dates and times, rather than `ArrayField`. The `submitted_by`, `date_created`, `submission_status`, and `report_id` fields are top-level standard fields because we anticipated that those would be the main fields we would be using to make queries about submissions, but that doesn’t seem to apply here. We could add a `JSONField` called `events`, or, if we want to be more generic, `meta_information`, make it an array, and store the status transitions events in that array as JSON objects, something like:

```json
{
    "type": "status_transition",
    "timestamp": "YYYY-MM-DDTHH:mm:ss.sssZ",
    "data": {
        "transition_from": "in_progress",
        "transition_to": "ready_for_certification",
        "by_user": user_id
    }
}
```

JeanMarie-TTS:
> What is the purpose of tracking these events? Is it for logging? Is it so that the UI can retrieve the latest date and current status of submissions? I’m not seeing a need for performance optimization on the field, which makes me think we should use another `JSONField`, probably with `DjangoJSONEncoder` to handle dates and times, rather than `ArrayField`. The `submitted_by`, `date_created`, `submission_status`, and `report_id` fields are top-level standard fields because we anticipated that those would be the main fields we would be using to make queries about submissions, but that doesn’t seem to apply here. We could add a `JSONField` called `events`, or, if we want to be more generic, `meta_information`, make it an array, and store the status transitions events in that array as JSON objects, something like:
> 
> ```json
> {
>     "type": "status_transition",
>     "timestamp": "YYYY-MM-DDTHH:mm:ss.sssZ",
>     "data": {
>         "transition_from": "in_progress",
>         "transition_to": "ready_for_certification",
>         "by_user": user_id
>     }
> }
> ```

We need these fields in the dissemination views.
The JSON option looks elegant. But are we over-engineering by doing this?
I don’t think django models support an array of JSON fileds. We could have an array inside the JSON field. But this may make the ETL process a little more complex.

JeanMarie-TTS:
Moving forward with the implementation using parallel arrays for now. It's working now and we can improve the implementation post-MVP.

jadudm (approved):
I will, but I'm uncomfortable with PRs that have no comments/description, implied discussion between developers, and no clear resolution.

PR #1317, opened 2023-06-22
Add quotation marks to self in Django settings.py for Content Security Policy

tadhg-ohiggins:
Use quotation marks
When referring to the self
Identity: weird

-----

Turns out you have to put quotation marks around `self`.


jadudm (approved):
This appears to be the correct way to quote 'self' in this context.

'self' *had to* be quoted to resolve correctly as part of the policy.

LGTM. Testing.

PR #1319, opened 2023-06-22
1281 additional ueis workbook

sambodeme:


backend/audit/test_excel.py
```diff
@@ -1,440 +1,508 @@
...
-        for field_name, value in test_cases:
-            with self.subTest():
-                _set_by_name(workbook, field_name, value)
-
-                findings = extract_findings_text(workbook)
-
-                self.assertRaises(
-                    ValidationError, validate_findings_text_json, findings
-                )
+import json
```
danswick, line 1: I've noticed a few diffs look like this. Any idea why it looks like the entire file was deleted and then recreated?

danswick:
@sambodeme I started reviewing this PR but will need to come back to it on Monday. So far, everything is looking good and I was able to open the generated workbook.

backend/audit/test_excel.py
tadhg-ohiggins: I assume it goes away if you hide the whitespace in the diff view? My guess is a line endings issue.

JeanMarie-TTS:
Is there a need to store a sequence number? Can the additional UEIs not be stored in an Arrayfield instead of a JSON field?

danswick:
> Is there a need to store a sequence number? Can the additional UEIs not be stored in an Arrayfield instead of a JSON field?

I don't believe the additional UEIs are associated with any further information, so we shouldn't need to store anything about their sequence.

sambodeme:
> > Is there a need to store a sequence number? Can the additional UEIs not be stored in an Arrayfield instead of a JSON field?
> 
> I don't believe the additional UEIs are associated with any further information, so we shouldn't need to store anything about their sequence.

While working on historical data dissemination stories, we realized the necessity of including a sequential number for rows in all the xlsx files we have created so far. Therefore, in the new xlsx files we are currently building, I took the extra precaution of including this sequential number to prevent encountering the same issue as with the existing xlsx file. However, if we are absolutely certain that it is unnecessary in this case, I will happily remove it. The effort required to remove it is significantly less than the effort required to add it later on.  Same answer to @JeanMarie-TTS comment.

JeanMarie-TTS:
@sambodeme Let's remove the sequence number.
What about using an Array in the database instead of JSON?

sambodeme:
> @sambodeme Let's remove the sequence number. What about using an Array in the database instead of JSON?

@JeanMarie-TTS That is doable. We can still do the validation in json format (as we're using JSON schema validation) and save as an array. Does array work better for you ?

JeanMarie-TTS:
@sambodeme Array is better. With JSON we end up defining an array inside the JSON, and I think this makes the structure more complex wothout adding more value.

sambodeme:
> @sambodeme Array is better. With JSON we end up defining an array inside the JSON, and I think this makes the structure more complex wothout adding more value.

As we want to save this as an Array (of UEI entries) we'll be losing track of the auditee_uei that in the header of the xlsx. Is it okay to remove auditee_uei from the xlsx header at this point? @jadudm

sambodeme:
@JeanMarie-TTS @danswick , I removed the sequence number [here](https://github.com/GSA-TTS/FAC/pull/1377) but I have not switched the format to array yet as I have an open question for @jadudm on the auditee_uei field.

jadudm:
I very much want us to consult with @timoballard before we make changes to the record row. He gave a great deal of thought to how/why we store the data the way we do.

For example: the workbooks all are stored under a key with their name (e.g. `FederalAwards: { ... }`). The reason is that they can all be composed into a single object, easily, if we desire for validation. This *might* still matter to us.

I do not want us to move away from storing a named/tagged JSON object in a casual way. The importance of consistency at this point is FAR MORE IMPORTANT than any notional optimization/inconsistency.

As long as we have separate documents, we *MUST* have the auditee UEI in the document. Auditors might have many auditees, and they could accidentally upload a document that belongs to another auditee. This would be undetectable without being able to cross-validate all of the documents and confirm they all have the same/correct UEI.

Again: we *MUST* think through all changes that take us away from established patterns/designs. This is doubly-true for our database/storage choices. @JeanMarie-TTS , @danswick , I would ask that you both help me be a guardian of consistency before any notional "improvements" or "optimizations." Or, at least discuss through and document the reasons that any such changes are worthwhile. We need to discuss the *reasons*, not "we could use an array vs. JSON." The user-centered, behavior-centered reasons are what determine if we're making a breaking choice or not. 

I'm concerned in this case that removing the structure around the array very much makes storing this document *significantly worse*, to the point that it would degrade our confidence in the integrity of our data. Verbosity is not a problem. Consistency is not a problem. Bad data is not acceptable. That needs to be our mantra.

sambodeme:
@jadudm  thanks. This makes sense. I've only made [updates](https://github.com/GSA-TTS/FAC/pull/1377) that remove the seq_number, leaving the JSON structure untouched and consistent with the existing xlsx JSON models. I'll proceed with merging this into the current branch. If we decide to opt for the Array approach later, I'll tackle this in a separate PR.

timoballard:
Just catching up here, but wanted to voice my agreement with keeping things consistent at this stage. Storing as JSON in this case may not add information that we wouldn't have with an array, strictly speaking, but right now, the ingestion/validation pipeline for all of the workbooks has the same shape. I think that has value in terms of reduced complexity of the system as a whole for the comparatively low price of having a slightly larger-than-necessary data structure

JeanMarie-TTS:
> 

Agreed. In addition as @sambodeme pointed out we do need to store the parent UEI without repeating it. So a JSON is fine.

sambodeme:
@jadudm This is ready now.

jadudm (approved):
Reviewed as part of group review, and updated to reflect changes in another branch re: merged cells. Bringing it in to `main`.

Issue #1280, opened 2023-06-17
Secondary auditors workbook


jadudm:
### Tasks

```[tasklist]
- [x] Convert secondary auditors form to Jsonnet
- [x] Add unit tests for secondary auditors form checking validation boundaries
- [x] Add migration to add secondary auditors column in 2023 data model
- [x] Add Django mapping to store secondary auditors form
- [x] Add end-to-end test for happy path on secondary auditors form
```

Broken out for visibility from #1219

danswick:
We're done here!

PR #1331, opened 2023-06-23
Secondary auditors workbook

danswick:
Addresses #1280


backend/schemas/source/base/FederalProgramNames.json
danswick, line 1: I don't think this file should have changed. It probably got caught up in a merge issue.

backend/schemas/output/excel/xlsx/additional-ueis-template.xlsx
danswick, line 1: I ran the schema `make` command after catching up to main and it updated all of the other workbooks, which may be expected? They only appear to have changed by a few bytes apiece.

danswick:
@jadudm am I missing a process for adding new excel templates to the repo? After some stumbling around, I'm running into some issues because my new excel template is gitignored.

backend/schemas/output/excel/xlsx/additional-ueis-template.xlsx
jadudm: Make always remakes all the files. We're currently keeping the workbooks in the tree, and this is gonig to be a side-effect of that. Normally, build products would not be in the tree.

When time permits, I'd be game to have a way to tag/release workbooks via actions. For now...

sambodeme (approved):

PR #1332, opened 2023-06-23
Fix submitter-only bug on audits table view

tadhg-ohiggins:
Let every role see
Their current submission list
Not just creators.

-----

It turns out that the **In-progress Audit Submissions** table at `/audit` was only showing users the submissions they themselves created, not all of the submissions they’re associated with. This should fix that.

We should probably also have tests for this.

### DIFF
--- a/backend/audit/views.py
+++ b/backend/audit/views.py
@@ -24,7 +24,7 @@ from audit.mixins import (
     CertifyingAuditorRequiredMixin,
     SingleAuditChecklistAccessRequiredMixin,
 )
-from audit.models import ExcelFile, SingleAuditChecklist
+from audit.models import Access, ExcelFile, SingleAuditChecklist
 from audit.validators import (
     validate_federal_award_json,
     validate_corrective_action_plan_json,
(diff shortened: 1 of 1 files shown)


tadhg-ohiggins:
Tim is out until the middle of next week, so I can't ask him to review this.

The best way to test this is probably as follows:

1.  Set up two separate login.gov sandbox accounts.
2.  Log in to https://fac-dev.app.cloud.gov/ with account 1 and create a new submission in which account 2’s email address is assigned to one of the roles.
3.  Copy the `report_id` of the submission.
3.  Log out of account 1 and log into account 2.
4.  Verify that the submission is not visible at https://fac-dev.app.cloud.gov/audit/ for account 2 (this absence is the bug).
4.  Verify that despite submission not showing up there, account 2 can nonetheless access it at https://fac-dev.app.cloud.gov/report_submission/general-information/{report_id} where `{report_id}` is the `report_id` from step 3. (If it’s not accessible, something else, possibly unrelated to this bug, is going on.)

Then check out and run this branch locally and repeat the above steps on localhost. This time the submission should be visible on `/audit/` for account two.

jperson1 (approved):
I was able to verify that this change works by:
1. Making a second sandbox login.gov account and starting a submission that referenced the first account. 
2. Notice that the first account could _not_ see the new submission when running on the main branch, but _could_ still access it when linking directly to it.
3. Notice that the first account _could_ see the new submission when running this branch.

PR #1342, opened 2023-06-27
Update staging pull request name

asteel-gsa:
Updating the PR title to the following format `YYYY-MM-DD | MAIN --> PROD | DEV (Short Sha) --> STAGING`


JeanMarie-TTS (dismissed):
I suggest we keep the existing description i.e. main to prod as these are the branches we are merging from and to. We may want to rename the branch `prod` at some point to avoid confusion with `production` which is an environment.

asteel-gsa:
@JeanMarie-TTS Please Re-review

mogul (approved):

PR #1344, opened 2023-06-27
Use Django static template tag for xlink:href

tadhg-ohiggins:
We want our sources
Static via dynamic.
That’s not confusing.

-----

Use [the static template tag](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#std-templatetag-static) for `xlink:href` to make sure assets load correctly in all environments, not just local.


jperson1 (approved):
Makes a lot of sense, thanks!

PR #1345, opened 2023-06-27
Jadudm/additional xlsx in book validations

jadudm:
This PR lays a foundation for further work on the workbooks and validation in the workbooks.

1. It improves `render.py`, breaking out `parse.py` to improve our confidence on the processing of the Jsonnet specs.
2. It adds the ability to add columns of text (based on enumerations in our JSON schema validations) to the workbooks, and then use those columns of text as part of drop-downs in the data entry.
3. Moves the build process to a top-level `Makefile`, so that all of the data gets transformed, all of the validations built, and all of the sheets built on those validations in order.

This may impact work that @sambodeme and @danswick are doing. Whether we want to merge into `main` and use this to help make sure we have everything building consistently, or do something else with the work can be part of the discussion on this PR.

Tagging @sambodeme and @danswick for review, as it impacts them most directly.


asteel-gsa:
added this, which seems to have resolved the linting issue on `backend/schemas/scripts/render.py | backend/schemas/source/excel/render.py`
```toml
[tool.mypy]
ignore_missing_imports = true
explicit_package_bases = true
```

sambodeme (approved):

neilmb:
(Squash-) Merging to attempt to unblock federal awards uploads.

PR #1353, opened 2023-06-27
Put the hash symbol outside the Django call so it doesn't get escaped

tadhg-ohiggins:
Escapes will get you
Maybe outside the brackets
You will be safer.

-----

Try putting `#` outside the Django `static` invocation.


neilmb (approved):
LGTM. The only way to test if this does what we want is to deploy it to `dev` because it appears that it works differently on local. So let's see if it does the right thing!!

PR #1360, opened 2023-06-28
Make Excel file saving set unique filename

tadhg-ohiggins:
Set our own filenames.
Even in the filesystem.
We mean it this time.

-----

Actually set the filesystem name for the file, not just the attribute in the database model.

### DIFF
--- a/backend/audit/models.py
+++ b/backend/audit/models.py
@@ -475,12 +475,22 @@ class Access(models.Model):
         ]
 
 
+def excel_file_path(instance, filename):
+    """
+    We want the actual filename in the filesystem to be unique and determined
+    by report_id and form_section--not the user-provided filename.
+    """
+    base_path = "excel"
+    report_id = SingleAuditChecklist.objects.get(id=instance.sac.id).report_id
+    return f"{base_path}/{report_id}--{instance.form_section}.xlsx"
+
+
 class ExcelFile(models.Model):
     """
     Data model to track uploaded Excel files and associate them with SingleAuditChecklists
     """
 
-    file = models.FileField(upload_to="excel", validators=[validate_excel_file])
+    file = models.FileField(upload_to=excel_file_path, validators=[validate_excel_file])
     filename = models.CharField(max_length=255)
     form_section = models.CharField(max_length=255)
     sac = models.ForeignKey(SingleAuditChecklist, on_delete=models.CASCADE)
--- a/backend/audit/models.py
+++ b/backend/audit/models.py
@@ -475,14 +475,12 @@ class Access(models.Model):
         ]
 
 
-def excel_file_path(instance, filename):
+def excel_file_path(instance, _filename):
     """
     We want the actual filename in the filesystem to be unique and determined
     by report_id and form_section--not the user-provided filename.
     """
-    base_path = "excel"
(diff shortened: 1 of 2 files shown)


timoballard:
Will the overall effect here be that we overwrite the Excel document in S3 (so there is only one per form section per report id), but we still accumulate `ExcelFile` database entries for each upload (all of which point to the same, most recently uploaded document due to the name overlap)?

If that's accurate, wondering if we should also only have a single database entry per form section per report id?

backend/audit/models.py
```diff
@@ -475,12 +475,22 @@ class Meta:
         ]
 
 
+def excel_file_path(instance, filename):
+    """
+    We want the actual filename in the filesystem to be unique and determined
+    by report_id and form_section--not the user-provided filename.
+    """
+    base_path = "excel"
+    report_id = SingleAuditChecklist.objects.get(id=instance.sac.id).report_id
```
neilmb, line 484: **question (non-blocking)**: Isn't this just `instance.sac.report_id`? If there is an `instance.sac` object to look up `id` on, then won't it also have a `report_id` attribute that we can use directly?

backend/audit/models.py
```diff
@@ -475,12 +475,22 @@ class Meta:
         ]
 
 
+def excel_file_path(instance, filename):
```
neilmb, line 478: **question (non-blocking)**: Since we don't use the `filename` argument at all, is it worth calling it `_filename`? I think likely not, but given a recent PR, I'm curious.

neilmb (approved):
LGTM, again quibbles.

backend/audit/models.py
```diff
@@ -475,12 +475,22 @@ class Meta:
...
+    """
+    We want the actual filename in the filesystem to be unique and determined
+    by report_id and form_section--not the user-provided filename.
+    """
+    base_path = "excel"
+    report_id = SingleAuditChecklist.objects.get(id=instance.sac.id).report_id
+    return f"{base_path}/{report_id}--{instance.form_section}.xlsx"
```
timoballard, line 485: Technically we [allow both `xls` and `xlsx`](https://github.com/GSA-TTS/FAC/blob/6b9cf8b34c558ba2209af6c6aa4635c0a1337d2a/backend/audit/validators.py#L31) extensions here, so we should probably extract and reuse the extension from the user-provided filename (at this point it will have already been validated to be either `xls` or `xlsx`)

timoballard: Alternatively, we should remove `xls` as an allowed extension. Actually starting to lean in this direction, considering the uploaded Excel files originate from us as `xlsx`

backend/audit/models.py
tadhg-ohiggins: It definitely should be `_filename`

backend/audit/models.py
tadhg-ohiggins: No, no, it's better to throw in a redundant query just to make sure the oil flows smoothly in the database.

backend/audit/models.py
tadhg-ohiggins: I'm going to remove `.xls` from the list of allowed extensions.

timoballard (approved):

PR #1361, opened 2023-06-28
Serve the USWDS SVG sprite from Django to get around the cross-domain issues

tadhg-ohiggins:
Sprites sound so friendly
But in the wild, they’re tricky.
How appropriate.

-----

1.  ~~Copy `sprite.svg` into the Django template includes directory at deploy time locally and in cloud.gov.~~
2.  ~~Include the contents of `sprite.svg` directly into our HTML pages.~~
3.  ~~Create a new custom Django tag to use so that if we change this in future we can hopefully just do that in one place.~~
4.  ~~Replace all the current instances of the sprite in our template with the new tag.~~

-----

1.  Create custom view to serve `sprite.svg`.
2.  Create a new custom Django tag to use so that if we change this in future we can hopefully just do that in one place.
3.  Replace all the current instances of the sprite in our template with the new tag.

Hopefully this actually does work in our cloud.gov environments!

The rationale for why we ended up trying this approach follows:

USWDS uses a specific pattern with its icons to load a single SVG file as a sprite and then refer to fragments of it for the actual images for specific icons:

```html
<svg class="usa-icon" aria-hidden="true" focusable="false" role="img">
    <use xlink:href="path-to-svg-file#logout"></use>
</svg>
```

The SVG file in question is with all of our other static assets, so that it’s in the container for local development but in an S3 bucket in our cloud.gov environments. This means that the SVG is on a different domain from our app, and that means that browsers may (and do) block loading of the file to prevent cross-domain exploits.

In cloud.gov, we don’t serve any assets from our instance, only from the S3 bucket, and adding the ability to do so is tricky.

Similarly, adding the machinery to serve assets from our domain that are actually coming from S3, transparently to browsers, is tricky.

~~jThe SVG can, however, be included as part of the HTML document itself, at which point its fragments—assuming there are no name collisions—become available for use with `use` and, since it’s just part of the current document, no cross-domain implications arise.~~

~~Disadvantages:~~

1.  ~~The SVG file is 70K in size and by inserting it into every document we’re basically ensuring that it’s uncacheable, and most of the size seems to come from icons we don’t actually use.~~
2.  ~~We have to copy the file around manually.~~

Possible future enhancement approaches:

1.  Serve assets from the application instance and serve it that way.
2.  Serve assets from a specific path from the S3 bucket so that the browser doesn’t think they’re coming from different domains.
3.  Either compile just the icons we need into the SVG file or winnow the file down to just the icons we need.
4.  ~~Use JavaScript to grab the file, since JS should be cross-domain enabled in this scenario, and use it to insert the entire thing onto the page, as now at least the file will be cached.~~

### DIFF
--- a/backend/.profile
+++ b/backend/.profile
@@ -41,4 +41,6 @@ python manage.py migrate &&
 echo 'Finished migrate' &&
 echo 'Starting collectstatic' &&
 python manage.py collectstatic --noinput &&
-echo 'Finished collectstatic'
+echo 'Finished collectstatic' &&
+# This is part of inlining this sprite due to cross-domain issues with the svg use element; see docs/deploying.md#questions
+cp static/img/sprite.svg templates/includes/
--- a/backend/.profile
+++ b/backend/.profile
@@ -42,5 +42,3 @@ echo 'Finished migrate' &&
 echo 'Starting collectstatic' &&
 python manage.py collectstatic --noinput &&
 echo 'Finished collectstatic' &&
-# This is part of inlining this sprite due to cross-domain issues with the svg use element; see docs/deploying.md#questions
-cp static/img/sprite.svg templates/includes/
--- a/backend/api/views.py
+++ b/backend/api/views.py
@@ -1,20 +1,19 @@
 import json
 from typing import List
 
-from audit.models import Access, SingleAuditChecklist
-from audit.permissions import SingleAuditChecklistPermission
-from django.http import Http404, HttpResponse
+from django.http import Http404, HttpResponse, JsonResponse
 from django.urls import reverse
-from django.views import View
-from config.settings import AUDIT_SCHEMA_DIR, BASE_DIR
+from django.views import View, generic
+from django.core.exceptions import ValidationError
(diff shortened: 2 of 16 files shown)


neilmb:
As a half-measure towards addressing some of the drawbacks of this, instead of setting our whole Django instance up to serve static files at some path like `/static/` then we could just make a single view method on the URL `/static/sprite` that returns the SVG file. If we set the right correct cache-control headers in our method response, then it should be cache-able and would not inflate page sizes.

Just another idea towards the same purpose, sort of a manual staticfiles module.

danswick (approved):
This approach looks good to me so far. Maybe consider if we want to live with a `hacky_sprite` tag, but I defer to you 😄

heymatthenry (approved):
LGTM `hacky_sprite` ftw

neilmb (approved):
LGTM, nothing but quibbles.

tadhg-ohiggins:
> LGTM `hacky_sprite` ftw

@heymatthenry I'm following Neil's suggestion to change it to `sprite_helper` 😿

PR #1363, opened 2023-06-29
Jadudm/scrub census data

jadudm:
This code allows us to take a (current) historical dump of Census data and

1. apply fixes to the data that is broken, 
2. `\copy` it into a local Postgres DB (in a running FAC stack)
3. `pg_dump` it back out as plaintext (`.sql`) and binary (`.dump`) files.

Census is still working on getting their export process to produce clean/reliable data for us. This is a stop-gap to allow us to proceed.

The code wants to merge as a reference and starting point. Next steps include:

1. Validating the resulting tables (e.g. do the tables have the same number of rows and columns as the CSVs?)
2. Considering how to better integrate/leverage the tables in our stack
3. ...

I would rather not lose the code by having it on my local machine, but it doesn't (yet) live up to a production standard. However, we do need the ability for multiple engineers to work with this data/build on these ideas.

Either we merge to `main`, or leave it in a branch.


asteel-gsa:
used `black -v .` and `black --diff .` to find suggested changes and publish for review

asteel-gsa:
added this to suppress [this](https://docs.openstack.org/bandit/1.4.0/plugins/try_except_pass.html). Should

```toml
[tool.bandit]
skips = ["B110"]
exclude_dirs = [
    "audit/test_schemas.py",
    "audit/test_validators.py",
    "node_modules",
]
```

Current errors on mypy
```
data_distro/census_raw/strip-repeated-headers.py:34: error: Argument 1 to "write" of "TextIOBase" has incompatible type "str | None"; expected "str"  [arg-type]
data_distro/census_raw/strip-repeated-headers.py:36: error: Item "None" of "str | None" has no attribute "split"  [union-attr]
data_distro/census_raw/strip-repeated-headers.py:36: error: Incompatible types in assignment (expression has type "list[str] | Any", variable has type "str | None")  [assignment]
data_distro/census_raw/strip-repeated-headers.py:45: error: Value of type "str | None" is not indexable  [index]
```

PR #1334, opened 2023-06-26
Build and Publish Web Container

asteel-gsa:
```mermaid
graph TD;
  A[Schedule 6a EST M-F]--> B[Trivy Scan];
  B-->C[Publish To Security Tab];
  A-->D[Build Web Container];
  D-->E[Publish Container to ghcr];
  A-->F[Lint Code];
  F-->G[Pull Container from ghcr];
  G-->H[Run Django Tests];
  H-->I[Run Lighthouse Tests];
  I-->J[Run a11y Tests];
  A-->K[Create Pull Request to Staging];
  K-->L[No Action - Exit 1];
  K-->M[Pull Request Created];
  M-->N[Ready for Review];
```

### Note:
- Using `docker-compose-web.yml` for now, just to ensure that there are two `docker-compose` files. One for locally building the image if one chooses, and the other to pull the image from `ghcr.io/gsa-tts/fac/web-container:latest`
- Current Trivy Scan in `scan-images.yml` only checks for `postgrest` and `clamav` running on a weekly schedule, adding `trivy.yml` to scan the web container before building and publishing to `fac ghcr` as this will run daily (weekdays only). 
- `fac/web-container` will not be version controlled by `YYYYMMDD` and opts for `:latest` only.

### Required inputs:
- Trivy
    -  work-dir: ./backend
    - docker-name: fac
- Build Container
    - docker-name: fac
    - image-name: web-container
    - repo-name: gsa-tts/fac
    - work-dir: ./backend

PR #1365, opened 2023-06-29
Add conditional testing for modifying requirements

asteel-gsa:
Test and lint with build: [GHA run where one of the 4 files were changed](https://github.com/GSA-TTS/FAC/actions/runs/5417025341/jobs/9847512683)
Test and lint with GHCR: [GHA run where no files changed](https://github.com/GSA-TTS/FAC/actions/runs/5417088891)

Basically, if the following items are modified on the PR, we need to ensure that we build the container (but not publish) to allow the tests to run and complete with the new dependencies. Otherwise, we want to use the latest from ghcr as rebuilding isnt necessary.
```
- './backend/requirements.txt'
- './backend/dev-requirements.txt'
- './backend/Dockerfile'
- './backend/package.json'
```
Then, once the merge to main occurs, build and publish from `main` and run workflows as necessary.


asteel-gsa:
Just as a note for when we merge this into main.. lets definitely `squash merge` this, because this history shouldn't be in main

mogul (dismissed):
Really impressive stuff! You're really getting the hang of the GHA YAML.

To be thorough, let's 

1. `s/docker-compose/docker compose/` in `docs/testing.md`
2. same in `docs/development.md`
3. same in `.github/workflows/end-to-end-test.yml`

I'm presuming you're going to add the `DOCKER_CONTENT_TRUST=1` env var in a separate PR; is that right?

.github/workflows/test-and-lint-docker-build.yml
```diff
@@ -0,0 +1,136 @@
+---
+name: Runs linting and tests
+on:
+  workflow_call:
+
+jobs:
+  linting:
```
neilmb, line 7: **question (non-blocking)**: I don't know enough about the granularity available in Github actions jobs, but it looks to me like there is no docker differences in the `linting` job which happens inside the Github runner rather than in our docker image. Is it worth splitting that job out so that it doesn't have to be repeated between these two invocations?

.github/workflows/triggers-dev.yml
```diff
@@ -5,9 +5,23 @@ on:
...
+      docker-name: fac
+      image-name: web-container
+      repo-name: gsa-tts/fac
+      work-dir: ./backend
+
   # testing
   test-and-lint:
-    uses: ./.github/workflows/test.yml
+    needs: [build-container]
+    uses: ./.github/workflows/test-and-lint-from-ghcr.yml
```
neilmb, line 24: **question (non-blocking)**: Do we know at this point that the new GHCR image from the PR to main has been built and pushed to GHCR and will be the one pulled by this test run? It'd be a shame to fail a test run at this point because the new dependencies weren't quite there in GHCR yet.

neilmb (approved):
LGTM, as we discussed it. Just a couple questions from my own ignorance.

asteel-gsa:
> Really impressive stuff! You're really getting the hang of the GHA YAML.
> 
> To be thorough, let's
> 
> 1. `s/docker-compose/docker compose/` in `docs/testing.md`
> 2. same in `docs/development.md`
> 3. same in `.github/workflows/end-to-end-test.yml`
> 
> I'm presuming you're going to add the `DOCKER_CONTENT_TRUST=1` env var in a separate PR; is that right?

Thank you for reminding me about that, will address tomorrow morning

.github/workflows/test-and-lint-docker-build.yml
asteel-gsa: While they are added in two places, they only execute once regardless. The conditional logic [here](https://github.com/GSA-TTS/FAC/blob/64ea9e3ef2858d53add14ac116db025dbfafb75d/.github/workflows/triggers-PR.yml#L47) and [here](https://github.com/GSA-TTS/FAC/blob/64ea9e3ef2858d53add14ac116db025dbfafb75d/.github/workflows/triggers-PR.yml#L41) determine which `test.yml` should be run and lint/test accordingly.

```mermaid
graph TD;
  A[Any of the 4 files changes]-->B[Lint]
  A-->C[Build container and run django/a11y]
```
```mermaid
graph TD;
  A[No changes to dependencies files]-->B[Lint]
  A-->C[Pull container from gchr and run django/a11y]
```

.github/workflows/triggers-dev.yml
asteel-gsa: `triggers-dev.yml` has a [push hook on main](https://github.com/GSA-TTS/FAC/blob/64ea9e3ef2858d53add14ac116db025dbfafb75d/.github/workflows/triggers-dev.yml#L2-L5C13). So every time we merge, before it tests and lints (regardless of whether there were changes to the 4 files or not) we will [build the image and publish to ghcr](https://github.com/GSA-TTS/FAC/blob/64ea9e3ef2858d53add14ac116db025dbfafb75d/.github/workflows/triggers-dev.yml#L9-L19) by invocation of [build-docker-container.yml](https://github.com/GSA-TTS/FAC/blob/64ea9e3ef2858d53add14ac116db025dbfafb75d/.github/workflows/build-docker-container.yml)

So on merge to `main` we will publish a new `web-container:latest` and then test with the ghcr image.

At 6a EST M-F, as part of the nightly deploy to staging, it will trivy scan, build and publish, test with gchr image, and create the PR for review [staging-scheduled-deploy.yml](https://github.com/GSA-TTS/FAC/blob/64ea9e3ef2858d53add14ac116db025dbfafb75d/.github/workflows/staging-scheduled-deploy.yml). More info/diagram on #1334

mogul:
I think you have to do more on the push side with secrets or somesuch.

mogul:
> ```yaml
>       - name: Push Web Container
>         run: |
>           export DOCKER_CONTENT_TRUST=1 && docker push --all-tags ghcr.io/${{ env.GH_REPO }}/${{ env.IMAGE }}
> ```

There's need for the `export` and the `&&` since the var is just for the invocation of that command. You can just do:

```bash
DOCKER_CONTENT_TRUST=1 docker push --all-tags ghcr.io/${{ env.GH_REPO }}/${{ env.IMAGE }}
```

For this one:
> ```yaml
>       - name: Start services
>         working-directory: ./backend
>         run: docker compose -f docker-compose.yml up -d && export DOCKER_CONTENT_TRUST=1
> ```

You want the env var set before you run `docker compose`, so similar to the Makefile example you'd do:

```bash
run: DOCKER_CONTENT_TRUST=1 docker compose -f docker-compose.yml up -d
```

Issue #1352, opened 2023-06-27
Manage S3 CORS headers in Terraform and add them for production


tadhg-ohiggins:
### Background

We manually added the required CORS headers to the S3 buckets for fac-dev.app.cloud.gov and fac-staging.app.cloud.gov but should manage this with Terraform.

WE DID NOT DO THIS FOR THE PRODUCTION ENVIRONMENT. So that still needs to be set up, preferably by the Terraform process that will cover all of the environments.

We used this command to set the CORS headers:
```sh
aws s3api put-bucket-cors --bucket [BUCKET_NAME] --cors-configuration file://cors.json
```

where `[BUCKET_NAME]` is the name of the bucket for that environment. The contents of `cors.json` were:

```json
{
   "CORSRules": [
      {
         "AllowedHeaders": [
            "Authorization"
         ],
         "AllowedMethods": [
            "HEAD",
            "GET"
         ],
         "AllowedOrigins": [
            "https://[ENV_DOMAIN]"
         ],
         "ExposeHeaders": [
            "ETag"
         ]
      }
   ]
}
```

where [ENV_DOMAIN] was `fac-dev.app.cloud.gov` for `dev` and `fac-staging.app.cloud.gov` for `staging`.

For reference, the commands for this are included below:

```sh
SERVICE_INSTANCE_NAME=fac-public-s3;
KEY_NAME=fac-public-s3-key;
cf create-service-key "${SERVICE_INSTANCE_NAME}" "${KEY_NAME}";
S3_CREDENTIALS=`cf service-key "${SERVICE_INSTANCE_NAME}" "${KEY_NAME}" | tail -n +2`;
export AWS_ACCESS_KEY_ID=`echo "${S3_CREDENTIALS}" | jq -r .access_key_id`;
export AWS_SECRET_ACCESS_KEY=`echo "${S3_CREDENTIALS}" | jq -r .secret_access_key`;
export BUCKET_NAME=`echo "${S3_CREDENTIALS}" | jq -r .bucket`;
export AWS_DEFAULT_REGION=`echo "${S3_CREDENTIALS}" | jq -r '.region'`;
echo $BUCKET_NAME;
aws s3api put-bucket-cors --bucket $BUCKET_NAME --cors-configuration file://cors.json;
aws s3api get-bucket-cors --bucket $BUCKET_NAME;
cf delete-service-key "${SERVICE_INSTANCE_NAME}" "${KEY_NAME}";
```

[...]

PR #1374, opened 2023-06-30
Add private S3 bucket to manifest.

tadhg-ohiggins:
[Bret said](https://github.com/GSA-TTS/FAC/issues/1352#issuecomment-1613918754) to add this
To manifest services.
Thus I make it so.

-----

Add `fac-private-s3` to the list of services in the cloud.gov manifest so that we don’t need to remember to do it manually for each environment.


neilmb (approved):
Looks quite reasonable.

PR #1384, opened 2023-07-05
Initial ETL code

pluribus-tony:
This is the first PR for the ETL code.  The intention is only to get at least a skeleton of the code needed to populate our General table.  We still have gaps in knowledge for the source of a lot of the data that will be coming in.  We also will need to start working on code for all the other (much smaller) tables once we have a better idea of how General will work.


backend/Makefile
```diff
@@ -82,4 +83,4 @@ docker-clean:
 
 docker-lint:
 	docker compose build
-	docker-compose run web bash -c 'flake8 && black --check . && bandit -c pyproject.toml -r . && mypy . && djlint .'
\ No newline at end of file
+	docker-compose run web bash -c 'flake8 && black --check . && bandit -c pyproject.toml -r . && mypy . && djlint .'
```
jadudm, line 86: I think/hope this was picked up elsewhere... in fact, I'm confident it will be on merge?

`docker-compose` should become `docker compose` everywhere. If a merge from `main` doesn't pick this up, then we'll go looking for whether it is lingering in a branch...

asteel-gsa: This branch is [46 commits ahead of](https://github.com/GSA-TTS/FAC/compare/main...tt/1146), [111 commits behind](https://github.com/GSA-TTS/FAC/compare/tt/1146...main) main.

```
git checkout main
git pull
git checkout tt/1146
git merge main
```

backend/audit/etl.py
```diff
@@ -0,0 +1,312 @@
...
+            )
+            state_cluster_name = (
+                cluster.get("state_cluster_name")
+            )
+            other_cluster_name = (
+                cluster.get("other_cluster_name")
+            )
+            federal_award = FederalAward(
+                award_id=entry.get("seq_number", 1),
+                federal_program_name=program["program_name"],
```
jadudm, line 116: I think these are getting picked up now from an enumeration that is burried in the Jsonnet validation code. 

https://github.com/GSA-TTS/FAC/blob/main/backend/schemas/source/base/FederalProgramNames.json

specifically. That list will be considered authoritative going forward. The cluster names also get worked into the validation schemas.

Don't know if that helps.

backend/dissemination/api/db_views.sql
```diff
@@ -0,0 +1,67 @@
+
+begin;
+
+drop view if exists api.vw_general;
```
jadudm, line 4: It doesn't have to be in this PR, but I _think_ we want to get rid of the `vw` in our views.

That is, the end-user of the API doesn't care that this is a view. Are we able to do that, or do we end up with name conflicts?

Actually, ignore that comment. We're probably going to have to do a small refactor on this anyway, as we may need to use a different schema. That is, we need to be able to version the API, and we do that by having multiple schema with multiple APIs live at the same time. In this regard, we may want an api_v1_0_0 schema. 

So, for now, ignore this comment, but it is a "thought placeholder" for the future. I'll get the API versioning doc into an ADR format for consideration, and that will help with documenting/ticketing how we need to handle the API schemas.

(It will be a small refactor, I think, hence it doesn't *need* to be picked up right now.)

jadudm (approved):
Team reviewed, as well as my read and local testing. We're merging this with light testing because we know that 1) that is still needed, but also 2) that moving it into `main` unblocks the ability of the team to work in parallel on this and related work around the dissemination model. Good stuff.

PR #1385, opened 2023-07-05
Dynamic SAC top-level property generation

tadhg-ohiggins:
It might not be Lisp
But it’s a bit magical
So, what could go wrong?

-----

Dynamically generate the General Information properties we want at the top level of the `SingleAuditChecklist` object.


backend/audit/models.py
```diff
@@ -50,7 +52,44 @@ def create(self, **obj_data):
...
+                return getattr(self, toplevelproperty)[key]
+            except KeyError:
+                pass
+            except TypeError:
+                pass
+            return None
+
+        return inner
+
+    schemadir = settings.BASE_DIR / "schemas" / "output" / "sections"
```
timoballard, line 82: should we use `settings.SECTION_SCHEMA_DIR` here?

backend/audit/models.py
```diff
@@ -50,7 +52,44 @@ def create(self, **obj_data):
...
+    filename = fname or f"{name}.schema.json"
+    toplevelproperty = toplevel or camel_to_snake(name)
+    mixinname = classname or f"{name}Mixin"
+
+    def _wrapper(key):
+        def inner(self):
+            try:
+                return getattr(self, toplevelproperty)[key]
+            except KeyError:
+                pass
```
timoballard, line 75: do you think it'd be worth adding some logging in these exception handlers? This looks like a place where problems could be tricky (or at least tedious) to pin down

jadudm:
I haven't had a chance to ticket it, but we do need to do a logging pass, and make sure it is making its way to New Relic, so that we can make sense of things that go wrong.

That may be getting picked up by the cloud.gov logging transfer to NR, but I have not had a chance to check. 

So: yes. We need to know when things go wrong, I think.

timoballard (approved):

PR #1386, opened 2023-07-06
1383 fac bug copying and pasting into excel worksheet with merged cells causes breakage

sambodeme:


backend/schemas/output/excel/json/federal-awards-expended-template.json
```diff
@@ -342,36 +323,41 @@
...
                   "formula1": 12,
                   "operator": "equal",
                   "type": "textLength"
-               }
+               },
+               "width": 36
             },
             {
-               "formula": "=SUM(FIRSTCELLREF:LASTCELLREF)",
+               "formula": "=SUM('Form'!FIRSTCELLREF:LASTCELLREF)",
```
jadudm, line 352: Also a ticket for a future refactoring:

Is there a way we can make the name of the sheet (e.g. `Form`) a variable in the Jsonnet? (Perhaps it is.) If it isn't, we should pull those out, so it is easy to change the names of the sheets in the forms, and not break the whole doc. (But, it will break the Python... so, this is a fragile thing. Something to think about. Perhaps we can ticket that we want to think about it... the specs and schemas are becoming strong "source of truth" docs for the full stack, so thinking about how we leverage the form names might be part of that...)

backend/schemas/output/excel/json/federal-awards-expended-template.json
```diff
@@ -30,7 +11,7 @@
                },
                "range_name": "federal_agency_prefix",
                "title": "Federal Agency Prefix",
-               "title_cell": "A3",
+               "title_cell": "A1",
```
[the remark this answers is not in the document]
sambodeme: Will ticket this.

backend/schemas/output/excel/json/federal-awards-expended-template.json
sambodeme: Hmm, I meant to change this but forgot. Nevertheless, I view this as essential for code maintainability, so I've now gone ahead and updated the code.

jadudm (approved):
This fixes known issues with merged cells and copy-paste, passes tests. Validates our named range approach to XLSX workbook construction.

PR #1393, opened 2023-07-06
Disable schedule for auto-deploy

asteel-gsa:
Disable Auto Deploy for 10 days for pentesting


jadudm (approved):
Necessary for ATO. Must close our pushes to staging and prod during pen testing.

PR #1400, opened 2023-07-07
Update Trivy

asteel-gsa:
Changes:
- Rename `scan-images.yml` to `pull-containers-and-push-to-ghcr.yml` to better reflect the action as we already have a `trivy.yml`
- Removed code from weekly scans of third party containers and condensed it to always run every sunday at `1230a est` regardless of version
- Added matrix scan of `fac ghcr` published containers as part of trivy scans in `trivy.yml`
- Forced env vars to be local rather than pass-through
- Added Trivy hook for `branch = prod` and `tag - v1.*`
    - This should now allow us to filter on branches for vulns across all envs.

### Third Party Scanning Weekly
```mermaid
graph TD;
    A[Pull 3rd Party Containers]-->B[Scan];
    B-->C[Detect Vuln from Upstream];
    C-->D[Do not publish to fac ghcr];
    B-->E[Publish to fac ghcr];
```

### FAC Published Containers Scanning on scheduled deploy to `staging from main before PR`, `push to prod` and `tag creation of v1.*`
```mermaid
graph TD;
    A[Pull Published Fac Containers]-->B[Scan];
    C[Build Fac Dockerfile]-->D[Scan];
    B-->E[Publish findings to security tab];
    D-->E;
```


jadudm (approved):
Per conversation, there will likely be ongoing renaming/refactoring to reflect what things do vs. what they were named when started. Good call on the rename.

Need to remove auto actions (for now) because of LATO process. Thank you.

LGTM.

PR #1457, opened 2023-07-10
Ignore changes to ASGs outside Terraform

mogul:
Until the Terraform provider enables fine-grained manipulation of a space's ASGs by an OrgManager, we have to do it manually. Ignoring changes here using the `lifecycle` attribute will prevent Terraform from trying to undo changes made manually (eg removal of the public egress ASGs on the non-egress spaces).


asteel-gsa (approved):
Talked through on call. Approved.

tadhg-ohiggins (approved):

PR #1476, opened 2023-07-11
Update Federal Award model and ETL, add test for Federal Award.

pluribus-tony:


JeanMarie-TTS (changes requested):
Why are making changes to package-lock.json?

backend/dissemination/models.py
```diff
@@ -266,16 +242,10 @@ class FederalAward(models.Model):
     #     help_text=docs.questioned_costs_FederalAward,
     # )
 
-    # SK - Note: Need to add the following:
-    # audit_report_type
-    # elecauditsid
-    # federal_agency_prefix
+    # TODO: Do we need to add any of the following?:
```
jadudm, line 245: If the question is "will we need is_guaranteed, is_passed, or subrecipient_amount?," the answer will be yes.

jadudm (approved):
Good incremental add, pushing through... although, I'll wait to see why there were packaage-lock.json changes. Perhaps back those out?

JeanMarie-TTS (approved):

PR #1482, opened 2023-07-12
Moving the new API into v1.0.0-beta

jadudm:
This creates a path to multiple APIs active at once, and versioning them.


JeanMarie-TTS (changes requested):
There are a couple of migration files that I think should be removed from the PR. There was a PR implemented that already handles this particular migration.

PR #1487, opened 2023-07-12
1416 auditor table has etl from collection to dissemination

JeanMarie-TTS:


tadhg-ohiggins:
The changes to add Additional/Secondary Auditors are being tracked by https://github.com/GSA-TTS/FAC/issues/1280 .

In addition, I don't think we should be trying to abstract out auditors via something like `GenAuditor` as it raises data-cleaning issues we're not addressing for MVP.

JeanMarie-TTS:
@tadhg-ohiggins GenAuditor exists only in the dissemination schema and associated views.  It should have no bearing on what we do in the intake and the validations. Please explain if you still have a concern.
Conceptually, GenAuditor is similar to FederalAward, Finding, etc in that a JSON object (in the audit model ) which iincludes an array of dictionaries is being pulled out into its own table for dissemination.

JeanMarie-TTS:
Functionality implemented in other PRs

PR #1488, opened 2023-07-12
Cross validation

timoballard:


backend/audit/cross_validation/additional_ueis.py
```diff
@@ -0,0 +1,49 @@
...
+    if addl_ueis_checked:
+        """
+        We need to check that:
+
+        +   all_sections.additional_ueis isn't an empty list
+        +   The first object in it has a value for additional_uei
+        +   None of the values are the same as auditee_uei
+        """
+        if not addl_ueis:
+            msg = "".join(
```
neilmb, line 20: **question (non-blocking)**: This is a reasonable solution for formatting long strings, but I wonder how it differs from the possibly more idiomatic

```
("general_information.multiple_ueis_covered is checked, "
"but no additional UEIs were found.")
```

which accomplishes the same thing without the explicit `"".join(`?

backend/audit/cross_validation/additional_ueis.py
```diff
@@ -0,0 +1,49 @@
...
+                    "general_information.multiple_ueis_covered is checked, ",
+                    "but no additonal UEIs were found.",
+                ]
+            )
+            return [{"error": msg}]
+        addl_uei_list = [_["additional_uei"] for _ in addl_ueis]
+        if auditee_uei in addl_uei_list:
+            msg = "The additional UEIs list includes the auditee UEI."
+            return [{"error": msg}]
+    else:
```
neilmb, line 39: **comment (non-blocking)**: This is a long way away from the initial `if`, is it reasonable to include a comment about what it goes with?
```suggestion
    else:  # addl_ueis_checked is false
```

backend/audit/cross_validation/test_auditee_ueis_match.py
```diff
@@ -0,0 +1,56 @@
...
+from audit.models import SingleAuditChecklist
+
+from .auditee_ueis_match import auditee_ueis_match
+from .sac_validation_shape import sac_validation_shape
+
+from model_bakery import baker
+
+
+class AuditeeUEIsMatchTests(TestCase):
+    def test_all_sections_none(self):
```
neilmb, line 12: **comment**: These test names are reasonable, but I think that they could still use docstrings that explain the different situations they are trying to test, e.g.
```suggestion
    def test_all_sections_none(self):
         """For a SAC with all sections empty, the auditee UEIs match"""
```

backend/audit/cross_validation/test_auditee_ueis_match.py
```diff
@@ -0,0 +1,56 @@
...
+        sac = baker.make(SingleAuditChecklist)
+
+        shaped_sac = sac_validation_shape(sac)
+
+        validation_result = auditee_ueis_match(shaped_sac)
+
+        self.assertEqual(validation_result, [])
+
+    def test_single_auditee_uei(self):
```
neilmb, line 21: **comment**: docstring, as above?

backend/audit/cross_validation/test_auditee_ueis_match.py
```diff
@@ -0,0 +1,56 @@
...
+        sac.general_information = {"auditee_uei": "123456789"}
+
+        shaped_sac = sac_validation_shape(sac)
+
+        validation_result = auditee_ueis_match(shaped_sac)
+
+        self.assertEqual(validation_result, [])
+
+    def test_multiple_matching_auditee_ueis(self):
+        sac = baker.make(SingleAuditChecklist)
```
neilmb, line 33: **comment**: Docstring, as above?

backend/audit/cross_validation/test_auditee_ueis_match.py
```diff
@@ -0,0 +1,56 @@
...
+        sac.federal_awards = {"auditee_uei": "123456789"}
+
+        shaped_sac = sac_validation_shape(sac)
+
+        validation_result = auditee_ueis_match(shaped_sac)
+
+        self.assertEqual(validation_result, [])
+
+    def test_multiple_mismatched_auditee_ueis(self):
+        sac = baker.make(SingleAuditChecklist)
```
neilmb, line 45: **comment**: Docstring, as above?

backend/audit/models.py
```diff
@@ -206,22 +208,36 @@ class STATUS:
...
-            self.findings_text,
-            self.findings_uniform_guidance,
-            self.additional_ueis,
-            self.notes_to_sefa,
-        ]
-        if all(section for section in all_sections):
-            return True
-        return True
+        shaped_sac = audit.cross_validation.sac_validation_shape(self)
+        all_functions = audit.cross_validation.functions
```
neilmb, line 236: **comment (non-blocking)**: In context, the general term `functions` doesn't seem like it describes what this list is very well. Perhaps `all_validations` or `validation_functions`?

backend/audit/models.py
```diff
@@ -206,22 +208,36 @@ class STATUS:
...
+        errors = list(chain.from_iterable([func(shaped_sac) for func in all_functions]))
+        if errors:
+            return {"errors": errors, "data": shaped_sac}
+        return []
```
neilmb, line 240: **question**: This seems like a strange type mismatch. If the non-empty return value is a dict, should the empty return be an empty dict?
```suggestion
        return {}
```

backend/audit/models.py
```diff
@@ -294,11 +310,12 @@ def transition_to_submitted(self):
         the appropriate privileges will done at the view level.
         """
 
-        from audit.etl import ETL
+        # from audit.etl import ETL
```
neilmb, line 313: **question**: Is there an issue here with multiple workstreams that are not merged in yet? Why would we be commenting out a whole piece of functionality in this PR that is otherwise adding things?

backend/audit/templates/audit/not-ready-for-certification.html
```diff
@@ -0,0 +1,22 @@
...
+                  Not ready for Certification
+              </legend>
+              <p>While validating across sheets we encountered the errors below. You must correct these errors in order to proceed to certification.</p>
+              <p>Errors:</p>
+              <pre>{{ errors.errors | pprint}}</pre>
+              <p>Data submitted:</p>
+              <pre>{{ errors.data | pprint}}</pre>
+              <p>Audit {{ report_id }} has a submission status of: {{ submission_status }}</p>
+          </fieldset>
+      </form>
```
neilmb, line 19: **question**: Should this page contain an action button to link people back to the `report-submission` progress page so that they can start fixing these errors?

neilmb (approved):
No major issues, just some docstrings and questions. Is this still a Draft if a review was requested?

backend/audit/models.py
timoballard: Good catch - this was commented out earlier on when that ETL stuff was (I think) in a broken state

Issue #1279, opened 2023-06-17
Notes to SEFA workbook


jadudm:
### Background info

The "Notes to SEFA" portion of as submission used to a webpage/form, but we may want to turn it into a workbook instead that looks something like the following, though it might be possible to do this all in one sheet:

Sheet 1, mandatory notes:
| Describe significant accounting... | Did you use de minimis cost rate? | If yes, explain |
| --- | --- | --- |
| Description | Yes | We had to |

Sheet 2, additional notes
| Note title | Note content |
| --- | --- |
| Note one | This is a great note...|

* [Figma designs](https://www.figma.com/file/1bNncVt1TcFFDhQslqFC2o/FAC-Playground?type=design&node-id=5551-32825&mode=design&t=rIzhDf73erHw2b6D-0).
* [Previous discussion](https://github.com/GSA-TTS/FAC/issues/570).

### Tasks

```[tasklist]
- [ ] Convert notes to SEFA to Jsonnet
- [ ] Add unit tests for SEFA notes that check validation boundaries
- [ ] Add migration to add SEFA notes column in 2023 data model
- [ ] Add Django mapping to store notes to SEFA
- [ ] Add end-to-end test for happy path on notes to SEFA
```

Broken out for visibility from https://github.com/GSA-TTS/FAC/issues/1219

### Open questions

1. Do we need to pre-fill any of the notes? If not, do we need to provide additional text guidance? See figma mocks.

PR #1491, opened 2023-07-12
1175 fac bug missing formulas in federalawards xlsx federal program total and cluster total columns

sambodeme:
## Description

- Added missing formulas to auto calulate federal program total and cluster total in federal awards xlsx.
- Added an index column to federal awards xlsx and updated federalAwards JSON object accordingly. 
- Added some missing columns to federal awards sheet.
- Added logic to format columns as text/dollar when needed, to hide columns when needed, to lock columns when needed and to hide rows when needed.
- Fixed various bugs.
- Code cleaning.


backend/audit/excel.py
```diff
@@ -413,17 +408,33 @@ def _extract_data(
     return result
 
 
+def _remove_invalid_award_entries(data):
+    # Filter out invalid entries
```
jadudm, line 412: I think it would be good to expand this comment for the future.

What is an `invalid entry`? I'm *assuming* you mean something where the... well, actually, I don't know. The idea of the intake filtering data is something that is worrying, so I thought I'd at least ask.

backend/schemas/output/excel/json/federal-awards-expended-template.json
```diff
@@ -302,7 +310,11 @@
                "title_cell": "S1",
                "type": "open_range",
                "validation": {
-                  "type": "NOVALIDATION"
+                  "allow_blank": "True",
+                  "custom_error": "The Audit Report Type must be empty if Major Program is \"N\"",
+                  "custom_title": "Invalid Audit Report Type",
+                  "formula1": "=IF(R{0}=\"Y\",AuditReportTypes!$A$2:$A$5,AuditReportTypes!$A$6:$A$6)",
```
jadudm, line 316: Should we add a named range to this workbook for the A2:A5 and A6:A6 ranges? We can add named ranges even if they're not used for extraction, and that makes our formulas cleaner. Thoughts?

backend/schemas/scripts/render.py
```diff
@@ -28,7 +28,8 @@
 
 MAGIC_MAX_ROWS = 1048576
```
jadudm, line 29: I have a suspicion we should get rid of the `MAGIC_MAX_ROWS` and `MAX_ROWS` variables? `XLSX_MAX_ROWS` is repetitive w/ `MAGIC`, but it is named better.

backend/schemas/source/excel/libs/SheetValidations.libsonnet
```diff
@@ -91,4 +99,13 @@ local ComplianceRequirementValidation = {
   YoNValidation: YoNValidation,
   ReferenceNumberValidation: ReferenceNumberValidation,
   ComplianceRequirementValidation: ComplianceRequirementValidation,
+  LoanBalanceValidation: LoanBalanceValidation,
+  AuditReportTypeValidation(formRef): {
+    type: 'list',
+    allow_blank: 'True', 
+    // FIXME MSHD: for improvement, will need to pull column refs from this formula and retrieve that dynamically.
+    formula1: '=IF(R{0}="Y",' + formRef + '!$A$2:$A$5,' + formRef + '!$A$6:$A$6)',
```
jadudm, line 107: Another place for a named range?

backend/schemas/source/excel/templates/federal-awards-expended-template.jsonnet
```diff
@@ -107,9 +115,11 @@ local open_ranges_defns = [
     'If Other Cluster, Enter Other Cluster Name',
     'other_cluster_name',
   ],
-  // 20230525 HDMS FIXME: A formula to auto calculate federal_program_total in the excel is missing (see instructions in census template)!!!
   [
     Sheets.open_range {
+      keep_locked: true,
+      format: 'dollar',
+      formula: '=SUMIFS(E:E,V:V,V{0})',
```
jadudm, line 122: Also a place for named ranges? Or, do we need to use the cell refs, because of the `{0}`?

backend/schemas/source/excel/templates/federal-awards-expended-template.jsonnet
```diff
@@ -207,17 +222,65 @@ local open_ranges_defns = [
...
+      help: Help.unknown,
+    },
+    SV.NoValidation,
+    'Award Index (Read Only)',
+    'award_index',
+  ],
+  [
+    Sheets.open_range {
+      keep_locked: true,
+      formula: '=CONCATENATE(A{0},B{0})',
```
jadudm, line 239: For discussion/ticketing for another day:

We should, at this point, create a JSON mini-language for specifying the formulas.

E.g.

```
formula: { 
  function: CONCATENATE
  parameters: [
    { cell: { column: "A", row: CURRENT } },
    { cell: { column: "B", row: CURRENT } },
  ]}
```

or similar. Why? We're already parsing the JSON structure, and then walking it (compiling it) to generate the XLSX docs. Eliminating complex strings (the formulas, for example) and replacing them with a well-defined JSON object, that can be turned into a string with the appropriate values in a clean and reliable way, would make sense.

This is not a comment for this code right now, but a note for us to remember/ticket/consider in the future. Point being: use a data structure as opposed to trying to embed a language/semantics into a string.

backend/audit/excel.py
```diff
@@ -395,7 +390,7 @@ def _extract_data(
                 set_fn(
                     result,
                     f"{parent_target}[{index}].{field_target}",
-                    _ensure_string_conversion(field_target, value),
+                    value,
```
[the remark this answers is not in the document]
sambodeme: I added logic to format columns as text (or currency) when needed, so this code became unnecessary.

backend/audit/excel.py
sambodeme: Sorry, I meant to add comments to a few places before publishing this PR. However, as I switched my focus to #1279, I lost track of my thoughts. After adding formulas to auto-calculate federal awards total and cluster total on the sheet, Excel set all the targeted cells associated with these named ranges to 0. So, when uploading federalAwards, Python mistakenly reads extra rows (consisting solely of empty cells but two set to 0) into the JSON object. To resolve this issue, I implemented this method to filter out these unwanted rows (which I am referring to as invalid here). I will update the comment here and possibly rename the method for clarity once I push 1279.

backend/schemas/source/excel/libs/SheetValidations.libsonnet
sambodeme: @jadudm This is the same code you commented on earlier. Actually, what you commented on earlier was the output generated jsonnet whereas this code is the source. What I did was to parameterize the formula, treating it like a function. This allows me to pass a value for "formRef" wherever I need to utilize that formula (in the xlsx template). But, I agree with your suggestion. A named range would be simpler in this case.

backend/schemas/source/excel/templates/federal-awards-expended-template.jsonnet
sambodeme: I believe we could change E:E and V:V to named ranges but not V{0}. Also, I will have to try this in each case as I am not sure how excel react to named range in formulas versus cell refs

backend/schemas/scripts/render.py
jadudm: Yep. I... suspect the 3000 comes from me discovering that slowdown at some point in the past. Thank you for the reminder. (Also, all these questions/comments are largely conversation... not critique.)

backend/data_fixtures/audit/excel_schema_test_files/federalawards-pass-01.json
```diff
@@ -4,37 +4,38 @@
...
+                ]
+              },
+              "cluster": {
+                "cluster_name": "N/A",
+                "cluster_total": 0
+              },
+              "subrecipients": {
+                "is_passed": "N"
+              },
+              "award_index": "AWARD-00001"
```
sambodeme, line 38: @jadudm Any reason award key should be a string (e.g. AWARD-00001) versus a number (e.g. 1)? I have no preference.

backend/schemas/source/sections/FederalAwards.schema.jsonnet
```diff
@@ -327,6 +326,7 @@ local FederalAwardEntry = Types.object {
     loan_or_loan_guarantee: Parts.LoanOrLoanGuarantee,
     program: Parts.Program,
     subrecipients: Parts.Subrecipients,
+    award_index: Types.string,
```
sambodeme, line 329: This is an auto-generated key in FederalAwards.xlsx. It is required in order to (uniquely) identify each award but the 'required' flag cannot be enforced via JSON schema. The reason for this is that when the xlsx sheet is auto-filled by python, auto-generated cells (like award_index which are read-only) do not get populated, causing all award validation tests to fail. Since the "award_index" get populated correctly when manually filling the sheet I am suggesting that we validate this field exists for each award during cross-validation. FYI, a unique identifiyer will be needed for linking awards and findings. Any thought ???

sambodeme: See #1514

backend/data_fixtures/audit/excel_schema_test_files/federalawards-pass-01.json
sambodeme: We maintain the text format but reduce to 4 digit. i.e. AWARD-#### instead of AWARD-#####.

danswick (approved):
These changes look good to me, @sambodeme 👍

PR #1493, opened 2023-07-12
Pass API version in all the places

danswick:
Proposal to address failing tests in the API versioning branch.

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -56,7 +56,7 @@ docker-first-run:
 	docker compose run web bash -c '\
 		python manage.py makemigrations &&\
 		python manage.py migrate &&\
-		python manage.py create_views \
+		python manage.py create_views --api_version "api_v1_0_0_beta"\
 		'
 	# docker compose run web python manage.py load_test_data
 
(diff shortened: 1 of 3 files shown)


backend/Makefile
```diff
@@ -56,7 +56,7 @@ docker-first-run:
 	docker compose run web bash -c '\
 		python manage.py makemigrations &&\
 		python manage.py migrate &&\
-		python manage.py create_views \
+		python manage.py create_views --api_version "api_v1_0_0_beta"\
```
danswick, line 59: I didn't add the `create_views` function here because it's probably better to do this in one place.

danswick: Maybe `create_views` just creates every versioned view _unless_ it's passed an `--api_version` flag for development/testing purposes.

jadudm: That's a good idea. Note that this all wants for a revision. I slammed the API versioning in pretty quick.

jadudm (approved):
Thank you.

This particular commit/feature is, I think, a symptom of having too many people split too many ways.

Thank you for fixing my `bash`. And, I agree: we should have the Python script create all of the views by default. Much better than having that value denoted in multiple places. We need to get that list of versions in *one* place.

Issue #1278, opened 2023-06-17
Audit information form


jadudm:
### Tasks

```[tasklist]
- [x] Convert audit information form to web form
- [x] Add unit tests for audit information form checking validation boundaries
- [x] Add migration to add audit information column in 2023 data model
- [x] Add Django mapping to store audit information form
- [x] Add Jsonnet validation for audit information form
```

Broken out for visibility from https://github.com/GSA-TTS/FAC/issues/1219

jadudm:
@eazapata23 , I removed the E2E bullet from this ticket, because the testing work is going to pick it up as a matter of course.

jadudm:
Starting a branch

https://github.com/GSA-TTS/FAC/tree/jadudm/audit-information-schema

for the Jsonnet/validation schema piece.

JeanMarie-TTS:
@jperson1 Are the following fields included?
1.  sp_framework
2.  is_sp_framework_required
3.  sp_framework_auditor_opinion

PR #1497, opened 2023-07-13
Audit Info Form Frontend

jperson1:
# Audit Info Form Frontend - Django View/HTML/JS

Issue #1278

## Changes
1. New `AuditInfoFormView` under `audit` that handles the getting/posting of audit info form. 
2. New HTML `audit-info-form.html`, which is a formatted form that should match very closely to the associated [Figma doc](https://www.figma.com/file/1bNncVt1TcFFDhQslqFC2o/FAC-Playground?node-id=4954%3A24748&mode=dev). 
3. New JS file to handle requiring of checkboxes. HTML5 doesn't have a good way to 'require' multiple choice selections (that thing that jumps you to an incomplete question when you don't answer it). So, I use JS to require/unrequire to mimic the correct behavior. 

Form validation is handled in Django. Unanswered questions will be jumped to if you try to submit without answering them. All questions available in the form are required, so you have to answer everything. 

## How to test
Switch to this branch and run normally. These changes should not require a rebuild, but may require you to restart your container to recognize the JS. Navigate to your [audit home page](http://localhost:8000/audit/submission-progress/YOURIDHERE), and click "[Upload the audit report PDF](http://localhost:8000/audit/audit-info/YOURIDHERE)". Fill out the form. Successful entries will redirect you to your audit home page, and will not currently save anything to the DB.

## Screenshots:
image
image


tadhg-ohiggins:
We probably want to consider loading the agency list once, at app startup, rather than each time a user hits the form.

backend/audit/views.py
```diff
@@ -428,6 +430,61 @@ def get(self, request, *args, **kwargs):
...
+            sac = SingleAuditChecklist.objects.get(report_id=report_id)
+            form = AuditInfoForm(request.POST, request.FILES)
+
+            if form.is_valid():
+                # Save the form, yay!
+                print("Saving audit info form!")
+                return redirect(reverse("audit:SubmissionProgress", args=[report_id]))
+            else:
+                form.clean_booleans()
+                print("hello", form.cleaned_data)
```
jadudm, line 470: Does `views.py` have `logging` imported? If so, please feel free to include some logging statements. Or, perhaps we should remove the `print` statements.

Point being: having some noise come from the submission process is fine. Having it go to the logger means it will (ultimately) make its way to our New Relic instance. This is fine, too. (I do want to do a walk of the code and add logging throughout, so we can debug, etc. But, this is a *should* for go-live, not a *must*, as crash logs will still make it through.)

Either way, don't be afraid to include some logging. Just make sure the logging message makes clear what you're doing (instead of just `"hello"`, for example, from a few lines down.)

jperson1: Yep, it sure does. The prints are from me neglecting to clean up after myself. The print from the audit info form has been changed to use logger, and the others have been removed. Thanks for catching that!

jadudm (approved):
Looks good. Lets get this in `main` so that we can integrate it into E2E and UAT.

Issue #1471, opened 2023-07-11
Epic: workbook refinement 


danswick:
## Description

Once workbooks are completed, add more data validation, formatting, etc.

**Who**: Submitters
**What**: Submitting valid data
**Why**: Because we want confidence that forms will be submitted correctly

```[tasklist]
### Stories
- [ ] https://github.com/GSA-TTS/FAC/issues/1431
- [ ] https://github.com/GSA-TTS/FAC/issues/1175
- [ ] https://github.com/GSA-TTS/FAC/issues/1227
- [ ] https://github.com/GSA-TTS/FAC/issues/1269
- [ ] https://github.com/GSA-TTS/FAC/issues/1435
- [ ] https://github.com/GSA-TTS/FAC/issues/1513
- [x] Add `OTHER_CLUSTER` (in `Base`) to the list of cluster names allowed (per HDMS) in the Jsonnet validation
- [x] https://github.com/GSA-TTS/FAC/issues/1617
- [ ] https://github.com/GSA-TTS/FAC/issues/1660
- [ ] https://github.com/GSA-TTS/FAC/issues/1473
- [ ] https://github.com/GSA-TTS/FAC/issues/1472
- [ ] https://github.com/GSA-TTS/FAC/issues/1519
- [ ] https://github.com/GSA-TTS/FAC/issues/1470
```

#1341 will close via this epic.

Issue #1175, opened 2023-05-31
FAC - Bug: Missing Formulas in FederalAwards xlsx (Federal Program Total and Cluster Total columns)


sambodeme:
### Describe the bug

The FederalAwards xlsx template does not have formulas for Federal Program Total column and Cluster Total column. This is a miss since the original FederalAwards xlsx has a formula for each of these columns.
Also, it is not clear if these two columns are needed, TBD. 

### Steps to reproduce the bug

1. Download the new template [here](https://github.com/GSA-TTS/FAC/blob/main/backend/schemas/output/excel/xlsx/federal-awards-expended-template.xlsx) 
2. Download the old template [here ](https://github.com/GSA-TTS/FAC/blob/main/backend/excel_templates/FederalAwardsExpendedTemplateUG2019.xlsx)
3. Compare the formulas associated with respectively with Federal Program Total column and Cluster Total column from each file.

### Expected Behavior

The formulas associated with Federal Program Total column and Cluster Total column should be captured in the new templates or these column should be removed.

### Screenshots

![image](https://github.com/GSA-TTS/FAC/assets/26260777/f37c489c-29b6-45ba-9233-825183e6fac5)

### Code of Conduct

- [X] I agree to abide by the [Digital.gov Community Guidelines](https://digital.gov/communities/community-guidelines/) and the [TTS Code of Conduct](https://handbook.tts.gsa.gov/code-of-conduct/). Respect your peers, use plain language, and be patient.
- [X] I checked the [current issues](https://github.com/GSA-TTS/FAC/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

PR #1500, opened 2023-07-13
1279 notes to sefa workbook

sambodeme:
## Description
- Added Notes for SEFA Workbook


JeanMarie-TTS (commented):
@sambodeme There can be any number of notes in the notes_to_sefa. Please confirm that there is a sequence number field in each entry. I'm not sure how to confirm this by looking at the changes in the PR.

jadudm:
We don't currently have an explicit sequence number in the sheets. Do we need an auto-generated seq_number? If so, does it cross-validate against another sheet?

JeanMarie-TTS:
> We don't currently have an explicit sequence number in the sheets. Do we need an auto-generated seq_number? If so, does it cross-validate against another sheet?

There is no cross validation against anything else. However, we do need a sequence number as there is nothing else to distinguish one note from another. I'd rather not have ETL generate a sequence number.

jadudm:
Got it.

@sambodeme , do you think we can add a hidden seq_number field that auto-populates? Or, does this take us into the space of other questions that we haven't had a chance to touch base on? 

Happy to do a call to figure out what is the best/most consistent approach here.

sambodeme:
> @sambodeme There can be any number of notes in the notes_to_sefa. Please confirm that there is a sequence number field in each entry. I'm not sure how to confirm this by looking at the changes in the PR.
@JeanMarie-TTS, @jadudm  I am including it now. I started to include this seq number by default for xlsx files that involve multiple entries but I was advised that it may not always be necessary, so I backed off that plan. 
Please let me know the xlsx files you need a sequence number for, and I will make sure to address them in a separate PR (if necessary).

jadudm:
How about we add any seq_number additions here:

https://github.com/GSA-TTS/FAC/issues/1471

which we know we want to do, but it puts it on an epic that's about touching all of the workbooks in a consistent way/final way before going liev.

sambodeme:
> How about we add any seq_number additions here:
> 
> #1471
> 
> which we know we want to do, but it puts it on an epic that's about touching all of the workbooks in a consistent way/final way before going liev.

That was my intention, but it seems that this won't be useful if the only remaining Excel file is this one, as I have already added the sequence number to it. However, if there are more files that require a sequence number, I will definitely open a ticket and add it to the queue. Previously, I created a ticket for this, but JeanMarie mentioned that the required sequence numbers are already addressed through another ticket, so I closed mine. Therefore, I wanted to confirm if there is still a need for it before creating another ticket.

backend/schemas/source/excel/templates/notes-to-sefa-template.jsonnet
```diff
@@ -0,0 +1,106 @@
...
+    Sheets.open_range {
+      width: 56,
+      help: Help.plain_text,
+    },
+    SV.NoValidation,
+    'Note content',
+    'note_content',
+  ],
+  [
+    Sheets.open_range {
```
sambodeme, line 71: This should be read-only and hidden (if necessary). The logic to enforce these two requirements is currently sitting in #1175. Depending on which ever gets merged first, I will update the next one accordingly.

jadudm (approved):
It is the Notes to SEFA form. We have additional updates coming as part of #1175 as well the final walkthrough of the notebooks. LGTM, and thank you!

PR #1501, opened 2023-07-13
Squash dissemination migrations

pluribus-tony:


tadhg-ohiggins (approved):

backend/dissemination/migrations/0001_initial.py
```diff
@@ -1282,15 +1282,15 @@ class Migration(migrations.Migration):
                     ),
                 ),
                 (
-                    "modified_opinion",
+                    "is_modified_opinion",
```
sambodeme, line 1285: Is this done manually ? I would expect Django to auto generate migration files but I cannot find the source (i.e., what triggered these name changes)?

tadhg-ohiggins: @sambodeme it's in `dissemination/models.py`:

```
is_modified_opinion = models.BooleanField(
```

And the “change” comes from deleting the original migration file, which was presumably generated at a point when the model referred to `modified_opinion` instead.

sambodeme: Got it, I was expecting a change in models to trigger this (I missed the removal of the original migration file).

sambodeme (approved):

PR #1505, opened 2023-07-13
Add finding to ETL pipeline.

pluribus-tony:


backend/audit/etl.py
```diff
@@ -58,25 +58,19 @@ def load_findings(self):
         for entry in findings_uniform_guidance_entries:
             findings = entry["findings"]
             finding = Finding(
+                award_seq_number=entry.get("seq_number", 1),
```
sambodeme, line 61: I believe this is the award key. There is a [PR](https://github.com/GSA-TTS/FAC/issues/1175) open to add it to FederalAward Json. The value may be a number or a string depending on ongoing discussion. It is currently a string (e.g., AWARD-#####) named [award_index](https://github.com/GSA-TTS/FAC/blob/1175-fac-bug-missing-formulas-in-federalawards-xlsx-federal-program-total-and-cluster-total-columns/backend/schemas/source/sections/FederalAwards.schema.jsonnet#L329).

backend/audit/etl.py
```diff
@@ -58,25 +58,19 @@ def load_findings(self):
...
+                report_id=self.report_id,
+                finding_seq_number=1,  # TODO: Where does this come from?
```
sambodeme, line 63: Most likely it will be this value ["seq_number"](https://github.com/GSA-TTS/FAC/pull/1505/files#diff-bf46162f84e6a008a08ed3d0e697189966a10bbd12f4d2f1dacde7c72b20c60cR61)

backend/audit/etl.py
pluribus-tony: Pushed a new commit addressing this.

JeanMarie-TTS (approved):

backend/audit/etl.py
```diff
@@ -58,25 +58,19 @@ def load_findings(self):
         for entry in findings_uniform_guidance_entries:
             findings = entry["findings"]
             finding = Finding(
+                award_seq_number=entry["award_index"],
```
sambodeme, line 61: There is an open question for Matt on #1175 that might impact award_index type. Once he provides feedback, I will update you @pluribus-tony. Otherwise, this PR looks good to me.

sambodeme: @pluribus-tony it was not me (I was offline and just got back). I was going to get back to you with some clarification about my previous comment first. If the field "award_index" remains a string, I would suggest to rename award_seq_number something like award_key instead. However, if "award_index" is changed to a number then "award_seq_number" does not need be renamed.

PR #1508, opened 2023-07-14
Use save() instead of update()

tadhg-ohiggins:
`save()` is centralized,
These `update()`s aren’t needed,
So switch them over.

-----

We will need to funnel `SingleAuditChecklist` changes through a single point, so switch to `save()` from `update()` where we were using the latter.

Also some linter appeasement.


neilmb (approved):
LGTM

Issue #1050, opened 2023-04-26
Upload Page Timeout


jperson1:
Related to https://github.com/GSA-TTS/FAC/issues/928

# At a glance

The user is notified if their file upload was unsuccessful due to a timeout. 

# Acceptance Criteria

In the event of a server timeout for whatever reason (long processing time, crash, etc.), the frontend should timeout the response and let the user know. 

# Background

The relevant HTTP request is made in  `/backend/static/js/upload-page.js`. Some sort of timeout function should be written that aborts this request and updates the `info_box` on the frontend.

jadudm:
This was previously implemented, then commented out. We need, as a team, to decide what the timeout should be, and why. Then, we'll bring it back.

jadudm:
Lets choose 30 seconds. :gavel: Arbitrary. To be revisited based on testing.

mogul:
Note the lack of a timeout limit came up explicitly in our assessment. I'm leaving this a `SHOULD` for now, but it _may_ need to move up to a `MUST` if it's determined to be a big risk blocking our ATO (seems doubtful, but...).

PR #1517, opened 2023-07-17
Upload Page Timeout

jperson1:
# Upload Page Timeout (and Some JS Cleaning?)

Issue: #1050 

## Changes:
1. Implemented the 30s timeout for XLSX upload. If the frontend doesn't receive anything by then, it aborts the request and lets the user know.
2. Some minor cleanup of error reporting. All the `info_box` changes and `console.error()`'s are in one place, at least.
3. Moved the submit button below the `info_box`, so users are better able to see the feedback being presented. 

## How to test:
2. Navigate to an XLSX upload page, like http://localhost:8000/audit/submission-progress/SACIDHERE. 
3. You can try to find an upload that takes longer than 30 seconds, or you can navigate to `/backend/static/js/globals.js` and change it to a short value, such as 500 for 0.5 seconds. 
4. Upload a .xlsx file, and see a relevant error message both on the page and in the browser console.

## Screenshot:
image


jadudm (approved):
@jperson1 , this looks good. There's a few changes that feel like they're not strictly about the timeout. I'm going to approve, and you do the merge if you feel good about what is in this PR. 

(I'm seeing things like name changes on error handlers, etc., which I'm going to trust you to know to be OK, given that you're living in this space.)

PR #1520, opened 2023-07-18
Version control the PostgREST and ClamAV-rest imagerefs

mogul:
See discussion here in Slack:
https://gsa-tts.slack.com/archives/C03QCP4FSN8/p1689279742600559?thread_ts=1689273606.817079&cid=C03QCP4FSN8


terraform/shared/modules/env/clamav.tf
```diff
@@ -11,11 +11,9 @@ module "clamav" {
 
   cf_org_name   = var.cf_org_name
   cf_space_name = var.cf_space_name
-  clamav_image  = "ajilaag/clamav-rest:20230228"
+  clamav_image  = "ghcr.io/gsa-tts/fac/clamav:latest"
```
asteel-gsa, line 14: (Non-Blocking)

We can use YYYYMMDD if we want here, however, the assumption is that `:latest` will always be a "clean" image, as in order for it to get into our GHCR, it needs to have 0 scanned items via trivy.

terraform/shared/modules/env/postgrest.tf
```diff
@@ -16,7 +16,7 @@ resource "cloudfoundry_service_key" "postgrest" {
 resource "cloudfoundry_app" "postgrest" {
   name         = local.postgrest_name
   space        = data.cloudfoundry_space.apps.id
-  docker_image = var.postgrest_image
+  docker_image = "ghcr.io/gsa-tts/fac/postgrest:latest"
```
asteel-gsa, line 19: (Non-Blocking)

We can use YYYYMMDD if we want here, however, the assumption is that `:latest` will always be a "clean" image, as in order for it to get into our GHCR, it needs to have 0 scanned items via trivy.

asteel-gsa (approved):
Looks good, allows us to change images and a hash will be stored in commit if we do change images. `:latest` should theoretically always be fine.

terraform/shared/modules/env/clamav.tf
mogul: For the purposes of this PR, I am just replicating what was in the workflow.

I figured we can pin the ref to a specific date by making a PR on this line if we ever have a problem unrelated to scanning.

terraform/shared/modules/env/postgrest.tf
mogul: [Same as above](https://github.com/GSA-TTS/FAC/pull/1520#discussion_r1267085468)

JeanMarie-TTS:
> @JeanMarie-TTS leaving this here for you to comment/ask questions if needed. Otherwise, please merge!

If the 'latest' tag is used, how will TF know if an image has been updated in ghcr? We need to pick up a named tag like 20230718. Not sure how exactly the action that creates this tag can set a variable for TF to use in a subsequent deploy.

mogul:
> > @JeanMarie-TTS leaving this here for you to comment/ask questions if needed. Otherwise, please merge!
> 
> If the 'latest' tag is used, how will TF know if an image has been updated in ghcr? We need to pick up a named tag like 20230718. Not sure how exactly the action that creates this tag can set a variable for TF to use in a subsequent deploy.

We were [specifying `latest` in the GitHub Action workflow already](https://github.com/GSA-TTS/FAC/pull/1520/files#diff-b93545a2c0b7632dad79e706e4b7375b2134c9cff0349cc26688a9b42648a157L24-L25) so this doesn't PR change that situation. 

However, your point that there's no way for Terraform to know about a new image without the specific tag is a good one! So, I suggest we have Terraform retrieve info about the `latest` image using [this data source](https://registry.terraform.io/providers/kreuzwerker/docker/latest/docs/data-sources/registry_image), then change the imageref we use for deployment from `":latest"` to `"@sha256:${docker_resource_image.postgrest.sha256_digest}"`, for example.  I'll add a commit that does that!

asteel-gsa:
@jean

> @mogul , @asteel-gsa Not sure how we got to always fetching the latest tag in TF. If you see the file history of terraform-apply-env.yml there used to be some logic to determine the version tag.

@JeanMarie-TTS that artifacting was removed from the code base. https://github.com/GSA-TTS/FAC/commit/5cde89071531e95734f65c880a729ad396e12008#diff-b93545a2c0b7632dad79e706e4b7375b2134c9cff0349cc26688a9b42648a157

It was confusing using 3 different tags for the containers. It is currently running on `:latest` and has been since the above commit https://github.com/GSA-TTS/FAC/actions

JeanMarie-TTS:
The plan was for docker-compose to use the latest tag and for TF to use a specific version i.e. the last one to be be posted into ghcr.

mogul:
> TIn any case, we need to figure out how to get back to a state where TF will recognize when versions have changed and do the right thing.

My proposal above will do that:

> I suggest we have Terraform retrieve info about the latest image using [this data source](https://registry.terraform.io/providers/kreuzwerker/docker/latest/docs/data-sources/registry_image), then change the imageref we use for deployment from ":latest" to "@sha256:${docker_resource_image.postgrest.sha256_digest}", for example. I'll add a commit that does that!

mogul (commented):
I can't give this a ✅ because I'm one of the PR contributors. @JeanMarie-TTS can you approve?

mogul:
(Note [the plan](https://github.com/GSA-TTS/FAC/pull/1520#issuecomment-1640633690) which shows this is doing exactly what we want.)

asteel-gsa:
this has an ❌ because the e2e tests are currently set to run on PRs into main, fixed in #1611

mogul:
Can we merge that PR and then update this branch to incorporate the fix?

asteel-gsa:
> Can we merge that PR and then update this branch to incorporate the fix?

Think its still under review. I approved the actions, but ultimately @neilmb call to merge now or wait for another reviewer

PR #1524, opened 2023-07-18
1513 update federal awards and audit findings for award reference linking

sambodeme:
## Description

- Updated federal awards expended and moved the audit index to the far left.
- Renamed as "award reference" instead of "award index".
- Changed size to 4 digits (e.g. `AWARD-1234` vs `AWARD-12345`).
- Updated formulas due to column shift and used named ranges wherever possible.
- Updated unit/validation tests as necessary.
- Updated validation in audit findings workbook to reflect new size.
- Updated audit findings workbook to get rid of first four columns (agency prefix, extension, additional id, and prorgam name), and replaced with award reference.
- Updated mapping for audit findings to reflect new workbook shape.
- Updated validation schema for audit findings to reflect new workbook shape.


jadudm (approved):
Looks good.

Something I'm ticketing for revision/later: we have some errors that are "in the workbook," as opposed to out in the centralized error file. We'll want to move those out, so that Laura can review the content more easily. But, this is not critical at this time.

PR #1525, opened 2023-07-18
Add Note to ETL

pluribus-tony:


backend/audit/etl.py
```diff
@@ -133,16 +133,17 @@ def load_captext(self):
...
-            type_id="",  # TODO: What is this?
-            report_id=self.report_id,
-            version="",  # TODO: What is this?
-            sequence_number=-1,  # TODO: Where does this come from?
-            note_index=-1,  # TODO: Where does this come from?
-            content="",  # TODO: Where does this come from?
-            title="",  # TODO: Where does this come from?
-        )
-        note.save()
+        notes_to_sefa = self.single_audit_checklist.notes_to_sefa
```
sambodeme, line 136: I am not seeing the mandatory note items: `accounting_policies` and `is_minimis_rate_used` and `rate_explained`. FYI, `is_minimis_rate_used` can take any of these values: `Y` or `N` or `Y&N`.

pluribus-tony: None of those fields are in the fac.md file which is what I use as reference for what we need to pull in.

sambodeme: [Notes for SEFA](https://www.figma.com/file/1bNncVt1TcFFDhQslqFC2o/FAC-Playground?type=design&node-id=5551-32825&mode=design&t=02UWMva4DecKY1q1-0) consist of two sections: a mandatory section, indicated by a red mark on the screenshot, and an optional section, indicated by a blue mark.
The mandatory notes are single fields within the NotesToSefa object, meaning that there can only be one entry for each of these fields.
On the other hand, the additional notes (in the optional section) are represented as an array within the NotesToSefa object since there can be zero or multiple entries for these notes.
image

JeanMarie-TTS: @pluribus-tony  - This is my understanding. There are three note types:1, 2 and 3.  There is exactly one of Type 1 and its content is what @sambodeme is referring to as accounting_policies. There is exactly one note of type 2 - its content is rate_explained, and it should also store is_minimis_rate_used (perhaps note_index is used for this in the dissemination model). There can be any number of notes of type 3 with a title and content. 
I would recommend we implement it this way and have a discussion with Carley to make sure this is correct.

pluribus-tony: Added fields in new push.

JeanMarie-TTS (approved):

PR #1530, opened 2023-07-19
Read DOCKERFILE from .env with default value

pluribus-tony:


jadudm (approved):
Per convo in slack. Good stuff.

PR #1538, opened 2023-07-20
Disallow changes to submissions not `in_progress`

tadhg-ohiggins:
It’s not in progress.
So you can’t write to it now.
Regrets? That’s too bad.

-----

Add code to prevent changing submissions that have moved out of `in_progress`.


backend/audit/test_models.py
```diff
@@ -103,6 +110,25 @@ def test_submission_status_transitions(self):
...
+        """
+        bad_statuses = [
+            status[0]
+            for status in SingleAuditChecklist.STATUS_CHOICES
+            if status[0] != SingleAuditChecklist.STATUS.IN_PROGRESS
+        ]
+
+        for status_from in bad_statuses:
+            sac = baker.make(
+                SingleAuditChecklist, audit_type="annual", submission_status=status_from
```
timoballard, line 126: I think we're mixing up audit type and audit period here

backend/audit/views.py
```diff
@@ -598,6 +602,9 @@ def post(self, request, *args, **kwargs):
                 return render(request, "audit/upload-report.html", context)
         except SingleAuditChecklist.DoesNotExist:
             raise PermissionDenied("You do not have access to this audit.")
-        except Exception as e:
-            logger.info("Unexpected error in UploadReportView post.\n", e)
-            raise BadRequest()
+        except LateChangeError:
+            return render(request, "audit/already-certified.html")
```
timoballard, line 606: Nitpick: I wonder if we should call this something other than "already certified", since "ready for certification" (but not actually certified) would also fall into this late change category

timoballard (approved):
some minor things but otherwise 👍

backend/audit/test_models.py
tadhg-ohiggins: Yup, I changed that to actually have audit types here.

backend/audit/views.py
tadhg-ohiggins: I changed this to `no-late-changes.html`

PR #1551, opened 2023-07-20
1431 fac bug compliance requirement error on f4 in findings uniform guidance xlsx

### LINKED ISSUE #1431: FAC - Bug: Compliance Requirement error on F4 in Findings Uniform Guidance xlsx
### Describe the bug

When I open Findings Uniform Guidance xlsx
And select field F4
And enter a vaild combination of the letters: A,B,C,E,F,G,H,I,J,L,M,N,P (for e.g., AB or ABC)
Then I get a Compliance requirement error message

### Steps to reproduce the bug

1. Open Findings Uniform Guidance xlsx form
2. Select field F4
3. Enter a vaild combination of the letters: A,B,C,E,F,G,H,I,J,L,M,N,P 
4. A compliance requirement error message should appear preventing you from saving and proceeding further 

### Expected Behavior

You should be to  enter a a vaild combination of the letters: A,B,C,E,F,G,H,I,J,L,M,N,P 
and save and exit form. 

### Screenshots

image

https://github.com/GSA-TTS/FAC/assets/99770283/86b627f3-7081-45f3-9051-0f8adb384aed

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

- [X] I agree to abide by the [Digital.gov Community Guidelines](https://digital.gov/communities/community-guidelines/) and the [TTS Code of Conduct](https://handbook.tts.gsa.gov/code-of-conduct/). Respect your peers, use plain language, and be patient.
- [X] I checked the [current issues](https://github.com/GSA-TTS/FAC/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

sambodeme:
## Description

- Added a list of compliance requirement types.
- Used newly added list for validation in findings-uniform-guidance.xlsx
- This resolves #1431


backend/schemas/source/base/ComplianceRequirementTypes.json
sambodeme, line 1: @jadudm I did not find a CSV file for this in Google drive so I grabbed this list from Census template [here](https://github.com/GSA-TTS/FAC/blob/main/backend/excel_templates/FindingsUniformGuidanceTemplate2019-2022.xlsx). How do you feel about this?

backend/schemas/source/base/ComplianceRequirementTypes.json
```diff
@@ -0,0 +1,8195 @@
+{
```
neilmb, line 1: **question**: Is this too limiting because the multiple letters have to be in alphabetical order? If I were an auditor, I don't know if I would think to put them in this order and if I didn't, would the error message be clear enough to tell me that I should do that?

sambodeme: In previous workbooks from Census, this is how requirement types were input. No changes are being introduced here, except for the fact that the cells were not formatted as a list as they were in Census workbooks. So this ticket was just to rectify this by converting the cells back into a list format, just as they were before. Also, I have included a sheet in the workbook that contains the entire list, allowing users to refer to it if necessary.

sambodeme: About the error, I am tagging @jadudm. As of now we have a generic error message for any invalid list input that reads `Not in the lookup list`. Should we make an exception here and use the message previously used by Census `Please enter a valid Type of Compliance Requirement. (One or more of ABCEFGHIJLMNP, in alphabetical order).` ?

jadudm: Oh, yes, by all means... error message improvements are a good idea. Ultimately, the `link` field in the error will let us link back to `fac.gov`, where we can write a more thorough story about what the error is about. But, in the meantime, feel free to update that error message so it tells the user exactly what is going on.

backend/schemas/source/base/ComplianceRequirementTypes.json
jadudm: I don't know if there ever was a CSV for that. I think that's the list.

We could, in theory, add the list, and then put it in a sheet, with the descriptions. However, we still would need to provide the ordering check. So, for now, this seems good to me. (Unless I'm misunderstanding. But, I don't think I am.)

backend/schemas/source/base/ComplianceRequirementTypes.json
sambodeme: Done!, I used the same message from Census workbook.

sambodeme:
@jadudm / @neilmb can I get a thumb up on this PR ?
FYI, I included a fix to the issue that prevented the code from blocking xlsx files with missing required fields (such as UEIs, etc.) to upload.

jadudm (approved):
Thank you; looks good, and good additions/abstractions in the Jsonnet libraries for this. Merging!

PR #1554, opened 2023-07-20
Ez/cypress test suite updated

eazapata23:
Added updated workbooks and modified changes to e2e


.github/workflows/end-to-end-test.yml
```diff
@@ -1,6 +1,10 @@
 name: Runs end to end tests
 on:
-  workflow_call:
+  # workflow_call:
+  # workflow_dispatch:
+  pull_request:
+    branches:
+      - main
```
neilmb, line 7: **comment**: I don't think that these tests are ready to run against every pull request to main. Let's put this whole section back to just `workflow_call:` as it is on `main`.

backend/cypress/e2e/audit-findings-text.cy.js
```diff
@@ -1,36 +1,40 @@
...
+  describe('File upload fail', () => {
+    it('unsuccessful upload audit findings text', () => {
+      cy.intercept('POST', '/audit/excel/findings-text/*', {
+        statusCode: 400,
+        fixture: 'fail-res.json',
+      }).as('uploadFail')
+      cy.visit(`/report_submission/audit-findings-text/${reportTestId}`);
+      cy.get('#file-input-audit-findings-text-xlsx').attachFile('federal-awards-Test.xlsx');
+      cy.wait('@uploadFail').its('response.statusCode').should('eq', 400)
+      cy.wait(2000).get('#info_box').should('have.text', 'Field Error: undefined');
```
neilmb, line 36: **question**: I see where you are using this text to be able to handle the 400 result using the `#info_box` element, but I'm not sure if the frontend team is ready to make the change that uncaught upload errors should show up on the page like this. @jperson1 @heymatthenry Can you look at this and the change in `static/js/upload-page.js`?

backend/cypress/e2e/audit-findings.cy.js
```diff
@@ -1,35 +1,39 @@
 import 'cypress-file-upload';
 
 describe('Audit findings page', () => {
-    before(() => {
-      cy.visit('/report_submission/audit-findings/2022XB40001000002');
-    });
-    it('Page loads successfully', () => {
-      cy.url().should('include','/report_submission/audit-findings/2022XB40001000002');
-    });
+  const reportTestId = '2023MAY0001000001'
```
neilmb, line 4: **question**: I don't understand how the actual audit with this ID gets created in the first place. Moving forward with completely automated tests, do we need to use some kind of Cypress command to create the initial audit and get the ID of it?

neilmb (approved):
I am excited to see this on main @eazapata23. I had one question that needs attention from someone who knows the frontend better than me.

backend/cypress/e2e/audit-findings.cy.js
eazapata23: Initially, the report id is added after a one completes a create new audit then I manually enter that ID into Cypress. I prefer not to do it this way, I would ideally like to create a Cypress Command to bypass that, so yes.

backend/cypress/e2e/audit-findings-text.cy.js
jperson1: As of right now, that info box will populate with a number of different things depending on what's uploaded. The change in `static/js/upload-page.js` is a catch all for an unexpected field error (this suggests an incorrect template). That's fine for now, because on `main` the user may receive no feedback on the page. The error handling in there is due for some changes, some of which are captured in this [draft PR](https://github.com/GSA-TTS/FAC/pull/1517). If this goes in before I complete that PR, I'll be sure to update the tests accordingly. 

TL;DR I think it's fine, and it will certainly be fine after https://github.com/GSA-TTS/FAC/pull/1517 and more copy comes in from the design folks.

asteel-gsa (dismissed):
These need the same triggers from neils branch if this is going to be merged

nvm. will confirm in the other pr

neilmb:
Merging as-is to unblock some other things @eazapata23

PR #1588, opened 2023-07-21
Fix Excel file upload field error reporting in frontend

tadhg-ohiggins:
`res` already read
Thus it no longer exists
So don’t read from it.

-----

Field-specific Excel errors were showing up as `undefined` instead of passing the actual error through from the backend. This should fix that error.


jperson1 (approved):
Very true, thank you!

PR #1589, opened 2023-07-21
Error messages for late change Excel uploads

tadhg-ohiggins:
Late change rejection
Should tell you why it has failed.
This supports that case.

-----

Add meaningful error message for Excel uploads that happen after a submission should no longer be edited.


tadhg-ohiggins:
It may be worth changing this so that it's more generic and instead of creating a specific handle for the `no_late_changes` error on the frontend it looks to see if there is a `type` field in the error response and if it’s a type that’s not specifically handled the frontend code displays the contents of the `errors` field from the response, defaulting to assuming that that’s a string.

timoballard (approved):

PR #1591, opened 2023-07-21
Tweak display of Access objects in Django admin

tadhg-ohiggins:
`user` and `email`
Should always be redundant.
So just show `email`.

-----

Also display the related `SingleAuditChecklist` in the list.


danswick (approved):
Looks good to me!

PR #1593, opened 2023-07-21
Change SAC str representation

tadhg-ohiggins:
Add `report_id`
To represent submissions
Not just UEI.

-----

Changes how submissions look as strings (for example in the Django shell or in Django admin views) from e.g. `#4 - UEI(DJ4KNQ383PH7)` to  e.g. `#4--2023JUL0001000004--DJ4KNQ383PH7`.

### DIFF
--- a/backend/audit/test_models.py
+++ b/backend/audit/test_models.py
@@ -23,7 +23,7 @@ class SingleAuditChecklistTests(TestCase):
     def test_str_is_id_and_uei(self):
         """String representation of SAC instance is #{ID} - UEI({UEI})"""
         sac = baker.make(SingleAuditChecklist)
-        self.assertEqual(str(sac), f"#{sac.id} - UEI({sac.auditee_uei})")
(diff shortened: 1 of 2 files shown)


danswick (approved):
Looks good!

backend/audit/test_models.py
```diff
@@ -23,7 +23,7 @@ class SingleAuditChecklistTests(TestCase):
     def test_str_is_id_and_uei(self):
         """String representation of SAC instance is #{ID} - UEI({UEI})"""
```
danswick, line 24:
```suggestion
    def test_str_is_id_report_id_and_uei(self):
        """String representation of SAC instance is #{ID}--{report_id}--{UEI}"""
```

Should we update these as well?

danswick (commented):
Just a small suggestion

backend/audit/test_models.py
tadhg-ohiggins: Whoops, yes, I should have added that too.

tadhg-ohiggins: I'll add that in another PR.

PR #1604, opened 2023-07-24
Make addtional UEIs cross-validation check actual JSON shape of workook

tadhg-ohiggins:
The Department of
Redudancy Department
Presents: section names.

-----

Use the actual shape of the JSON that ends up in the DB after submitting an Additional UEIs workbook for validating the contents of that section.


timoballard (approved):
tested locally, looks good!

PR #1607, opened 2023-07-25
Update development.md

pluribus-tony:


jadudm (approved):
Yes.

asteel-gsa:
@mogul we have an issue here with the status checks. markdown files are excluded from `pull-request-checks.yml` but the repo settings have these. We may want to try and find an exclusion inside the repo settings?

mogul:
> We may want to try and find an exclusion inside the repo settings?

There's no way to do that there:
image

I'll just disable those checks temporarily while we figure this out.

PR #1614, opened 2023-07-26
Add .editorconfig

tadhg-ohiggins:
Define our settings
Using a common approach:
EditorConfig

-----

This just covers some pretty basic settings.


neilmb (approved):
LGTM. Do we need/want to mention in our developer documentation that we have this file now and developers should visit editorconfig.org to find out if they need a plugin for their editor?

tadhg-ohiggins:
@neilmb good point, I added that note.

Issue #1617, opened 2023-07-26
Ensure the passthrough entity ID column is formatted as text to prevent any federal awards upload failures when users input only numerical values in these cells.


sambodeme:

sambodeme:
Resolved through #1618

PR #1618, opened 2023-07-26
Sambodeme/format passthrough entity id column as text

sambodeme:


sambodeme:
This PR closes #1617

neilmb (approved):
Looks very reasonable. 

No blocker for merging, but I wonder how you noticed this issue and if there's a way that we could add or include a test case somewhere that could verify that this has been fixed and will stay fixed? I know that we have talked about various styles of end-to-end testing, but in my dream world we would have a way to make test cases for workbook bugs/changes.

sambodeme:
> Looks very reasonable.
> 
> No blocker for merging, but I wonder how you noticed this issue and if there's a way that we could add or include a test case somewhere that could verify that this has been fixed and will stay fixed? I know that we have talked about various styles of end-to-end testing, but in my dream world we would have a way to make test cases for workbook bugs/changes.

`How I noticed` : I copied data from [this federal awards](https://drive.google.com/drive/u/2/folders/1IfGPigUKYIsd7tIsyowHFo7756hjk3-B) into our template and tried to upload it. 
During the upload, several issues were identified, most of which have been addressed through #1471. The remaining issues were related to missing values in our lookup lists (ALN and cluster name).So, I substituted the missing values with existing ones (as a workaround), enabling me to successfully load the aforementioned federal awards data. 
The spike for populating workbooks from csv/sqlite data files should allow us to catch these kind of issues.

PR #1621, opened 2023-07-27
Give value to is_public

pluribus-tony:


backend/audit/etl.py
```diff
@@ -271,14 +272,14 @@ def load_general(self):
...
-            entity_type=general_information["user_provided_organization_type"],
+            entity_type=entity_type,
             number_months=None,  # TODO: Where does this come from?
             audit_period_covered=general_information["audit_period_covered"],
             report_required=None,  # TODO: Notes say this hasn't been used since 2008.
             total_fed_expenditures=None,  # TODO: Where does this come from?
             type_report_major_program=None,  # TODO: Where does this come from?
             type_audit_code="UG",
-            is_public=None,  # Should be coming from SingleAuditChecklist
+            is_public=entity_type != "tribal",
```
JeanMarie-TTS, line 282: We should implement the logic in audit/models.py and call the method in etl.py.
The implementation is likely to evolve over time, and we don't want this in etl.py

pluribus-tony: Updated and pushed new commit.

JeanMarie-TTS (approved):

PR #1645, opened 2023-07-28
Update to ZAP Action

asteel-gsa:
Change the docker container implementation to the ZAP Action.


asteel-gsa:
Backing this PR out to test the report into sarif

Issue #1473, opened 2023-07-11
Add coversheet to all workbooks


danswick:
```[tasklist]
### Workbooks
- [x] Federal awards
- [x] Audit findings
- [x] Corrective action plan
- [x] Audit findings text
- [x] Secondary auditors
- [x] Notes to SEFA
- [x] Additional UEIs
```

The coversheet is the first tab the user sees. It should include 

1. Auditee UEI
2. Any auto-computed fields that are single cells (varies by workbook)
3. `version`

The sheet should be called "Coversheet." 

It might have a look roughly like:

| Federal Audit Clearinghouse | fac.gov | Version X.Y.Z |
|---|---|---|
| Auditee UEI: | ... | |
| Total Expenditures: | 123,456 | | 

where perhaps the header row is blue. 

However, these could also all be cells with no header row.

| | |
|---|---|
| Federal Audit Clearinghouse | fac.gov | 
| Version | 1.2.3 |
| Auditee UEI | ... |
| Total Expenditures | ... |

Simple is good. 

We do not *yet* know how we will version the workbooks. this work should set all of them to `1.0.0`. 

Note a leading `v` is not technically valid semver.

danswick:
Testing this out in [a branch](https://github.com/GSA-TTS/FAC/tree/ds-auditors-coversheet) and thinking it might be worth adding a new cells/range type to handle both text-only content (title, website, version) and vertically-oriented input cells. We could merge these heading cells, but 1) I think merged cells might be more trouble than they're worth and 2) the resulting sheet could still be awkward and confusing. Open to other ideas using the current parse/render workflow for generating workbooks. 

<img width="541" alt="image" src="https://github.com/GSA-TTS/FAC/assets/2365503/ef139e29-f67b-4ce2-86d2-840f58da7653">

jadudm:
1. Agree we want to avoid merges at this point.
2. The option-two visualization above might keep things simpler with side-by-sides. :shrug:

Certainly, for the data inputs, we already have the `single_cell`, which might make things easier?

Issue #1472, opened 2023-07-11
Add metadata to all workbooks


danswick:
### Requirements:
* As locked cells in cover sheet.
* Values to include:

| Cell title | Named range | Validation | Notes |
| --- | --- | --- | --- |
| Version number | `version_number` | semver | A single version number referenced by all jsonnet templates. [Discussion](https://github.com/GSA-TTS/FAC/issues/1314) |
| Section name | `section_name` | string | Will be added to the template from [fixtures](https://github.com/GSA-TTS/FAC/blob/main/backend/audit/fixtures/excel.py#L61). |

Both values should be defined once and referenced by both the template and the schema.

```[tasklist]
### Workbook templates
- [x] Federal awards
- [x] Audit findings
- [x] Corrective action plan
- [x] Audit findings text
- [x] Secondary auditors
- [x] Notes to SEFA
- [x] Additional UEIs
```

```[tasklist]
### Schemas
- [x] Federal awards
- [x] Audit findings
- [x] Corrective action plan
- [x] Audit findings text
- [x] Secondary auditors
- [x] Notes to SEFA
- [x] Additional UEIs
```

```[tasklist]
### Tests
- [x] Federal awards
- [x] Audit findings
- [x] Corrective action plan
- [x] Audit findings text
- [x] Secondary auditors
- [x] Notes to SEFA
- [x] Additional UEIs
```

@jadudm and @lauraherring to verify canonical list of workbooks.

PR #1647, opened 2023-07-28
Add metadata, brief instructions, and UEI entry all workbooks' coversheets

danswick:
This should be ready for review. I dropped the rendered workbooks into a [Google Drive folder](https://drive.google.com/drive/folders/1Un_vaJp9R05I8pblk3uRdQPu_HvO6ifi?usp=sharing) (current as of 8/2 / 699f6357d82571fb4c9ff52c56a66df4eb373511) if that's helpful during review.

Ref: https://github.com/GSA-TTS/FAC/issues/1473
Ref: https://github.com/GSA-TTS/FAC/issues/1472

```[tasklist]
### Tasks
- [x] Section name
- [x] Updated tests
- [x] Formatting tweaks
- [x] Clean up stray comments
- [x] ~~Possibly refactor the `meta-cell` idea~~
- [x] Possibly some missing entry fields on coversheet
- [x] Manually review workbooks for completeness and regressions
```


danswick:
Made some more updates here (comments inline for more details). The format is still...not ideal. I think it should be clearer which fields are editable and which are pre-filled, but I don't see an obvious way to do that without adding a bunch of conditionals to `single_cell` or creating a new cell type. 

I do like the idea of including some narrative/helper text in the coversheet. My first attempt at a format is kind of weird, but maybe we can iterate on it.

image

backend/schemas/scripts/parse.py
```diff
@@ -86,6 +88,24 @@ def parse_single_cell(spec):
     )
 
 
+# Meta cells probably don't need named ranges
```
danswick, line 91: Maybe these could behave more like a `single_cell` but with an optional range.

backend/schemas/source/excel/templates/additional-ueis-template.jsonnet
```diff
@@ -3,18 +3,58 @@ local Help = import '../libs/Help.libsonnet';
...
+local meta_cells = [
+  Sheets.meta_cell {
+    keep_locked: true,
+    title: 'Federal Audit Clearinghouse\nfac.gov',
+    title_cell: 'A1',
+    help: Help.unknown,
+  },
+  Sheets.meta_cell {
+    keep_locked: true,
+    width: 48,
```
danswick, line 18: I believe width will be determined by single cells in the sheet.

backend/schemas/source/excel/libs/Sheets.libsonnet
```diff
@@ -23,12 +30,30 @@ local y_or_n_range = open_range {
   range_name: 'Example YorN range',
 };
 
+// TODO: import from audits.fixtures.excel
```
danswick, line 33: The "schemas" directory is set up a bit like its own project, which makes importing from fixtures a little complicated. I tested out a couple of options, but none seemed great so I just copypasta'd for now.

.editorconfig
jadudm, line 1: I suspect your editor configuration should not be part of the repo?

backend/schemas/scripts/parse.py
```diff
@@ -141,14 +161,17 @@ def parse_text_range(spec):
     )
 
 
-def parse_sheet(spec):
-    sc, opr, mc, mur, hi, tr = None, None, None, None, None, None
+def parse_sheet(spec):  # noqa: C901
```
danswick, line 164: the new spec pushed this function over the "complexity" threshold.

.editorconfig
danswick: I'm not sure how it got in there or how it was removed, but it's no longer there 😵‍💫

backend/audit/test_validators.py
```diff
@@ -659,7 +659,14 @@ def test_validation_is_applied(self):
         template = json.loads(template_definition_path.read_text(encoding="utf-8"))
         invalid = json.loads('{"CorrectiveActionPlan":{}}')
         expected_msg = str(
-            ("A", "2", "Auditee UEI", template["sheets"][1]["single_cells"][0]["help"])
+            [
```
danswick, line 662: I don't understand why this happened, but these validator errors started coming back as a list (which is what I would expect given the `list()` call here: https://github.com/GSA-TTS/FAC/blob/main/backend/audit/validators.py#L156). So maybe I'm not sure why they _weren't_ lists before and which part of this set of commits changed it.

jadudm: We'll pick this up *real soon now* when we start doing more E2E testing with lots of real data.

backend/cypress/fixtures/additional-ueis-UPDATE.xlsx
jadudm, line 1: Perhaps pull this?

jadudm: Not sure if this is part of the `make` process.

backend/schemas/source/base/ClusterNames.json
jadudm, line 1: We need to modify `scripts/generate_lookup_schemas.py` to include a UNIX line ending:

https://stackoverflow.com/questions/41842875/how-to-explicitly-set-carriage-return-when-doing-json-dump

We have a Windows colleague who might be forcing line ending changes on every commit.

backend/schemas/source/base/FederalProgramNames.json
jadudm, line 1: Previous comment will fix this also.

jadudm (changes requested):
One or two small changes, mostly for our future health. 

Do the line-ending thing, and then merge. (Plus any other minor things we found.)

backend/schemas/scripts/generate_lookup_schemas.py
```diff
@@ -70,5 +70,5 @@ def process_cluster_names(arg):
             print("Unknown filename, exiting")
             sys.exit(1)
 
-        with open(sys.argv[2], "w") as write_file:
+        with open(sys.argv[2], "w", newline="\n") as write_file:
```
danswick, line 73: Windows and unix have different philosophies about newlines. This can cause weird diffs that don't appear to show any actual differences between commits. Let's...commit...to one newline type.

backend/schemas/source/base/ClusterNames.json
danswick: https://github.com/GSA-TTS/FAC/pull/1647/commits/8ae1e1ff6c25c97271b1316d4eea82a2be3ac62e

backend/schemas/source/excel/templates/additional-ueis-workbook.jsonnet
```diff
@@ -35,18 +75,19 @@ local open_ranges_defns = [
...
-    name: additionalUeiSheet,
-    open_ranges: Fun.make_open_ranges_with_column(title_row, open_ranges_defns),
-    header_height: 48,
-    hide_col_from: 2,
+    name: coverSheet,
+    meta_cells: meta_cells,
+    single_cells: single_cells,
+    header_height: 60,
+    hide_col_from: 3,
+    //FIXME MSHD: commented this out until we figure out if it is needed
```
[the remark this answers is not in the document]
danswick: https://github.com/GSA-TTS/FAC/pull/1647/commits/b3a16c9d87c51ef463b9fc7062793025ad7e706a

jadudm (approved):
LGTM.

PR #1651, opened 2023-07-31
Add Schedule and environment input

asteel-gsa:
Fixes the schedule to allow inputs, without removing core functionality.
Adds `environment` input for future

Schedule does not natively support inputs, only `workflow_call` and `workflow_dispatch` do. The schedule saw `${{ inputs.url }}` as a `null` value.


neilmb (approved):
Awesomely quick fix.

PR #1652, opened 2023-07-31
Workaround for SAM.gov in OpenSSL v3

neilmb:
There is a backend error where the application can't contact SAM.gov because of an SSL error: `SSLError(1, '[SSL: UNSAFE_LEGACY_RENEGOTIATION_DISABLED] unsafe legacy renegotiation disabled (_ssl.c:1007)')`. Following this Stack Overflow https://stackoverflow.com/questions/71603314/ssl-error-unsafe-legacy-renegotiation-disabled, this implements a python-level solution by specifying a specific SSL context that allows this ONLY when we are calling SAM.gov.

@danswick and @jadudm worked out an OS-level solution so they could take a look at this, and @mogul should probably notice that we are making the possibly insecure accommodation for a limitation of SAM.gov.

To test this, visit http://localhost:8000/api/sac/ueivalidation and POST a payload of `{"auditee_uei": "nonemptystring"}` and look for a successful `"UEI was not found in [SAM.gov](http://sam.gov/)"` error.

### DIFF
--- a/backend/api/test_serializers.py
+++ b/backend/api/test_serializers.py
@@ -71,7 +71,7 @@ class UEIValidatorStepTests(SimpleTestCase):
         valid = {"auditee_uei": "ZQGGHJH74DW7"}
 
         # Valid
-        with patch("api.uei.requests.get") as mock_get:
+        with patch("api.uei.SESSION.get") as mock_get:
             mock_get.return_value.status_code = 200  # Mock the status code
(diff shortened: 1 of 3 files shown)


jadudm (approved):
Merging for purposes of test. We aren't confident we like this solution, but we know it works locally, and want to see what it does elsewhere.

Issue #1495, opened 2023-07-13
Revise certification workflow


jadudm:
# At a glance

[comment]: # "Begin with a short summary so intent can be understood at a glance."

[comment]: # "In order to: some objective or value to be achieved"
[comment]: # "as a: stakeholder"
[comment]: # "I want: some new feature"

**In order to** finalize my audit
**as an** auditor/auditee
**I want** a certification workflow.

# Acceptance Criteria

We use [DRY](https://docs.behat.org/en/latest/user_guide/writing_scenarios.html#backgrounds) [behavior-driven development](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications) wherever possible.

**Given**   a completed audit
**when**  I get to certification
...

[comment]: # "Each task should be a verifiable outcome"
```[tasklist]
### then...
- [x] Redesign the begin certification as a lock for certification page
- [x] Mock up the certification checklist in a "locked" state.
- [x] Develop content for the lock for certification page
- [x] Implement front-end "lock for certification" page
- [x] Update submission checklist to conditionally allow submission based on lock state
- [x] Update upload validation pathways (XLSX, PDF) to check for valid submission state
- [ ] https://github.com/GSA-TTS/FAC/issues/1516
- [x] Implement auditor certification workflow / page
- [x] Implement the auditee certification workflow / page
- [x] Implement final submission page
```

### Shepherd

[comment]: # "@ mention shepherds as we move across the board."

* Design shepherd: @lauraherring 
* Engineering shepherd: @heymatthenry 

# Background

[comment]: # "Any helpful contextual notes or links to artifacts/evidence, if needed"

We're now going to add a notion of forward-moving immutability to the certification process.

1. If this is a tribal submission, they must complete the public content attestation before certification locking. 
1. The "Ready to certify" page needs to be updated.
    1. We think we only want the Auditee Certifying Official to actually be able to push this button. It might be only they can see this link? (hm.)
    2. The user should be warned that locking for certification cannot be undone. They should make sure they have the final and correct versions of the workbooks and PDF in place.
    3. The button text should be clear that it is a locking action.
    4. Do we want a modal "Are you sure?" Possibly.
    5. Only the appropriate user(s) can hit the button successfully (in case someone shares a link to the page with a non-approved user)
    2. When the user "locks," we will set state to `READY_FOR_CERTIFICATION`. 
[...]

PR #1680, opened 2023-08-02
Revised Auditee & Auditor Certification Workflow

jperson1:
# Auditee & Auditor Certification Workflow

Issue: #1495 
Figma: [Auditor Cert Link](https://www.figma.com/file/lnkXJTfKzD3YBhiJZ2QFEv/FAC-final?node-id=1%3A20991&mode=dev)

## Changes:
1. The auditee and auditor certification HTML has been updated to match with the Figma docs. 
    * This includes a two-step certification form, for the checklist and signature
2. Auditee/auditor certification stores the certification objects to the SAC DB object.
    * One migration adds both fields to the DB. 
    * They have their own json schema files.
    * They are of form:
    ```
    auditor/ee_certification: {
        "auditor/ee_certification": {
            "checklist_item": True,
            "checklist_item": True,
            ...
        }
        "auditor/ee_signature:{
            "auditor/ee_name": "A Name",
            "auditor/ee_title": "A Title",
        }
    }
    ```
4. While the end data is saved to the DB, it does not autofill on the page. Autofill is handled by the session variable that the forms are stored in to bridge the gap between pages. This means that they _will_ autofill if a user reloads the page, but will _not_ autofill if the user closes the page and comes back - or does some other kind of hard-reset. Since this process is (a) short and (b) only meant to be done a single time, I think this is fine. 
5. The submission checklist _does_ update as the process is completed. But, it matters little, since it's getting a big update soon.

## How to test:
1. Switch to this branch and run normally. Make sure you apply the new migrations. You can also do a clean build with `make docker-clean` and `make docker-first-run`. 
2. Create an audit, or navigate to an existing one. 
3. Get your audit to the state that it can access `/audit/auditor-certification/{reportId}`. This can be done by completing the SF-SAC, or by manually setting the `submission_status` to `ready_for_certification` in the DB. During development, I cheated it by moving the dispatch up so that the role validation didn't kick me out. Good for local dev, bad for prod. 
4. Make sure the page works, and that the proper relics are loaded into the DB.

## Screenshots:
image
image
image
image


tadhg-ohiggins:
I noticed that the existing code requires some workbooks that are optional, namely additional UEIs and auditors, so I'm moving to draft until I work through those.

tadhg-ohiggins:
I changed this to be far more permissive—almost none of the possibly-optional workbooks are required, for example. We definitely need to tweak this, but one advantage of going this way is that it's easier to test the specific functionality we're trying to add with this PR, i.e. enabling certification.

I think folks should check out this branch and try it locally to see what seems most egregiously wrong with the approach I've taken.

tadhg-ohiggins:
@jadudm tagging you because I think this is related enough to the other PR we looked at.

backend/api/views.py
```diff
@@ -118,20 +121,33 @@ def access_and_submission_check(user, data):
 
         # Create all contact Access objects
         Access.objects.create(sac=sac, role="editor", email=user.email, user=user)
-        Access.objects.create(
+        cert_auditee_access = Access.objects.create(
```
tadhg-ohiggins, line 124: @timoballard I added these here because in trying to test this PR I ran into an issue that we discussed way back: if someone is logged into the system when they're added to a submission, they won't be able to access that submission until they log out and in again. Leaving that in place makes testing this a lot harder, since it would mean that you'd have to log out & in between steps (or have multiple users involved).

Let me know if you think this is reasonable enough to leave as part of this PR or if you think we should make a separate PR for it.

tadhg-ohiggins: Also I just learned that if you select a bunch of lines and then add a comment, it just defaults to the first line… most of what I'm referring to here is below, in the `to_reify` loop.

timoballard: removed as this was implemented separately in #1778

jadudm (approved):
LGTM.

jadudm (approved):
LGTM. Still.

jperson1:
Merge conflicts resolved. Needs a run-through to ensure all checklist elements display correctly (certification enables when you get to it, etc).

jperson1:
This should now be good to go. I've unrequired some sections to ensure the pre-submission validation is enabled when the mandatory uploads are complete.

PR #1708, opened 2023-08-03
Make sure requests sessions passes custom SSL context to proxy to account for SAM.gov legacy SSL version

tadhg-ohiggins:
Some complexity
Mixing ancient with proxies
But we persevere


jadudm (approved):
LGTM. 

A large team worked through these changes. We need to now test through to `staging`.

PR #1710, opened 2023-08-03
Add ready-for-cert cy+edited fed awards+workbook

eazapata23:
I added a test for ready for certification once workbooks have been uploaded. I want to add this to help accelerate e2e full submission + I've made some change on federal awards workbook test that has a passing federal award workbook.


backend/cypress/e2e/ready-for-cert.cy.js
jadudm, line 1: Going to merge this through; wondering if there page wants an assertion for where it will go after the click. However, it might be that we don't know/don't have that page yet, hence why no assertion.

jadudm (approved):
LGTM and Dan and Tadhg.

PR #1714, opened 2023-08-03
Make `proxy_manager_for` specify the proxy

tadhg-ohiggins:
Specify proxy
In our custom connection
Or else: no egress

-----

We have to tell our custom method which proxy to use.


jadudm (approved):
Lets find out. LGTM.

PR #1716, opened 2023-08-03
Actually get the CSRF token we need to send the UEI check to our own endpoint

tadhg-ohiggins:
Wild geese, red herrings
And, of course, cookie settings.
Will it work this time?

-----

The last set of SSL-related changes were incorrect; the regression test failures—this time—were likely caused by a long-standing cookie/CSRF issue preventing the form page from talking to another page in our app via XHR.

### DIFF
--- a/backend/static/js/csrft.js
+++ b/backend/static/js/csrft.js
@@ -1,15 +1,3 @@
 export const getCookie = (name) => {
-  let cookieValue = null;
-  if (document.cookie && document.cookie !== '') {
-    const cookies = document.cookie.split(';');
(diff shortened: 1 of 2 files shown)


danswick (approved):
Please accept this... _token_ of approval

backend/static/js/csrft.js
```diff
@@ -1,15 +1,3 @@
 export const getCookie = (name) => {
```
danswick, line 1:
```suggestion
export const getCookie = () => {
```

PR #1722, opened 2023-08-04
Fixing the cluster name validation

jadudm:
These changes fix our cluster name validation.

1. A valid cluster name is either: 1. N/A 2. One of the cluster names 3. STATE CLUSTER 4. OTHER CLUSTER NOT DEFINED
2. IF a user selects N/A or a cluster name, then both the state cluster and other cluster fields MUST be empty.
3. IF a user selects STATE CLUSTER, then the state cluster name must be completed, and the OTHER field must be empty.
4. IF a user selects OTHER CLUSTER, then the state cluster name must be empty, and the OTHER field must be completed.

Our previous validation was more confusing, and we may not have had all of the choices present.

This means we can use a rigid validation in our workbooks for state cluster. The user has the ability to pick OTHER and enter data in the appropriate fields. We should NOT allow them to edit the cluster name field in our workbooks.

This validation should also prevent submission if they do, or manage to.


backend/schemas/source/base/Base.libsonnet
```diff
@@ -383,15 +383,15 @@ local type_uei = Types.string {
...
-    },
+    }
   ],
 };
 
 local Compound = {
   AwardReference: Types.string {
     title: 'AwardReference',
     description: 'Award Reference',
-    pattern: '^AWARD-(?!0000)[0-9]{4}$',
```
sambodeme, line 394: These commas are automatically added by the JSONnet formatting tool when you run `jsonnetfmt -i <pathToYourFile>`.

backend/schemas/source/base/Base.libsonnet
```diff
@@ -466,8 +465,17 @@ local SchemaBase = Types.object {
     // },
     ClusterNames: {
       description: 'All cluster names',
-      enum: ClusterNames.cluster_names + [Const.NA, Const.OTHER_CLUSTER]
-    }
+      enum: ClusterNames.cluster_names,
+    },
+    ClusterNamesNAStateOther: {
+      description: 'All cluster names',
+      enum: ClusterNames.cluster_names + [Const.NA, Const.STATE_CLUSTER, Const.OTHER_CLUSTER],
```
sambodeme, line 472:
```suggestion
      enum: ClusterNames.cluster_names + [Const.NA, Const.OTHER_CLUSTER],
```

sambodeme: Const.STATE_CLUSTER is already included in the CSV

backend/schemas/source/base/Base.libsonnet
```diff
@@ -466,8 +465,17 @@ local SchemaBase = Types.object {
...
+    },
+    ClusterNamesStateOther: {
+      description: 'All cluster names',
+      enum: ClusterNames.cluster_names + [Const.STATE_CLUSTER, Const.OTHER_CLUSTER],
```
sambodeme, line 476:
```suggestion
      enum: ClusterNames.cluster_names + [Const.OTHER_CLUSTER],
```

sambodeme (approved):

backend/schemas/source/base/Base.libsonnet
```diff
@@ -383,15 +383,15 @@ local type_uei = Types.string {
...
+    }
   ],
 };
 
 local Compound = {
   AwardReference: Types.string {
     title: 'AwardReference',
     description: 'Award Reference',
-    pattern: '^AWARD-(?!0000)[0-9]{4}$',
+    pattern: '^AWARD-(?!0000)[0-9]{4}$'
```
jadudm, line 394:
```suggestion
    pattern: '^AWARD-(?!0000)[0-9]{4}$',
```

backend/schemas/source/base/Base.libsonnet
```diff
@@ -424,8 +424,7 @@ local Compound = {
   UnitedStatesPhone: Types.string {
     pattern: phone_regex,
   },
-  Zip: type_zipcode,
-
+  Zip: type_zipcode
```
jadudm, line 427:
```suggestion
  Zip: type_zipcode,
```

PR #1723, opened 2023-08-04
Adding tools for workbook generation and testing

jadudm:
This brings two tools into the tree that have been used for exploring workbook generation.

`generate-sqlite-files` takes public, pipe-delimited Census data and turns it into an SQLite3 database. This improves on previous tools that did this in the past.

`workbook-generator` is a script that takes a DBKEY and an SQLite3 database containing public Census data, and outputs a set of GFAC-style XLSX workbooks containing data.

To do this, `workbook-generator`

1. Loads one of our templates using `openpyxl`
2. Loads data from the SQLite3 database into the named ranges in the workbook template
3. Saves the template out to a new filename, populated with data.

The generator attempts to do everything right. For example, it attempts to follow old linking IDs (e.g. ELECAUDITID) and replace them with AWARD-####-style references. Similarly, it unpacks the odd design of the Notes to SEFA tables to convert the Census data back into a functional workbook.

This is not a fully validated tool. However, it does generate workbooks with authentic data, and they have been used to drive upload processes into our system. Using them, we have found errors in our validations that we did not know about previously.

If we continue using this tool, we will likely want to build it into our testing automation process. In theory, we could have hundreds (or thousands) of workbooks generated and ready for testing. We can also generate workbooks that explicitly test (e.g.) having secondary auditors (or not), or that exhibit specific other properties... all from existing, previously-validated data.

The generator also spits out a JSON document. That document records:

1. The table data was pulled from
2. The fields pulled
3. The values in those fields

The purpose of that document is to be able to do something like:

1. Load the workbooks through our pipeline
2. Have them undergo validation, cross-val, and ETL
3. Use the JSON document to compare what we pulled from the SQLite DB to what ended up in our dissemination DB.

In other words, the JSON document is to enable end-to-end testing of the dissemination pipeline. Ideally, we would do that final check using the API. This would let us use the JSON document to generate API calls that query the DB (from the "outside"), and verify that the API produces data we expect.


tools/generate-sqlite-files/.gitignore
```diff
@@ -0,0 +1,3 @@
+allfac*
+*.zip
+*.sqlite3
```
asteel-gsa, line 3:
```suggestion
*.sqlite3

```
Can you add a newline at the eof, just to remove the `-` at the end?

tools/workbook-generator/.gitignore
```diff
@@ -0,0 +1,5 @@
+__pycache__/**
+data/ay22/*.sqlite3
+output/**
+templates/*.xlsx
+.~lock*
```
asteel-gsa, line 5:
```suggestion
.~lock*

```
Same thing here, new line at eof to remove the `-`

tools/generate-sqlite-files/main.py
```diff
@@ -0,0 +1,44 @@
...
+
+def main(args):
+    # Name the sqlite the same as the zipfile.
+    sqlitefilename = f'{Path(args.zip).stem}.sqlite3'
+    print(f'Writing to {sqlitefilename}')
+    if os.path.isfile(sqlitefilename):
+        os.remove(sqlitefilename)
+
+    # Do everything in a temp dir.
+    # It will disappear when we hit the end of the with block.
```
asteel-gsa, line 25:
```suggestion
    # Do everything in a temp dir.
    # It will disappear when we hit the end of the with block.
```
This is a non blocker, but what I would personally like, is to have comments removed from file, and put in the `tools/generate-sqlite-files/readme.md` as code snippets with these comments. 

so in `readme.md`
`with tempfile.TemporaryDirectory('_fac') as tdir:` will create a temp dir for running this.. etc etc.

tools/workbook-generator/data/.gitignore
```diff
@@ -0,0 +1 @@
+*/**journal
```
asteel-gsa, line 1:
```suggestion
*/**journal

```
new line at eof to remove `-`

tools/workbook-generator/historic_workbooks
```diff
@@ -0,0 +1,15 @@
...
+    "101106"
+    "99989"
+    "175887"
+    )
+
+for key in "${keys[@]}"
+do
+   echo "$key"
+   python main.py --output ../../backend/data_fixtures/historic/ --dbkey $key
+done
```
asteel-gsa, line 15:
```suggestion
done

```
new line at eof to remove `-`

tools/generate-sqlite-files/main.py
jadudm: Lets pick this up in an iteration. I have a bunch of questions, actually, as to how we would even work some of this into a GH workflow. That would necessarily drive changes to this. For example, we might want to make it a Django command, so it is part of the FAC app. 

The number of changes coming for this script could be many. Or, we might dump it. So, at that point, we have opportunities to iterate/reshape the whole thing. Which we might have to.

asteel-gsa:
The move of comments into readme is very much a personal thing, it is not at all a necessity and can be ignored. I like the idea of dropping stackoverflow links for references, and its really up to you if you feel that those comments can be documented in a readme vs inside code. Who you ask will tell you yes/no, so i think what im trying to say, if you feel those are worth expanding upon in a readme, great, if not, leave them in if you feel they are necessary to the code. 

Otherwise, LGTM @jadudm

PR #1724, opened 2023-08-04
Jadudm/update dev requirements

jadudm:
I do not fully understand the difference between the `requirements.in` and `dev-requirements.in` files. 

But, I seem to often be missing a library that is needed. I don't know why they're different, and how it happens that we have libraries in one place but not another. 

I think I need to add these libraries to dev. :shrug:


jadudm:
Nevermind. Or, I did something wrong. I think I managed to paste something bad into an `in` file.

PR #1736, opened 2023-08-04
Jadudm/pdf to model

jadudm:
I would love eyes on this.

I believe this is a clean merge of code to bring PDF submissions all the way in.

However, I had a rough day in terms of making sense of branches.

So:

1. This should have one migration.
2. It should add the component pages to the data model
3. It should add the report to the admin view

This should allow PDF uploads to happen fully.


backend/audit/fixtures/__init__.py
```diff
@@ -1,9 +1,9 @@
-from .single_audit_checklist import (
-    load_single_audit_checklists,
-    load_single_audit_checklists_for_email_address,
-)
+# from .single_audit_checklist import (
```
tadhg-ohiggins, line 1: Was removing all of this intentional?

backend/schemas/source/base/Base.libsonnet
tadhg-ohiggins, line 1: `jsonnetfmt` seems to prefer trailing commas. Was removing these intentional?

jadudm: Probably not. I'll fix.

backend/audit/fixtures/__init__.py
jadudm: I'll double check that, but I think that came from a circular import, and the fix was suggested by Tim. However, it's worth me checking.

backend/audit/test_schemas.py
```diff
@@ -524,13 +524,12 @@ class FederalAwardsSchemaValidityTest(SimpleTestCase):
...
 
-    def test_schema(self):
+    def test_multiple_files(self):
         """Try to test FederalAwards first."""
         schema = self.FEDERAL_AWARDS_SCHEMA
-
-        in_flight_file = FEDERAL_AWARDS_TEST_FILE
-        in_flight = json.loads(in_flight_file.read_text(encoding="utf-8"))
-        validate(in_flight, schema)
+        for in_flight_file in FEDERAL_AWARDS_TEST_FILE:
```
tadhg-ohiggins, line 530: Isn't `FEDERAL_AWARDS_TEST_FILE` a JSON file (and therefore not something you can iterate over)?

backend/schemas/source/base/Base.libsonnet
```diff
@@ -386,7 +386,7 @@ local SchemaBase = Types.object {
   Compound: Compound {
     FederalProgramNames: {
       description: 'All Federal program names',
-      enum: FederalProgramNames.program_names,
+      enum: FederalProgramNames.program_names
```
jadudm, line 389:
```suggestion
      enum: FederalProgramNames.program_names,
```

backend/schemas/source/base/Base.libsonnet
```diff
@@ -334,15 +334,15 @@ local type_uei = Types.string {
     // Does not start with 9 digits in a row
     {
       pattern: '^(?![0-9]{9})',
-    },
+    }
```
jadudm, line 337:
```suggestion
    },
```

backend/schemas/source/base/Base.libsonnet
```diff
@@ -342,7 +342,7 @@ local Compound = {
   AwardReference: Types.string {
     title: 'AwardReference',
     description: 'Award Reference',
-    pattern: '^AWARD-(?!0000)[0-9]{4}$',
+    pattern: '^AWARD-(?!0000)[0-9]{4}$'
```
jadudm, line 345:
```suggestion
    pattern: '^AWARD-(?!0000)[0-9]{4}$',
```

backend/audit/views.py
```diff
@@ -439,6 +439,9 @@ def get(self, request, *args, **kwargs):
 
         try:
             sac = SingleAuditChecklist.objects.get(report_id=report_id)
+            audit_report = SingleAuditReportFile.objects.filter(sac_id=sac.id).latest(
```
tadhg-ohiggins, line 442: This will fail if there's no matching `SingleAuditReportFile`—which will be the case for any submissions where someone hasn't uploaded one yet. I think this probably needs its own try/catch block, although maybe there's a nicer way to do it.

tadhg-ohiggins (approved):

PR #1737, opened 2023-08-04
Make tests match the actual shape of the data, change validation functions to handle that shape

tadhg-ohiggins:
Brought to you by the
Redundancy Department:
Reshaping JSON.

-----

Our nascent cross-validation functions weren’t taking into account the fact that some of the sections (the workbooks) have additional top-level name fields containing the dictionary of their values. This fixes that and should give us a single place to track future differences like this.

Also add tests for the `additional_ueis` validation.


jadudm (approved):
Looks good. Thank you.

Issue #1738, opened 2023-08-04
Use 2019 baseline data in Cognizant assignment function


gsa-suk:
Breaking this out from its parent epic. See https://github.com/GSA-TTS/FAC/issues/1218 for more discussion and history.

This covers the implementation for cognizant assignment using 2019 baseline data.

PR #1739, opened 2023-08-04
Sk/cog over

gsa-suk:
Epic 1218 - Completed 1462 and 1699.


backend/audit/cog_agency.py
```diff
@@ -0,0 +1,93 @@
...
+    #   print("tot_amount_agency = ", tot_amount_agency)
+
+    tot_da_amount_agency = list(
+        sorted(tot_da_amount_agency.items(), reverse=True, key=lambda item: item[1])
+    )
+    #   print("tot_da_amount_agency = ", tot_da_amount_agency)
+
+    #   print("tot_da_amount_expended = ", tot_da_amount_expended)
+
+    if tot_amount_expended > 50000000:
```
jadudm, line 70: Lets remove all of the constants from this file, and put them out in `settings`. If these numbers ever change, I don't want to have to find them all over the codebase.

For now, it would be enough to have a set of constants at the top of the file in a dictionary. 

```
constants = {
   "amount_expended_limit" : 50000000,
   ...
}
```

or similar. We can pull them out of this file later.

I'd also like to suggest adding a few constants like

```
THOUSAND = 1000
HUNDRED_THOUSAND = 100000
MILLION = 1000000
```

so that we can write

```
constants = {
   "amount_expended_limit" : 50 * MILLION,
   ...
}
```

which makes it much easier to glance and see what the numeric constant is.

backend/audit/test_cog_agency.py
jadudm, line 1: This is a good start. However, we're going to want to turn this into a proper set of test cases.

If you look at other files called `test_...`, you'll see that they implement test classes derived from `TestCase` (or... similar... I forget at the moment). Within those, a series of test methods are written, each with the name `test_...`. The naming is important; the test module looks for that test class (derived from a base test class), and then looks for methods with names beginning with `test`. 

Finally, those methods should then `assert` what is expected. If they pass, they get counted up when we do a `make docker-test`. If they fail, we see it.

You can find more info at 

https://docs.djangoproject.com/en/4.2/topics/testing/overview/

and, ultimately, it's just the Python `unittest` library that is being used under the hood.

backend/audit/test_cog_over_assignment.py
```diff
@@ -0,0 +1,82 @@
...
+                        "direct_or_indirect_award": {"is_direct": "Y"},
+                    },
+                ],
+                "total_amount_expended": 51_000_000,
+            }
+        }
+
+    def test_cog_over_assignment(self):
+        cog_agency, over_agency = cog_over_assignment(self.federal_awards_for_test)
+        print("cognizant agency =  ", cog_agency)
```
JeanMarie-TTS, line 81: Use assert statements to make sure your tests are passing. Please remove print statements before you finalize the PR.

backend/audit/cog_agency.py
```diff
@@ -0,0 +1,97 @@
...
+#       else
+#           over_agency = agency with max tot_amt in tot_amount_agency
+#############################################################
+
+from collections import defaultdict
+
+
+def cog_over_assignment(federal_awards_data):
+    MILLION = 1_000_000
+    constants = {"amount_expended_limit": 50 * MILLION}
```
JeanMarie-TTS, line 40: I suggest using upper case for constants - AMOIUNT_EXPENDED_LIMIT

backend/audit/cog_agency.py
```diff
@@ -0,0 +1,97 @@
...
+    cog_agency = 0
+    over_agency = 0
+    tot_amount_agency = defaultdict(lambda: 0)
```
JeanMarie-TTS, line 43: Consider factoring out the compiutation of amounts into another method. The smaller the methods/functions the easier it si to read and maintain the code.

backend/audit/cog_agency.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def calc_amount_expended_limit():
+    MILLION = 1_000_000
+    amount_expended_limit = 50 * MILLION
+    return amount_expended_limit
+
+
+def cog_over_assignment(federal_awards_data):
+    cog_agency = 0
```
JeanMarie-TTS, line 45: Should default value for agency not be None rater than 0?

backend/audit/cog_agency.py
```diff
@@ -0,0 +1,101 @@
...
+    over_agency = 0
+    tot_amount_agency = defaultdict(lambda: 0)
```
JeanMarie-TTS, line 47: I'd encourage you factor Lines 47 to 68 to a new method.

backend/audit/cog_agency.py
```diff
@@ -0,0 +1,101 @@
...
+            if (
+                tot_da_amount_expended
+                >= 0.25 * federal_awards_data["FederalAwards"]["total_amount_expended"]
+            ):
+                cog_agency, val = tot_da_amount_agency[0]
+            else:
+                cog_agency, val = tot_amount_agency[0]
+        #           print("cognizant agency = ", cog_agency)
+    else:
+        #       Oversight agency
```
JeanMarie-TTS, line 91: 91 to 100 could also go to a separate method

backend/audit/cog_agency.py
```diff
@@ -0,0 +1,101 @@
...
+    )
+    #   print("tot_da_amount_agency = ", tot_da_amount_agency)
+
+    #   print("tot_da_amount_expended = ", tot_da_amount_expended)
+
+    if (
+        federal_awards_data["FederalAwards"]["total_amount_expended"]
+        > calc_amount_expended_limit()
+    ):
+        #############################################################
```
JeanMarie-TTS, line 73: 73 to 89 can be a method just deal with cog agency.

JeanMarie-TTS (commented):
The test should definitely be modified to use assertions instead of exceptions. I have made some refactoring suggestions, but these are not show stoppers.

backend/audit/cog_agency.py
gsa-suk: This is the agency prefix, which is a two digit integer value.

backend/audit/cog_agency.py
gsa-suk: I am restructuring code as part of https://github.com/GSA-TTS/FAC/issues/1738 to be able to handle 2019 data or current data.

backend/audit/cog_agency.py
gsa-suk: I am restructuring code as part of https://github.com/GSA-TTS/FAC/issues/1738 to be able to handle 2019 data or current data.

backend/audit/cog_agency.py
gsa-suk: I am restructuring code as part of https://github.com/GSA-TTS/FAC/issues/1738 to be able to handle 2019 data or current data.

backend/audit/test_cog_over_assignment.py
```diff
@@ -0,0 +1,84 @@
...
+                            "loan_balance_at_audit_period_end": 0,
+                        },
+                        "direct_or_indirect_award": {"is_direct": "Y"},
+                    },
+                ],
+                "total_amount_expended": 51_000_000,
+            }
+        }
+
+    def test_cog_over_assignment(self):
```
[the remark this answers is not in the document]
gsa-suk: Between cognizant agency and oversight agency, one of them will always be 0.  The other will contain an agency prefix number.

PR #1746, opened 2023-08-07
Remove field limits

jadudm:
If we want to limit the size of inputs to the DB, we either:

    1. Do it at the boundaries, or
    2. We have clear error handling pathways all the way back to the user.

    We currently do not do #1, and #2 is hard, because it suggests that we
    pass errors back from deep in the application.

    We should do #1.


backend/audit/models.py
```diff
@@ -218,12 +218,10 @@ class STATUS:
         null=True,
     )
 
-    report_id = models.CharField(max_length=17, unique=True)
+    report_id = models.CharField(unique=True)
```
danswick, line 221: I _think_ `unique=True` only works with `max_length` of 255 or less. https://docs.djangoproject.com/en/4.2/ref/databases/#notes-on-specific-fields

danswick:
Since we're looking at removing a lot of `max_length`s, maybe we want to use a [TextField](https://docs.djangoproject.com/en/4.2/ref/models/fields/#textfield) instead. I _think_ this is the preferred Django way to handle text fields that don't need enforced length limits.

backend/audit/models.py
jadudm: That's under the MySQL section. This may, or may not, apply to Postgres. (It's not in the Postgres notes. But, if we were to switch DB engines, this could be A Thing.)

danswick:
I think you can override the default widget, but it might not be worth it since there's unlikely to be a real performance difference for our configuration. 

Another potential consideration: with `TextField` you can specify a `max_length` for the frontend form validation only. If we're using validation schemas to do all of our error handling and messaging back to the user, this probably isn't a benefit.

PR #1754, opened 2023-08-08
Bring up preview environment

mogul:
[The preview environment is live!](https://fac-preview.app.cloud.gov) 

@asteel-gsa and I identified tech-debt that will need to be addressed eventually... but since it's not every day that we add a whole new environment, I'm going to list it below for now. If I don't have a chance to address it in the short term, I'll put it in an issue which I expect will sit in the icebox.

---

- Things that would be addressed by having Terraform also deploy the app
  - We had to manually create a stub `gsa-fac` app so that Terraform was able to look up the app GUID and use it for configuring network policies to the proxies
  - We had to manually create the `fac-key-service` the first time so that deploy workflow could update it
  - We had to add an explicit `vars-preview.yml` file for the `manifest.yml` to refer to
- Other things we had to do:
  - We had to manually create a stub `fac-av-preview` app so that Terraform was able to look up the app GUID and use it for configuring a network policy for the https egress proxy
    - We came to the conclusion that to avoid circular dependencies, the proxy modules should not create their own network policies for clients; instead this should be hoisted up to the `env` module which already manages resources across the two related spaces.
  - We had to explicitly list the `PREVIEW` environment option in every location that Python decides that it's running in cloud.gov, rather that GHA or locally.
    - We came to the conclusion that the logic should just be to look for `VCAP_APPLICATION` as the sole signifier rather than hardcoding the environment names in the code and supplying them in the manifest!
  - We had to add the OpenID redirection target URLs to [the login.gov partner dashboard](https://idp.int.identitysandbox.gov/)
  - We had to add the `public_networks_egress` ASG to the `preview-egress` space


mogul:
I don't know what Black is complaining about [over here](https://github.com/GSA-TTS/FAC/pull/1754/files#diff-f884efe80cb587210580a6c67855feab0b02f9b4bc77e1a7329aecd05c6921b1)... Anyone have ideas?
image

backend/data_distro/management/commands/create_distro_api.py
```diff
@@ -22,7 +22,7 @@ def handle(self, *args, **kwargs):
         else:
             files = ["basic_views.sql", "findings.sql", "general.sql"]
 
-        if settings.ENVIRONMENT not in ["DEVELOPMENT", "STAGING", "PRODUCTION"]:
+        if settings.ENVIRONMENT not in ["DEVELOPMENT", "PREVIEW", "STAGING", "PRODUCTION"]:
```
tadhg-ohiggins, line 25:
```suggestion
        cloudgov = ["DEVELOPMENT", "STAGING", "PRODUCTION"]:
        if settings.ENVIRONMENT not in cloudgov:
```

.github/workflows/staging-scheduled-deploy.yml
```diff
@@ -2,7 +2,8 @@
 name: Scheduled Deploy From Main to Staging
 on:
   schedule:
-    - cron: '0 10 * * 1-5'
+    # Invoke every Mon-Sat
+    - cron: '0 10 * * 1-6'
```
asteel-gsa, line 6: Have been meaning to add this, since previously, it will not detect any merges into main after friday, this will fix that and ensure no merges are missing over the weekend

.github/workflows/terraform-apply-env.yml
```diff
@@ -26,6 +26,8 @@ jobs:
       TERRAFORM_PRE_RUN: |
           apt-get update
           apt-get install -y zip python
+      TF_VAR_postgrest_image: 'ghcr.io/gsa-tts/fac/postgrest:latest'
```
asteel-gsa, line 29: For some reason, we never had the postgrest or clamav image explicitly defined inside `terraform-apply-env.yml`. Since `preview` will not be using any prs, and instead will have a `workflow_dispatch` called from any branch, this should be here. Plus, it should be here regardless.

jadudm (approved):
All looks good. Definitely see the point about VCAP...

PR #1781, opened 2023-08-10
Missing ampersand FTW

tadhg-ohiggins:
Trailing ampersands
Must always travel in pairs
Or they will recede.

-----

A lone trailing ampersand here may be causing these database commands to be run in incorrect order.


jadudm (approved):
Ampersands alone
Whispers in the void that binds
Pain, pain, & more pain.

PR #1785, opened 2023-08-10
Jadudm/etl refactor

jadudm:
For displaying the audit before certification we're going to walk the entire audit. This refactor makes it so that the ETL class can either:

1. Be invoked without arguments to store the submitted work to the dissemination tables, OR
2. Be invoked with `write_to_db=False` so that the objects are not written.

These can then be retrieved with `get_loaded_objects`. These are in-memory, and can either be used to render data out to HTML, or could be saved later. For now, this is so we can ETL the tables, render them, and *not* write them.


backend/audit/etl.py
```diff
@@ -13,49 +13,63 @@
     SecondaryAuditor,
 )
 from audit.models import SingleAuditChecklist
+from copy import deepcopy as dc
```
jadudm, line 16:
```suggestion
```

backend/audit/etl.py
```diff
@@ -13,49 +13,63 @@
...
-    def __init__(self, sac: SingleAuditChecklist) -> None:
+    def __init__(self, sac: SingleAuditChecklist, write_to_db=True) -> None:
         self.single_audit_checklist = sac
         self.report_id = sac.report_id
         audit_date = sac.general_information.get(
             "auditee_fiscal_period_start", datetime.now
         )
         self.audit_year = int(audit_date.split("-")[0])
+        self.write_to_db = write_to_db
+        self.loaded_objects = []
```
jadudm, line 30:
```suggestion
```

backend/audit/etl.py
```diff
@@ -13,49 +13,63 @@
...
+        for key, load_method in load_methods.items():
             try:
-                load_method()
+                objs = load_method()
+                results[key] = objs
             except KeyError as key_error:
                 logger.warning(
                     f"{type(key_error).__name__} in {load_method.__name__}: {key_error}"
                 )
+        self.loaded_objects = results
```
jadudm, line 51:
```suggestion
```

backend/audit/etl.py
```diff
@@ -13,49 +13,63 @@
...
+        return results
+    
+    def get_loaded_objects(self):
+        return self.loaded_objects
```
jadudm, line 55:
```suggestion
```

backend/audit/intake_to_dissemination.py
```diff
@@ -17,45 +17,62 @@
...
+        return self.loaded_objects
+
+    def save_dissemination_objects(self):
+        for key in self.loaded_objects.keys():
+            for o in self.loaded_objects[key]:
+                o.save()
 
     def load_finding_texts(self):
         findings_text = self.single_audit_checklist.findings_text
         findings_text_entries = findings_text["FindingsText"]["findings_text_entries"]
```
sambodeme, line 62: `findings_text["FindingsText"]["findings_text_entries"]` will raise a KeyError if `findings_text_entries` is missing, which could occur with empty submissions. Line 46 catches this error but using the get("findings_text_entries") method would be a more graceful way to handle the absence of this key and it would allow us to return an empty list [] instead of raising an exception

backend/audit/intake_to_dissemination.py
```diff
@@ -65,6 +82,7 @@ def load_findings(self):
             "FindingsUniformGuidance"
         ]["findings_uniform_guidance_entries"]
```
sambodeme, line 83: Similarly to my previous comment using `get("findings_uniform_guidance_entries")` rather than `["findings_uniform_guidance_entries"]` will prevent this from raising an error for empty submissions and allow us to retun an empty list instead.

sambodeme: This comment also applies to the remaing load functions below.

backend/dissemination/models.py
```diff
@@ -384,189 +228,124 @@ class General(models.Model):
...
-    fy_end_date = models.DateField(
-        "Fiscal Year End Date", null=True, help_text=docs.fy_end_date
+        "The date at which the audit transitioned to 'submitted'",
     )
+    # auditor_signature_date = models.DateField(
+    #     "The date on which the auditor signed the audit",
+    # )
+    # auditee_signature_date = models.DateField(
+    #     "The date on which the auditee signed the audit",
+    # )
```
danswick, line 343: @jadudm after clearing up the dates issue, do these signature fields need to be added back in?

jadudm: I... think we only need to disseminate the field that is auto-generated. 

We can talk about that, but I'm not sure if we need to disseminate the name, role, and date from that form. (We capture it elsewhere, right?) However, given that the user can enter anything they want in that field, and what we care about is the system timestamp, I think the date we disseminate is the one that is auto-recorded.

danswick (approved):
Walked through all changes with @sambodeme and @timoballard. Let's merge!

PR #1786, opened 2023-08-10
Resolve total_fed_expenditures, type_report_major_program, type_requirement

JeanMarie-TTS:


backend/audit/templates/audit/audit-info-form.html
```diff
@@ -6,7 +6,7 @@
             <div class="usa-navbar bg-primary-darker text-white width-full flex">
                 <p class="margin-left-2">
                     <a href="/" class="text-gray-10">My audit submissions</a> »
-                    <a href="{% url 'audit:SubmissionProgress' report_id %}"
+                    <a href="/audit/submission-progress/{{ report_id }}"
```
danswick, line 9: I think James has been gradually changing all of these to [Django's tagged style](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#url) so we don't have to update them manually whenever there's a URL change.

danswick (changes requested):
This looks good to me, but I think an unrelated change snuck into `audit-info-form.html`

danswick (approved):
LGTM!

PR #1787, opened 2023-08-11
1658 cross validation validate number of findings across awards findings and findings text workbooks

sambodeme:


sambodeme:
The parent ticket has been updated so I turned this PR into draft so that I can re-work it

PR #1796, opened 2023-08-11
Fix Cypress

jperson1:
Changes:
1. Change schemas to properly include `audit_period_other_months`
2. Changes to Cypress tests:
    * Mock SAM.gov API calls (avoid rate limiting)
    * Full submission looks at the new checklist links
    * Full submission does not go to the end of submission, because the checklist does not enable endsteps without actually uploading


jadudm (approved):
Looks good. We believe we have correctly updated all the build products with this PR as well.

PR #1797, opened 2023-08-11
Add full name field to access and submission page

danswick:
Closes https://github.com/GSA-TTS/FAC/issues/1761

Adds several new "full name" fields to the access and submission page and updates Cypress tests.

Preview:
image


backend/cypress/e2e/check-access.cy.js
```diff
@@ -247,13 +326,13 @@ describe('Create New Audit', () => {
         });
 
         cy.get('.auditee_contacts .delete-contact').click();
-        cy.get('.auditee_contacts .grid-row').should('have.length', 1);
+        cy.get('.auditee_contacts .grid-row').should('have.length', 3);
```
danswick, line 329: I put the new full name fields in their own row, which means _two_ new rows are added each time you add a new contact.

backend/cypress/support/check-access.js
```diff
@@ -1,23 +1,23 @@
...
-    '#certifying_auditee_contact_email', '#certifying_auditee_contact_re_email',
-    '#certifying_auditor_contact_email', '#certifying_auditor_contact_re_email',
-    '#auditee_contacts_email', '#auditee_contacts_re_email',
-    '#auditor_contacts_email', '#auditor_contacts_re_email' 
-  ].forEach(field => addValidEmail(field))
-}
-
 export function testValidAccess() {
-  completeFormWithValidInfo();
+  cy.wrap(accessFields).each((field) => {
```
danswick, line 18: `cy.wrap()` might not be necessary (or used correctly here), but I was having some trouble with the cypress commands executing in the right order. My understanding is that Cypress commands have their own special queueing system that kinda looks and feels like promises but isn't exactly the same and can cause issues with sequencing. This helped tests pass consistently, but I'm not committed to the idea!

jadudm:
Just checking: this adds the field to the front-end, but does not yet percolate it back? It makes sense to bring it in this way, from my point of view, and we can do the backend work as a separate ticket (if that's your intent).

danswick:
@jadudm The [original ticket](https://github.com/GSA-TTS/FAC/issues/1761) says to implement in the form "\*first\*", so that's where I started! But I agree that it makes sense to keep them as separate PRs to make each one more reviewable.

jadudm (approved):
LGTM.

danswick:
@Leighdiddy saw your suggestion too late for this PR but added it to the backend follow-up here: https://github.com/GSA-TTS/FAC/pull/1853/commits/8337cc20f384376e0a1a623497bca5e79710898b

Issue #1658, opened 2023-07-31
Cross-validation: validate number of findings across awards, findings


jadudm:
## Validation Description

[//]: # (A description of the validalidation in plain language.)

This validation is about `AWARD-####` references.

In federal awards, the `AWARD` ref is auto-generated where number of findings is non-zero. In findings, the auditor/auditee enters the number. In findings, each `AWARD` ref should occur `n` times, where `n` is the number of findings reported in the federal awards workbook.

For example, if federal awards is:

| findings | award-ref |
| ---------- | ------------- |
| 1 | AWARD-0001 |
| 3 | AWARD-0002 |

then the findings workbook will have

| award-ref | ... |
| --- | --- |
| AWARD-0001 | ... |
| AWARD-0002 | ... |
| AWARD-0002 | ... |
| AWARD-0002 | ... |

## Section(s):
- [ ] This validation touches ALL sections

[//]: # (Spacing)

&nbsp;

[//]: # (Spacing)

- [ ] Pre-SAC
- [ ] General Information
- [ ] Audit Information
- [x] Federal Awards
- [x] Audit Findings
- [ ] Audit findings text
- [ ] CAP Text
- [ ] Additional UEIs
- [ ] Additional Auditors
- [ ] Audit package (PDF)
- [ ] Tribal data consent
- [ ] Auditee certification
- [ ] Auditor certification

## Error(s)

- [x] This generates one error
- [ ] This possibly generates multiple errors

### Error text

[//]: # (Use Python format string syntax. E.g. f"The section {general-information} is missing everything.")

```python
f"You reported {findings_count} findings in the Federal Awards workbook, but have {findings_rows}..."
```

### Error number

[//]: # (Short identifier followed by timestamp. E.g. UEI_202307050927)

```
FINDINGS_20230731
```

## Notes

[//]: # (Any additional notes for implementation.)

```[tasklist]
### Tasks
- [ ] Error text is edited and approved by @lauraherring 
- [ ] Validation is implemented
- [ ] When data is correct, validation produces no errors
- [ ] When data is incorrect, validation produces expected errors.
```

PR #1803, opened 2023-08-14
1658 cross validation validate number of findings across awards and audit findings

sambodeme:
## Description

See #1658


backend/audit/cross_validation/number_of_findings.py
```diff
@@ -0,0 +1,52 @@
+from .errors import (
+    err_number_of_findings_inconsistent,
+)
+from collections import defaultdict
+
+
+def number_of_findings(sac_dict):
```
tadhg-ohiggins, line 7:
```suggestion
def number_of_findings(sac_dict, *_args, **_kwargs):

```
In another branch we realized that we might need a little more flexibility with what we pass to these functions, so all of them should actually have this signature; I think your existing tests and other code will still work fine with this change.

tadhg-ohiggins (approved):

PR #1804, opened 2023-08-14
Use EST as default time zone

### LINKED ISSUE #1637: The timestamps are wonky. They should be fixed.
Long-and-short is that timestamps, as presented to the user, do not reflect their timezone. Although our storage in UTC is appropriate, it makes no sense to a user to see a timestamp in their future.

When presenting timestamps to the user, they should see a value that makes sense given their current, lived experience.

timoballard:
Partially resolves #1637 

Slack thread: https://gsa-tts.slack.com/archives/C03QCP4FSN8/p1692019397246089

- Change the default time zone to EST (from UTC)
- Update submission checklist to display time zone information as part of "created by" timestamp

image


jperson1 (approved):
I like it!

PR #1814, opened 2023-08-14
Jadudm/test in memory

jadudm:
E2E testing for the data pathway, from workbook generation to validation the data is transformed and accessible via the API.


backend/audit/etl.py
```diff
@@ -21,6 +22,8 @@ class ETL(object):
     def __init__(self, sac: SingleAuditChecklist) -> None:
         self.single_audit_checklist = sac
         self.report_id = sac.report_id
+        # MCJ QUESTION: What is the second parameter to `get` for?
+        # https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.get
```
jadudm, line 26:
```suggestion
```

backend/audit/etl.py
```diff
@@ -76,23 +79,32 @@ def load_findings(self):
 
         for entry in findings_uniform_guidance_entries:
             findings = entry["findings"]
+            program = entry["program"]
             finding = Finding(
-                award_reference=entry["award_reference"],
+                award_reference=program["award_reference"],
                 report_id=self.report_id,
-                finding_seq_number=entry["seq_number"],
+                # finding_seq_number=entry["seq_number"],
```
jadudm, line 86:
```suggestion
```

backend/audit/etl.py
```diff
@@ -122,9 +133,10 @@ def load_federal_award(self):
                 award_reference=entry["award_reference"],
                 federal_agency_prefix=program["federal_agency_prefix"],
                 federal_award_extension=program["three_digit_extension"],
-                additional_award_identification=program[
-                    "additional_award_identification"
-                ],
+                # CONDITIONAL
+                additional_award_identification=self.conditional_lookup(
```
jadudm, line 137: If a helper to do conditional lookups in the JSON object is going to be used, use it consistently everywhere. Or, don't use it.

backend/audit/etl.py
```diff
@@ -144,52 +162,50 @@ def load_federal_award(self):
...
+                        accounting_policies=accounting_policies,
+                        is_minimis_rate_used=is_minimis_rate_used,
+                        rate_explained=rate_explained,
+                    )
+                    note.save()
+                else:
+                    for entry in entries:
+                        note = Note(
+                            report_id=self.report_id,
+                            note_seq_number=entry["seq_number"],
```
jadudm, line 201: Where does the `seq_number` come from in this code? We don't have it in the workbooks.

backend/run.sh
```diff
@@ -19,9 +19,21 @@ fi;
...
 python manage.py drop_deprecated_api_schema_and_views && 
-  python manage.py drop_api_schema && 
-  python manage.py create_api_schema && 
-  python manage.py create_api_views
+python manage.py drop_api_views &&
+python manage.py drop_api_schema && 
+python manage.py create_api_schema && 
+python manage.py create_api_views
+
+if [[ ("${ENV}" == "LOCAL" || "${ENV}" == "STAGING") && "$LOAD_TEST_DATA" == 1 ]];
```
danswick, line 29: This block looks like it was maybe used for experimenting?

backend/audit/test_schemas.py
danswick, line 1: The diff is odd here. This file matches `main` but is still showing up as a change.

backend/run.sh
jadudm: This block is necessary to control data loading; we run `run.sh` in our local container stack. `docker-compose-data.yml` sets the LOAD_TEST_DATA env var. I think the other ENV tests are not needed, but this controls whether or not `run.sh` is running under our "normal" stack, or a stack that loads the historic data.

backend/dissemination/migrations/0027_additionalein_and_more.py
danswick, line 1: Replaced 0026 with the one from main and regenerated as 0027.

backend/run.sh
danswick: I should've highlighted the whole thing, but I was referring to the hardcoded dbkey list on line 32. Is that dbkey list stable / did you have a particular characteristic in mind when picking them or are they just convenient examples? Maybe just a comment would suffice for future travelers.

backend/schemas/source/sections/FederalAwards.schema.jsonnet
danswick, line 1: Are any of the changes below supposed to build on the ones from 1909 or should this file be reset now that 1909 is in main?

backend/docker-compose-data.yml
danswick, line 1: This is out of scope for this PR, but we might consider modularizing these services since they're almost identical between this compose file and the default one.

backend/audit/management/commands/workbooks_e2e.py
```diff
@@ -0,0 +1,223 @@
...
+
+def step_through_certifications(sac, SAC):
+    sac.transition_name.append(SAC.STATUS.SUBMITTED)
+    sac.transition_date.append(datetime.date.today())
+    sac.transition_name.append(SAC.STATUS.AUDITOR_CERTIFIED)
+    sac.transition_date.append(datetime.date.today())
+    sac.transition_name.append(SAC.STATUS.AUDITEE_CERTIFIED)
+    sac.transition_date.append(datetime.date.today())
+    sac.transition_name.append(SAC.STATUS.CERTIFIED)
+    sac.transition_date.append(datetime.date.today())
```
danswick, line 54: TODO: submit the certification object so the frontend renders correctly.

backend/audit/fixtures/workbooks/sac_creation.py
```diff
@@ -0,0 +1,372 @@
...
+    sac.report_id = dbkey_to_test_report_id(Gen, dbkey)
+
+    Access = apps.get_model("audit.Access")
+    Access.objects.create(
+        sac=sac,
+        user=user,
+        email=user.email,
+        role="editor",
+    )
+    # Why not give me all the access? This way, I can run the test all the way through!
```
danswick, line 256: Question for later: when we get to cross-validation, using the same access credentials for certifying 'tee and 'tor should fail, right?

backend/audit/fixtures/workbooks/federal_awards.py
```diff
@@ -0,0 +1,271 @@
...
+
+    cfdas = Cfda.select().where(Cfda.dbkey == dbkey).order_by(Cfda.index)
+    map_simple_columns(wb, mappings, cfdas)
+
+    # Patch the clusternames. They used to be allowed to enter anything
+    # they wanted.
+    valid_file = open(f"{settings.BASE_DIR}/schemas/source/base/ClusterNames.json")
+    valid_json = json.load(valid_file)
+    # This was removed from the CSV...
+    valid_json["cluster_names"].append("STATE CLUSTER")
```
danswick, line 196: We removed N/A as well. Do we need something like this? 

```suggestion
    valid_json["cluster_names"].append("STATE CLUSTER")
    valid_json["cluster_names"].append("N/A")
```

jadudm:
This merged in other pieces.

PR #1817, opened 2023-08-14
Update submission progress check to work as a validation function, and related changes.

tadhg-ohiggins:
Certification
Is wrapped up with submission.
Try to advance both.

-----

Fixes the conditional display of the pre-submission check.

Moves the function for checking the progress of a submission into cross-validation, because it turns out that checking to see if there are any incomplete sections is also something we need to do at cross-validation time. Also change the signatures of the cross-validation functions to accommodate other arguments, here so that we’re able to send `SingleAuditReportFile` to `submission_progress_check`, but also in case there are other outliers in future (the most likely candidate being having to also pass `Access` objects for some reason.)

Adds `tribal_data_sharing_consent` and `submission_progress_check` to the list of functions run at cross-validation time.

Updates the tests to work with these changes.

### DIFF
--- a/backend/audit/cross_validation/submission_progress_check.py
+++ b/backend/audit/cross_validation/submission_progress_check.py
@@ -1,0 +1,131 @@
+def submission_progress_check(sac, sar=None, crossval=True):
+    """
+    Because this function was initially created in a view and not as a
+    cross-validation function, it needs the crossval argument to distinguish between
+    running in a cross-validation context and running in the context of needing to
+    return progress information to the submission progress view.
+
+    crossval defaults to True because we don't want to have to change all the calls to
+    the validation functions to include this argument.
+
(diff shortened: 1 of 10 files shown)


backend/audit/cross_validation/submission_progress_check.py
```diff
@@ -0,0 +1,131 @@
...
+
+    if not crossval:
+        return result  # return the cross-validation shape
+
+    if complete:
+        return []
+
+    return [
+        {
+            "error": f"The following sections are incomplete: {', '.join(incomplete_sections)}."
```
timoballard, line 90: given that we're statically returning a single `dict` item in this `array`, would it make sense for `result` to always be a `dict` which conditionally includes a top-level `error` key? In the `not crossval` case, the `error` (if present) could just be ignored by the caller

timoballard (approved):

Issue #1727, opened 2023-08-04
Audit submission access page: names


jadudm:
To be ticketed more.

This is the fact that our audit submission access page only takes email addresses.

We want to add back in the first/last name.

If we have a field called `Full name` along with each of the email/re-enter email blocks, we would be good.

This will involve the addition the field to the view, the validation, and... the dissemination *maybe*. 

```[tasklist]
### Tasks
- [ ] https://github.com/GSA-TTS/FAC/issues/1761
```

jadudm:
https://www.figma.com/file/lnkXJTfKzD3YBhiJZ2QFEv/FAC-final?type=design&node-id=1-31587&mode=design&t=YqjMan8AaqOY4f0U-4

PR #1853, opened 2023-08-16
Add access full name to model

danswick:
Closes https://github.com/GSA-TTS/FAC/issues/1727
Follow-up to https://github.com/GSA-TTS/FAC/pull/1797

This PR adds a new field to the `Access` model: `fullname'. There are some cascading changes related to that, including:

* Appending `_email` to the other access fields that were once named "certifying auditee contact" and the like. Now we have fields like `auditee_contacts_email` and `auditee_contacts_fullname`.
* Updates the frontend form to make the non-certifying contacts fields optional, BUT requires completeness if a user chooses to fill one out. There are some other frontend updates and fixes as well.

```[tasklist]
### Tasks
- [ ] ~~Test for complicated field validation stuff I added on 8/17~~ I believe the new edge cases are handled adequately on the frontend
- [x] Catch up to main
- [x] Resolve migration conflicts
- [x] Fix linting problems
```

## To review this PR:

1. Fetch this branch and run `make docker-clean && make docker-first-run && docker compose up`
2. You can start tests in another terminal window with `docker compose exec web /bin/bash -c "python manage.py test; /bin/bash"` or your preferred method for running tests.
3. Navigate to the locally running site, log in, and begin a new audit. 
4. When you get to step 3 (access and submission), try submitting the form in a few different configurations. You might try:
  * Filling out all of the default fields correctly
  * Adding some additional contacts. Deleting some to make sure they still work.
  * Test the completeness checks for auditee and auditor contacts. It should prevent you from progressing if you only fill out the fields part way.


danswick:
While trying to wire this up to our backend, I realized we have kind of an odd mix of optional and required fields for this contact information, particularly for Audit{ ee || or } contacts:

* Adding any contacts to these fields is optional. You only _need_ to list certifying officials (Is that right, @Carley-Sullivan?)
* We currently expect form data to come in as a flat list of entries:

    ```
    auditee_contacts_email: "blah@blah.com",
    auditee_contacts_fullname: "A. Human",
    auditee_contacts_email: "auditee@blerg.com",
    auditee_contacts_fullname: "A. Person",
    auditor_contacts_email: "auditor@blerg.com",
    auditor_contacts_fullname: "A. Dittee",
    ...
    etc.
    ```

* In that format, and assuming every field is optional, we just have to hope that users don't leave a name or email field blank, or things could get out of order.
* We could append an index to the end of every entry, but most of the backend expects these contacts as an undifferentiated list, so it seemed more difficult to parse and handle the entries on the backend.
* Another option (the one I went with) is to enforce _complete contacts sections_ on the frontend. That is, if a user provides an email, they must also provide a name and vice-versa. If that goes as expected, we can have some confidence that the contacts information will be delivered in the correct order.

danswick:
Here are a couple of screenshots from the frontend:

This is what happens when you enter an email but not aname. The error messages are contained to their "contacts block." _The error message spacing is **not great** but we can come back to that_.

image

This is what happens when you enter an email but not a name. 

image

tadhg-ohiggins (approved):

tadhg-ohiggins:
We discussed this extensively in a video meet; approved.

PR #1856, opened 2023-08-16
Cypress Full Submission Fix

jperson1:
# Cypress Full Submission Fix - Up To Cross-validation

Changes:
All changes are made with `full-submission.cy.js` in mind. This is the test that runs when attempting to deploy to staging. 

1. Some pre-filled excel files get their own subfolder under `/fixtures/test_workbooks`. They are not generated by code anywhere, that is a future enhancement. 
2. Workbook uploads through Cypress will actually upload a document. This will ensure there is data in the DB for cross-val to run on. 
3. The audit information form is filled out after uploading all the excel files. 
4. Cross validation is run, and then it's done. The submission is _not_ locked. 
5. Auditee information fiscal period dates are now 01/01/2022 -> 12/31/2022, instead of today -> today, due to form validation that _works_, woo!
6. Fixed the cross validation table. It slapped everything in one row, rather than several rows.


eazapata23:
Looks good i'm able to successfully complete a full submission

jperson1:
The workflow for Cypress now passes when pointed at this branch:

https://github.com/GSA-TTS/FAC/actions/runs/5883459366/job/15956217452

backend/cypress/support/cross-validation.js
```diff
@@ -1,8 +1,5 @@
 export function testCrossValidation() {
   cy.url().should('match', /\/audit\/cross-validation\/[0-9A-Z]{17}/);
   cy.get(".usa-button").contains("Begin Validation").click();
-  cy.get('.usa-fieldset').contains('Looks good!');
-
-  const reportTestId = '2023AUG0001000010';
-  cy.visit(`/audit/submission-progress/${reportTestId}`);
```
phildominguez-gsa, line 7: I had done this because the other `test*` methods seem to click `Continue` to return to the main submission page, but the cross validation page doesn't have such a link. I feel like we should still try to return to that page here, though, to match the others. Thoughts?

phildominguez-gsa (changes requested):

backend/cypress/support/cross-validation.js
jperson1: Makes sense! I've extended the test to walk through locking the submission for certification, which then sends it back to the checklist view.

phildominguez-gsa (approved):

eazapata23 (approved):
LGTM

PR #1857, opened 2023-08-16
Add pull request template for GitHub and link to it in development.md

tadhg-ohiggins:
Checklists are the way.
We should use them with pull requests.
Here is a template.

-----

Add a pull request template with the steps submitters and reviewers should follow on PRs.


danswick (approved):
Looks great. Thanks, @tadhg-ohiggins!

PR #1863, opened 2023-08-17
Move the PR template to the location required to make it the default

tadhg-ohiggins:
Defaults are the way.
We should set one with a file.
This is a template.

-----

The prior version of this PR merely made the template available as an option via query parameters; this should hopefully make it the default.


danswick (approved):
LGTM!

Issue #1678, opened 2023-08-02
MIssing fields in audit info


jadudm:
See 

https://github.com/GSA-TTS/FAC/issues/1278#issuecomment-1662495908

We need the design for the optional special purpose framework questions. That is, if someone checks "not a GAAP" audit, we then must present three more optional questions.

We will just reference *yet another ticket* to close this one:

```[tasklist]
### Tasks
- [ ] #1729 
```

jadudm:
We need to backfill the optional questions into the SoT. @Leighdiddy

jadudm:
This was done.

PR #1877, opened 2023-08-18
Audit Info - Optional Fields JS Fixes

jperson1:
# Audit Info Fixes - Optional Fields "required" and "hidden" Attributes

Issue: https://github.com/GSA-TTS/FAC/issues/1678

Changes:
1. Rewritten event handlers for the audit info form.
    * One handler on the form wipes the optional fields if they shouldn't send data. Such as in the case a user selects the relevant checkbox, fills out the optional questions, then unchecks that box. 
    * One handler on the "Not GAAP" checkbox that shows/hides the optional questions and makes them required/not required as necessary.
    * Several handlers on the optional questions so that they do not _stay_ required if they get answered.
2. Minor change to the audit info form schema, to make a bool a string
    * This field can be either True, False, or nothing. As a quick fix, I made it a string field where an empty string is used when the question is unanswered.

How to test:
1. Swap to this branch and run normally.
2. Create an audit or navigate to an existing audit, and find the audit information form
3. Try to break it. This may include:
    * Trying to submit without filling things out
    * Trying to submit with data in the optional fields, but without the "Not GAAP" field checked
    * Changing data that's already in your DB, and ensuring it updates properly

## PR checklist: submitters

- [x]   Do manual testing locally. Our tests are not good enough yet to allow us to skip this step.
- [ ]   If Git surgery is necessary at any point, repeat the testing after it’s finished.

## PR checklist: reviewers

- [x]   Pull the branch to your local environment and run `make docker clean; make docker-first-run && docker compose up`; then run `docker compose exec web /bin/bash -c "python manage.py test"`
- [x]   Manually test out the changes locally.
- [x]   If Git surgery is necessary at any point (such as during a merge party), repeat the testing after it’s finished.
        

jadudm:
`if not_gaap` is present, then we should have three additional boolean fields. (The `is_sp` sequence.) 

If it is *not* present, those three boolean fields should *not* be present.

backend/audit/templates/audit/audit-info-form.html
```diff
@@ -69,10 +68,9 @@ <h1>Financial statements</h1>
                             <input id="is_sp_framework_required--true"
                                 name="is_sp_framework_required"
                                 class="usa-radio__input"
-                                required
                                 type="radio"
                                 value="True"
-                                {% if form.cleaned_data.is_sp_framework_required is True %}checked{% endif %} />
```
tadhg-ohiggins, line 75: We should be using `is True` here unless we intend to be checking against a non-boolean value that we want to be coerced into True/False.

sambodeme: Agreed. I missed this one.

tadhg-ohiggins (approved):

PR #1909, opened 2023-08-22
transfer changes from PR 1814: schemas and assorted tests

danswick:
## PR checklist: submitters

## PR checklist: reviewers


backend/audit/test_schemas.py
```diff
@@ -403,77 +391,52 @@ def test_all_booleans(self):
...
-        for _ in range(5):
-            for n in range(1, 5):
-                opinions = random.sample(self.ALL_OPINIONS, n)
-                basis = random.sample(sp_framework_basis, n)
-                simple_case["sp_framework_opinions"] = opinions
-                simple_case["sp_framework_basis"] = basis
-                validate(simple_case, self.AUDIT_INFO_SCHEMA)
+        for result in gaap_results:
+            simple_case["gaap_results"] = [result]
+            validate(simple_case, schema)
```
danswick, line 412: This doesn't work because it's changing the `gaap_results` value for a case that explicitly doesn't include the required property.

backend/schemas/source/sections/AuditInformation.schema.jsonnet
```diff
@@ -110,6 +110,22 @@ local AuditInformation = Types.object {
...
+// WARNING: If a comment is the last thing in a Jsonnet file, then the formatter will
+// ALWAYS add a new line to the end of the file, meaning this file will change every
+// time we run the formatter.
+//
+// The directions below suggest that the array should be added to the end of the file
+// to run local tests. There might be a better way: create a new file that we run the tests
+// against that imports from this one, and wraps the AuditInformation object in an array.
+// This will have the same effect, but make it testable without modifying this file.
+
+//To manually text against data_fixtures/audit/excel_schema_test_files/audit-information-pass-01.js
```
sambodeme, line 122: This comment was meant to be a hint for whoever wants to do a quick manual testing against an array of AuditInformation.  If it causes issue then it is better to get rid of it. Moving up does not make sense as the item 1 below is now out of context.

backend/schemas/source/base/Base.libsonnet
```diff
@@ -407,32 +408,36 @@ local SchemaBase = Types.object {
...
+    ClusterNamesNAStateOther: Types.string {
       description: 'All cluster names',
-      enum: ClusterNames.cluster_names + [Const.NA, Const.OTHER_CLUSTER],
+      enum: ClusterNames.cluster_names + [Const.NA, Const.STATE_CLUSTER, Const.OTHER_CLUSTER],
     },
-    ClusterNamesStateOther: {
+    ClusterNamesStateOther: Types.string {
       description: 'All cluster names',
-      enum: ClusterNames.cluster_names + [Const.OTHER_CLUSTER],
+      enum: ClusterNames.cluster_names + [Const.OTHER_CLUSTER, Const.OTHER_CLUSTER],
```
danswick, line 433: Duplicate `Const.OTHER_CLUSTER`

backend/schemas/source/data/cluster-names-20230626.csv
```diff
@@ -1,42 +1,41 @@
-ID,NAME,SORT_NUM
-55,HEAD START CLUSTER,179
-50,477 CLUSTER,90
-51,"MATERNAL, INFANT, AND EARLY CHILDHOOD HOME VISITING CLUSTER",207
-1,N/A,1
-2,RESEARCH AND DEVELOPMENT,10
-3,STUDENT FINANCIAL ASSISTANCE,20
-4,STATE CLUSTER,30
```
danswick, line 8: `STATE CLUSTER` was removed in favor of a const in the libsonnet. Is that the right decision? If so, we should also remove `N/A` from line 5.

backend/audit/test_schemas.py
```diff
@@ -403,76 +391,81 @@ def test_all_booleans(self):
...
+    # 'gaap_results': ['not_gaap'], 'is_going_concern_included': True,
+    # 'is_sp_framework_required': True, 'sp_framework_basis': ['cash_basis',
+    # 'tax_basis', 'contractual_basis'], 'sp_framework_opinions':
+    # ['unmodified_opinion', 'qualified_opinion'],
+    # 'is_internal_control_deficiency_disclosed': False,
+    # 'is_internal_control_material_weakness_disclosed': True,
+    # 'is_material_noncompliance_disclosed': True,
+    # 'is_aicpa_audit_guide_included': True, 'is_low_risk_auditee': False,
+    # 'agencies': ['32', '45']}] should not be valid under {'required':
+    # ['is_sp_framework_required']}">
```
danswick, line 415:
```suggestion
```

backend/audit/test_schemas.py
```diff
@@ -872,7 +865,10 @@ def test_number_of_audit_findings(self):
             simple_case["FederalAwards"]["federal_awards"][0]["program"][
                 "number_of_audit_findings"
             ] = 0
-            self.assertRaises(exceptions.ValidationError, validate, simple_case, schema)
+            # We cannot find, in the UG, anything that suggests this is true.
+            # This seems like it should be allowed to pass.
+            # self.assertRaises(exceptions.ValidationError, validate, simple_case, schema)
+            validate(simple_case, schema)
```
danswick, line 871: Since this change reverses the meaning of the `test_number_of_audit_findings` test, would it be better to omit it altogether until we're clear on the requirements from UG?

backend/schemas/source/sections/FederalAwards.schema.jsonnet
```diff
@@ -402,16 +373,23 @@ local FederalAwardEntry = Types.object {
...
-  required: ['section_name'],
+  // allOf: [
+  //   // {
+  //   //   properties: {
+  //   //     section_name: {
+  //   //       const: Sheets.section_names.FEDERAL_AWARDS,
+  //   //     },
+  //   //   },
+  //   // },
+  // ],
```
danswick, line 391: Any reason to keep this comment? 
```suggestion
```

backend/audit/test_schemas.py
sambodeme: Makes sense. On the other hand, because we are now allowing 0 through the schema, the test case should reflect what we have in the schema and when we update the schema, this test will fail as a reminder for us to update the test cases accordingly.

backend/audit/test_schemas.py
```diff
@@ -403,76 +391,81 @@ def test_all_booleans(self):
...
-                simple_case["sp_framework_basis"] = bad_values
-                self.assertRaises(
-                    exceptions.ValidationError,
-                    validate,
-                    simple_case,
-                    self.AUDIT_INFO_SCHEMA,
-                )
+        for word in not_gaap_values:
+            simple_case["gaap_results"] = [word]
+            self.assertRaises(exceptions.ValidationError, validate, simple_case, schema)
```
sambodeme, line 465: Any reason why we'll want to test bad values for `gaap_results` only but not for  `sp_framework_opinions` and `sp_framework_basis` ? @danswick @tadhg-ohiggins

danswick: Good point. I added two very simple tests below.

backend/schemas/source/data/cluster-names-20230626.csv
danswick: I see what's happening. There are three conditions for "cluster name" in the Federal Awards awards:

1. One of the listed cluster names, which includes "N/A"
2. State cluster, which requires an entry in "state_cluster_name"
3. Other, which requires an entry in "other_cluster_name"

`STATE_CLUSTER` was removed from the source JSON so it doesn't show up in the options for condition `1`.  `N/A` remains because it's included in condition `1`. 

Given that, I think we should remove `N/A` from this file.

backend/schemas/source/base/ClusterNames.json
```diff
@@ -3,10 +3,8 @@
     "HEAD START CLUSTER",
     "477 CLUSTER",
     "MATERNAL, INFANT, AND EARLY CHILDHOOD HOME VISITING CLUSTER",
-    "NAN",
```
danswick, line 6: Looks like the schema build process was turning `N/A` into `"NAN", which explains why we needed the extra constant in `Base.libsonnet`

backend/audit/test_schemas.py
```diff
@@ -403,76 +391,77 @@ def test_all_booleans(self):
...
-                simple_case["sp_framework_basis"] = basis
-                validate(simple_case, self.AUDIT_INFO_SCHEMA)
+        for word in not_gaap_values:
+            simple_case["gaap_results"] = [word]
+            self.assertRaises(exceptions.ValidationError, validate, simple_case, schema)
 
-    def test_bad_gaap_results(self):
+    def test_bad_sp_framework_opinions(self):
+        schema = self.AUDIT_INFO_SCHEMA
         simple_case = jsoncopy(self.SIMPLE_CASE[0])
```
sambodeme, line 451: This should be self.SIMPLE_CASE[1] as `sp_framework_opinions` is only expected when `gaap_results=["not_gaap"]` and that condition is true for self.SIMPLE_CASE[1] but not for self.SIMPLE_CASE[0]. Otherwise a validationError will be triggered as expected but not for the expected reason. FYI, I will make the change

backend/schemas/source/base/ClusterNames.json
danswick: This is actually a default behavior for pandas: https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#na-values

backend/audit/test_schemas.py
```diff
@@ -403,76 +391,77 @@ def test_all_booleans(self):
...
-                    validate,
-                    simple_case,
-                    self.AUDIT_INFO_SCHEMA,
-                )
+        simple_case["sp_framework_opinion"] = "not_a_real_opinion"
+        self.assertRaises(exceptions.ValidationError, validate, simple_case, schema)
+
+    def test_bad_sp_framework_basis(self):
+        schema = self.AUDIT_INFO_SCHEMA
+        simple_case = jsoncopy(self.SIMPLE_CASE[0])
```
sambodeme, line 458: As per my previous comment, this should be self.SIMPLE_CASE[1] as `sp_framework_basis` only exists when `gaap_results=["not_gaap"]`. FYI, I will make the change

backend/audit/test_schemas.py
```diff
@@ -403,76 +391,77 @@ def test_all_booleans(self):
...
+
+        simple_case["gaap_results"] = "an_invalid_basis"
```
sambodeme, line 460:
```suggestion
        simple_case["sp_framework_basis"] = "an_invalid_basis"
```

backend/schemas/source/sections/FederalAwards.schema.jsonnet
```diff
@@ -402,16 +373,23 @@ local FederalAwardEntry = Types.object {
 local Meta = Types.object {
   additionalProperties: false,
   properties: {
-    section_name: Types.string {
-      enum: [Sheets.section_names.FEDERAL_AWARDS],
-    },
+    section_name: Types.string,
```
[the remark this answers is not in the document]
sambodeme: I will revert this for now. If we do want to skip this validation then we should skip that for all the sections (not `FederalAwards` only). Besides we have logic that relies on section_name to determine if a workbook originated from us (so this should be required as well).

backend/schemas/source/data/cluster-names-20230626.csv
jadudm: That makes sense. It probably made it by because of the conversion of N/A to NAN? Either way, affirming the logic here.

jadudm (approved):
Many, many thanks. The schemas are where, in many cases, a lot of attention is needed, and it was excellent to have the review of multiple people here. TY.

Issue #1603, opened 2023-07-24
General form should handle international addresses for auditors


danswick:
## Description

SF-SAC includes a provision for listing international primary auditors (`ELECAUDITHEADER.CPAFOREIGN`), which we haven't yet implemented in our intake form or data model. Note that there is no provision for international addresses for secondary auditors.

Here is a screenshot of the relevant section from the IDES manual: 
<img width="667" alt="screenshot of IDES manual section describing how to submit address information for international auditors" src="https://github.com/GSA-TTS/FAC/assets/2365503/8abeb236-ed4e-4b58-b837-3941234c0700">

## Next steps

In IDES, this functionality is built on a conditional UI. There are two UI's: one for US-based auditors (this is what we have now and is the default in IDES) and one for non-US-based auditors. UIs are conditionally shown based on which dropdown item a user selects. 

A rough outline of what implementation might look like:

- [ ] Update Figma designs to add a conditional "international audit firm/organization address" field. As far as I can tell, this should be a large, single text box, not a multi-field address form with fine-grained field validation. 
- [ ] Change the General form's auditor "country" field to a dropdown with two options: `US` and `non-US`. This is in the current Figma designs, but the web form is a text box right now.
- [ ] Update data model and validations to treat auditor country as a dropdown
- [ ] Update data model and validations to include an optional international address field (assuming this will just be a large string)
- [ ] Handle conditional country field in web form: if `US` is selected, the remaining address fields are required (this is the current state). If `non-US` is selected, only the "international address" text field is required.
- [ ] Handle conditional country field in intake validation: if `US`, the default address fields are required and validated as they are today. If `non-US`,  default address fields should be empty and there should be a long string for "international address". 
- [ ] Update ETL to handle the new international address field. 

cc @Leighdiddy for possible figma things, @jperson1 for frontend work, @Carley-Sullivan for truthiness, and @JeanMarie-TTS for eventual ETL connection.

timoballard:
Some initial work on this here https://github.com/GSA-TTS/FAC/pull/1646

JeanMarie-TTS:
PR ready for review

PR #1937, opened 2023-08-24
Jmm/1603 international addresses

JeanMarie-TTS:
## PR checklist: submitters

- [x]   Link to an issue if possible. If there’s no issue, describe what your branch does. Even if there is an issue, a brief description in the PR is still useful. #1603

- Had to **match migrations** with those in main for dissemination and audit in order to get tests to pass on GH.

## PR checklist: reviewers


sambodeme (changes requested):
I tested this feature and it mostly works well, but I did notice a couple of issues:

1. When a user selects `USA` and fills out the form, but then switches to a `non-USA` option and re-enters the data, the previous USA-specific inputs are cached and submitted with the form, causing validation to fail and crashes the UI (see screenshot below). The USA-only data should be cleared when users switch from a USA to a non-USA address.
2. When a `non-USA` address is entered,  the fields below come back empty. This makes sense for most fields, except for 'ado.' Should the 'ado' field be empty for non-USA addresses?

jperson1:
I've added a few things:
1. Added a conditional check & wipe for the auditor address in the GeneralInformationFormView.
2. Changed the international address field to a textarea rather than a textarea formatted input field. Now the cursor is top-left rather than center-left.
3. Merged the branch with main to keep it up-to-date for pulling.

jperson1:
@sambodeme 
> When a non-USA address is entered, the fields below come back empty. This makes sense for most fields, except for audit_period_other_months Should the audit_period_other_months field be empty for non-USA addresses?

`audit_period_other_months` should be empty if its field above the Auditee information is empty (or unselected). I think they're showing up next to each other because they are adjacent alphabetically? It's not related to the address fields. 

This one:
image

sambodeme:
@jperson1 I'll give it a try tomorrow. I noticed that when I toggle between the USA and non-USA settings, the `audit_period_other_months` field changes from non-empty to empty. I had `Audit Period`  set to "Annual" and I didn't alter that setting while switching between USA and non-USA.

sambodeme (approved):
Re-tested and LGTM

PR #1956, opened 2023-08-26
Make section names constants

tadhg-ohiggins:
Naming things is hard—
Metaclasses for the win.
Now names are constants.

-----

Add an `NC` class to `audit/cross_validation/naming.py` that has class-level read-only attributes storing the canonical snake-case section names.

This makes it possible to use e.g. `NC.GENERAL_INFORMATION` instead of the string `"general_information"`.

This PR uses that class in `audit/cross_validation/submission_progress_check.py` instead of strings.

`NC` has a deliberately short name so as to make the switch from strings to using it relatively painless and only add one character to line length.

This PR only uses `NC` in a very limited number of places, so this shouldn’t change anything. No features should change, so local test runs and code review should be sufficient. 

Uses metaclass wrangling to set up a class that has read-only class-level attributes containing the all_caps section names, each of which has the corresponding snake_case as its value.

## PR checklist: submitters

## PR checklist: reviewers

- [x]   Pull the branch to your local environment and run `make docker clean; make docker-first-run && docker compose up`; then run `docker compose exec web /bin/bash -c "python manage.py test"`
        

jadudm (approved):
Looks good. Walked this with the author, and @jperson1 is doing a local run and check before merge. I'll leave merge to @jperson1 .

jperson1 (approved):
Looks good! All tests passing, manual check is good, and no conflicts following a merge with main.

PR #1989, opened 2023-08-31
Create SECURITY.md

jadudm:
Was lost, but now was found. Adding per nudge from @pauldoomgov .

This PR only adds one file. However, the question is "is this the file and contents we want?" Hence, review might involve requesting, suggesting, or making changes.


mogul (approved):
This is fine to start and address AllStar's finding. We can iterate further on it if and when needed.

PR #2024, opened 2023-09-05
This fixes the data load script

asteel-gsa:
What needed to be done:
Set $PATH inside the calling script `load_data.sh` and ensure that when the script runs, it is executable.

## PR checklist: reviewers


.github/workflows/deploy-application.yml
```diff
@@ -86,4 +86,4 @@ jobs:
           cf_password: ${{ secrets.CF_PASSWORD }}
           cf_org: gsa-tts-oros-fac
           cf_space: ${{ env.space }}
-          command: cf run-task gsa-fac -k 6G -m 1G --name load_data --command ./load_data.sh
+          command: chmod +x ./load_data.sh && cf run-task gsa-fac -k 6G -m 1G --name load_data --command ./load_data.sh
```
mogul, line 89: This won't make a difference because the `chmod` will happen to the `load_data.sh` file in the GHA runner on the local checkout, while the `run-task` will happen in the context of a new container.

mogul (changes requested):
We need to figure out why the file is getting checked out and deployed without +x.

.github/workflows/deploy-application.yml
asteel-gsa: Good catch, that needed to be in `--command " "`

asteel-gsa:
> We need to figure out why the file is getting checked out and deployed without +x.

I think when I renamed/moved the file, it went back to 644, but with this commit it will be 755. 

I can take it out tomorrow morning, and see if the +x stuck this time.

jadudm (approved):
Looks good. This is a "test remote" feature, and Alex has debugged the local config nicely for remote use. This should deploy to `dev`, and then `staging`.

PR #2052, opened 2023-09-07
Manage ASGs, user perms, and deployer accounts using Terraform

mogul:
A few changes are happening here:

- The meta module should not default to writing files useful for local interaction with environments. They're not needed unless one is tinkering with the `terraform/` directory locally, which is generally discouraged. (CI/CD runs have been trying to create these local files every time we apply meta, and they shouldn't be.) Now the default is to avoid that behavior.
- The version of the CF provider was updated both in the meta module, and all the environment modules that it maintains.
- The meta module manages the ASGs explicitly now that the provider gave us this capability.
- The meta module manages the set of users again (which it was temporarily unable to do)
- The meta module now manages the existing set of deployer services and keys.


mogul:
Note that [the `meta` plan is now empty](https://github.com/GSA-TTS/FAC/pull/2052#issuecomment-1710689976) as expected.

terraform/meta/bootstrap-env/environment.tf
```diff
@@ -4,6 +4,7 @@ locals {
 
 # Environments often need to know the base org that everything is in
 resource "local_file" "cf_org" {
+  count           = (var.populate_creds_locally == true) ? 1 : 0
```
asteel-gsa, line 7: Defaults to `false`. `populate_creds_locally` will be set true only for local env `terraform apply -var populate_creds_locally=true`

asteel-gsa (approved):
Looked over the shoulder w/ bret. This will resolve terraform `meta` "changes" when creating a pr

mogul:
Moving back to draft status as I need additional changes in newer commits I've applied by hand to be included. Otherwise when this is applied, it will undo those changes, which might be destructive!

jadudm (approved):
It remains true that I do not know enough TF to confidently review these changes. Merging.

mogul:
> It remains true that I do not know enough TF to confidently review these changes. Merging.

This is some Very Advanced Terraform With Lots Of Fiddly Bits, so don't feel bad. It's tricky enough to get everything right both implementation- and compliance-wise that I'm definitely writing with an eye to make it a shared module in https://github.com/18F/terraform-cloudgov. That way other programs can benefit from not wrangling this in in future, but we bear that initial complexity in PR friction until it's "done" enough for me to move it out of our repo.

Thanks for jumping on the merge!

mogul:
(Also I make verrrrry tiny commits in Terraform so the list might be much easier to learn from than the full diff.)

terraform/meta/meta.tf
```diff
@@ -0,0 +1,36 @@
...
+}
+
+module "environments" {
+  for_each               = local.spaces
+  source                 = "./bootstrap-env"
+  name                   = each.key
+  org_name               = local.org_name
+  developers             = concat(local.developers, [data.cloudfoundry_user.meta_deployer.id])
+  managers               = local.managers
+  reponame               = "GSA-TTS/FAC"
```
mogul, line 24: Note to self: We should be hoisting this up into the locals specified in `config.tf`.

Issue #2022, opened 2023-09-05
Implement code related to modifications to flowchart


gsa-suk:
Breaking this out from its parent epic. See https://github.com/GSA-TTS/FAC/issues/1218 for more discussion and history.

This covers the implementation related to modifications to flowchart.

JeanMarie-TTS:
- [ ] Implement updated logic in a new django. app called support
- [ ] Remove cog_over code from audit
- [ ] Create CognizantBaseline table for the period 2021-2026
- [ ] Create CognizantAssignment table to track assignment of a cog_agency to a submission, including overrides.
- [ ] Trigger downstream updates when cog_assignment is overridden.
- [ ] Implement functionality in audit to invoke cog/over determination
- [ ] Use signals in both audit and support to trigger actions , and thereby avoiding circular references in the models.py classes

JeanMarie-TTS:
To be implemented in a follow on tocket:

- [ ] Admin screens for a staff user to view and create CpgnizantAssignments - Must
- [ ] Validate cog/over logic by doing a run using 2022 submissions from census - Could
- [ ] Seed CognizantBaseline data with 2021-2026 spreadsheet from Census - Must

JeanMarie-TTS:
PR stuck in a traffic jam.

PR #2065, opened 2023-09-10
Jmm/cog using history

JeanMarie-TTS:
## PR checklist: submitters

- [x]   Link to an issue if possible. If there’s no issue, describe what your branch does. Even if there is an issue, a brief description in the PR is still useful. #2022 
- [ ]   List any special steps reviewers have to follow to test the PR. For example, adding a local environment variable, creating a local test file, etc. _Many tests now spit out a warning that cog/over computation is being initiated for a sac with no fed awards. These tests need to be revisited at some point as this is an invalid state_

## PR checklist: reviewers


backend/audit/signals.py
```diff
@@ -0,0 +1,13 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import SingleAuditChecklist
+from support import cog_over
+
+
+@receiver(post_save, sender=SingleAuditChecklist)
+def fac_post_process(sender, instance, created, **kwargs):
```
timoballard, line 9: Wondering about whether this signal can/should live in the `support` app, so that all of the logic associated with `cog` assignment lives over there and nothing in `audit` needs to be responsible for triggering `cog` processing

backend/support/cog_over.py
```diff
@@ -0,0 +1,213 @@
...
+DA_THRESHOLD_FACTOR = 0.25
+
+
+def compute_cog_over(sac: SingleAuditChecklist):
+    """
+    Compute cog or oversight agency for the sac.
+    Return tuple (cog_agency, oversight_agency)
+    """
+    if not sac.federal_awards:
+        print(
```
timoballard, line 24: we can change this to `logger.warn` so that this will show up in new relic if it's ever hit

backend/audit/signals.py
JeanMarie-TTS: Suggestion accepted and implemented

backend/support/cog_over.py
JeanMarie-TTS: Done

JeanMarie-TTS:
Implemented @jadudm 's recommendation to not allow nulls wherever possible.

timoballard (approved):

timoballard (approved):

PR #2079, opened 2023-09-12
Make data loading transparent and repeatable

jadudm:
This brings the container back that loads historic data for dev and test.

Ultimately, this all has to be replaced, becuase the data will change over the course of the next three months.

The current load solution does not work well locally for two reasons:

1. It loads the container stack in the background, making it hard to debug via the container logs
2. It is not repeatable, because `load` assumes that it will only ever run once.

This brings back the container that loads the data. It works as part of the stack, and was designed to work repeatably. While it is *different* from how the GH action runner behaves, this works every time in the local stack.

We will treat aligning the deploy and local environments as tech debt. This is blocking devs from doing their work. This unblocks devs.

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/docker-compose.yml
+++ b/backend/docker-compose.yml
@@ -19,6 +19,17 @@ services:
       retries: 10
 
   #---------------------------------------------
+  # Historic data
+  #---------------------------------------------
+  historic-data:
+    image: ghcr.io/gsa-tts/fac-historic-public-csvs/load-historic-public-data:20230912
+    depends_on:
+      db:
+        condition: service_healthy
+    environment:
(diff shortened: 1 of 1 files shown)


backend/docker-compose.yml
```diff
@@ -18,6 +18,17 @@ services:
       timeout: 5s
       retries: 10
 
+  #---------------------------------------------
+  # Historic data
+  #---------------------------------------------
+  historic-data:
```
asteel-gsa, line 24:
```suggestion
  historic-data:
```
This needs to be added to both docker compose files

asteel-gsa (approved):

Issue #1992, opened 2023-08-31
General info user interaction bugfix


jadudm:
# At a glance

[comment]: # "Begin with a short summary so intent can be understood at a glance."

[comment]: # "In order to: some objective or value to be achieved"
[comment]: # "as a: stakeholder"
[comment]: # "I want: some new feature"

**In order to** track my submission accurately in the checklist 
**as a** submitter
**I want** incomplete work on the forms to be reflect accurately in the checklist

# Acceptance Criteria

We use [DRY](https://docs.behat.org/en/latest/user_guide/writing_scenarios.html#backgrounds) [behavior-driven development](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications) wherever possible.

[comment]: # "ACs should be clearly demoable/verifiable whenever possible."
[comment]: # "Given: the initial context at the beginning of the scenario"
[comment]: # "when: the event that triggers the scenario"
[comment]: # "then: the expected outcome(s)"
[comment]: # "Repeat scenarios as needed, or repeat behaviors and lists within a scenario as needed."

[comment]: # "The scenario should be a short, plain language description."
[comment]: # "Feeling repetative? Apply the DRY (Don't Repeat Yourself) principle!"

### Scenario: An auditee bails out right after assigning accesses

**Given** that I just assigned accesses
**when**  I submit them
...

[comment]: # "Each task should be a verifiable outcome"
```[tasklist]
### then...
- [x] I should be redirected to the checklist.
```

### Scenario: A  auditee enters part of the general info form

**Given** that I don't have all the info
**when**  I finish my partial form
...

[comment]: # "Each task should be a verifiable outcome"
```[tasklist]
### then... 
- [x] At the top of the form, "All fields are required before a final submission can be made."
- [x] The form fields I have completed undergo light front-end validation
- [x] I can go to the bottom and save my (partial) work and return to the checklist
- [x] The checklist runs the validator; if an error returns, the form is considered incomplete. If the validator returns `OK`, then it should be marked complete.
```

The light validation means:

```[tasklist]
### Tasks
- [x] (Break state list out from Jsonnet into a stand-alone JSON doc)
- [x] The state is a drop-down. Use above list for populating.
- [x] EIN validates as a 9-character string made up only of [0-9]
[...]

PR #2085, opened 2023-09-12
General Information - Allowed to be Partially Complete

jperson1:
# General Information - Allowed to be Partially Complete

Issue: https://github.com/GSA-TTS/FAC/issues/1992

## Changes:
1. The general information form can now be submitted partially complete.
    * HTML reflects this difference in that nothing is required anymore, and many error messages were struck.
    * JS reflects this difference in that no fields are made required, and null value errors have been removed.
    * The django form no longer requires all the fields, and has some regex validators to limit 400 responses from the gen info page
    * The validator (jsonnet schema) still runs, but it allows most fields to be empty.

2. The submission checklist page runs a different validator on the general information section
    * The new schema is `GeneralInformationComplete.jsonnet`, and it's similar to the other one, but it allows no empty data and requires all necessary fields.
    * When visiting the submission checklist page, you'll be able to tell when the form hasn't been totally completed.

3. Many test changes:
    * Some now need to be looking for errors in the page, rather than 400 responses
    * Some need to handle dates a little differently
    
4. Some other things:
    * The sidenav (top left) is fixed. It now has three sections that match up with the three major form groups. That broke at some point as we were updating the form. 

## How to test:
1. Switch to this branch and start a clean run. This will clear old submissions, which may or may not now be broken. 
2. Start a new submission, get to the general information form.
3. Verify that:
    * You can exit at this point and see your submission in the homepage
    * Visiting the checklist for that submission displays the form as incomplete
    * Navigating to the general information form still autocomplete some sections (fiscal year dates, UEI)
4. Fill out some fields and submit the form. Verify that:
    * The form still shows as incomplete
    * Those fields saved properly
    * Those fields auto-fill when revisiting the form.
5. Fill out some fields very wrong on purpose. Try to submit, and verify that these fields give reasonable error messages:
    * EINs (non-numbers, too long/short values)
    * Ciities (numbers, special characters like underscores)
    * Dates (not MM/DD/YYYY)
    * Phone numbers
6. Finish the form, submit, and verify that it displays as complete.
7. Play around with submitting/not submitting certain fields to try and find a case where it shows as complete when it isn't, or the opposite. 
8. Step through to cross-al and verify that general information gives errors if it isn't done, and gives no errors if it is done.

## Screenshots:
Submission progress page, incomplete general information:
image

Submission progress page, completed general information (the same as before):
image

General information with errors:
image

## PR checklist: submitters

## PR checklist: reviewers


tadhg-ohiggins:
I might have missed this part, but what indicates to the user the missing fields in the case where General Information has no errors but isn't finished?

jperson1:
> what indicates to the user the missing fields in the case where General Information has no errors but isn't finished?

That's a great point. Nothing gives the user any particular missing field. We just have an "All fields are required" message up top. The only way the user knows the form is incomplete is because the link in the checklist doesn't turn green, and because cross-val will let them know directly if they get to it without completing the form. 

A potential quick fix is to add a conditional red asterisk (or other indicator) to the form so they can more easily see which fields are missing. Or, wrap them up and present the missing fields on the checklist or in the form somewhere.

jperson1:
@tadhg-ohiggins From what I can tell, the validate will pass an error on the first field it runs into a problem on, but not on all the missing fields. For example, leaving the auditee address and city blank would throw an error on the address but not the city. Even if it's not friendly, this will still be great to have in corss-val, so unexpected validation issues can get caught. 

I wonder if a helper in submission_progress_check could walk it and check for missing fields, then display in the checklist? That wouldn't be as complex or comprehensive as the schema errors.

jperson1:
@tadhg-ohiggins I've merged your cross val change into this branch. Since the "GeneralInformation.schema.json" version in this branch is super permissive, I have it using the "GeneralInformationComplete.schema.json". In whatever way the required fields set is merged in, I assume we'll want the more strict schema running in cross validation. 

It works great! Like you said, it's not user friendly, but it's certainly better than giving no errors.
image

tadhg-ohiggins (approved):

backend/audit/cross_validation/submission_progress_check.py
```diff
@@ -114,3 +120,26 @@ def get_num_findings(award):
...
+    """
+    Given a base "progress" dictionary and the general_info object from a submission,
+    run validations to determine its completeness. Then, return a dictionary with
+    "general_information" as the key and the progress as the value.
+    """
+    try:
+        is_general_info_complete = bool(
+            validate_general_information_complete_json(general_info)
+        )
+    except Exception as err:
```
jadudm, line 135:
```suggestion
    except ValidationError as err:
```

backend/audit/cross_validation/submission_progress_check.py
```diff
@@ -114,3 +120,26 @@ def get_num_findings(award):
...
+        print("General information form not complete based on conditions:", err)
```
jadudm, line 136:
```suggestion
```

backend/report_submission/views.py
```diff
@@ -202,12 +211,59 @@ def post(self, request, *args, **kwargs):
             raise PermissionDenied("You do not have access to this audit.") from err
         except ValidationError as err:
             message = f"ValidationError for report ID {report_id}: {err.message}"
-            print(message)
             raise BadRequest(message)
         except LateChangeError:
             return render(request, "audit/no-late-changes.html")
+        except Exception as err:
+            message = f"Unexpected error in GeneralInformationFormView post. Report ID {report_id}: {err}"
```
jadudm, line 218:
```suggestion
            message = f"Unexpected error in GeneralInformationFormView post. Report ID {report_id}"
```

jadudm (approved):
LGTM.

PR #2091, opened 2023-09-13
Bump docker action versions

asteel-gsa:
Resolves #2087 and #2088


jadudm (approved):
Passes tests and deploys. No local testing possible. LGTM.

Issue #1940, opened 2023-08-24
FAC - Bug: Notes to SEFA workbook does not allow tables 


Leighdiddy:
In `cap_text` and `audit_findings_text`, we have a column that says `Did Text Contain a Chart or Table?`

The workbook `notes_to_sefa` is missing this column. Users need that column. 

```[tasklist]
### Tasks
- [x] Add this column to the workbook sonnet spec for `notes_to_sefa`
- [x] Regenerate the workbook XLSX doc for visual testing
- [x] Add the column to the mapping (which converts the workbook to JSON)
- [x] Add the column to the validation for the workbook
- [x] Run tests. The tests should fail at this point.
- [x] Update the `notes_to_sefa` test object in the test harness to include this field. (Tests should fail until you do, if you update the validation first)
- [x] Run tests; they should now pass.
- [x] Update the dissemination model for `notes`
- [x] Update the dissemination code in `IntakeToDissemination` to include this field
- [x] Update the API view for `v1_0_0` to disseminate this field as part of `notes`
- [ ] Update the workbooks on the static site so that this new template is distributed to users
- [x] Generate/update workbooks for Cypress testing
```

To generate new workbooks, we will need... to fix the workbook generator.

Original bug below.

-----

### Describe the bug

Open test user feedback, Notes to SEFA workbook does not allow tables to be inserted, see screen grab for explanation: 
<img width="719" alt="Screenshot 2023-08-24 at 5 11 27 PM" src="https://github.com/GSA-TTS/FAC/assets/119363060/0bbc3acb-c8ed-4030-af68-e8b7771056f4">

### Steps to reproduce the bug

see above

### Expected Behavior

see above 

### Code of Conduct

- [X] I agree to abide by the [Digital.gov Community Guidelines](https://digital.gov/communities/community-guidelines/) and the [TTS Code of Conduct](https://handbook.tts.gsa.gov/code-of-conduct/). Respect your peers, use plain language, and be patient.
- [X] I checked the [current issues](https://github.com/GSA-TTS/FAC/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

jadudm:
I thought that the Notes to SEFA had a column that asked if a table was included. And, my understanding was that if they said "yes," then the table would be in the PDF document, not in the workbook. That is, I thought:

1. The user indicates there is a table,
2. The table is expected to be found in the PDF.

[...]

PR #2135, opened 2023-09-15
Add contains_chart_or_table column to Notes to SEFA

danswick:
fixes https://github.com/GSA-TTS/FAC/issues/1940

## PR checklist: submitters

## PR checklist: reviewers


backend/dissemination/docs.py
```diff
@@ -102,6 +102,7 @@
 ein_list = "Data sources: SF-SAC 1997-2000: I/5/a; SF-SAC 2001-2003: I/5/a; SF-SAC 2004-2007: I/5/a; SF-SAC 2008-2009: I/4/a; SF-SAC 2010-2012: I/4/a; SF-SAC 2013-2015: I/4/a; SF-SAC 2016-2018: I/4/a; SF-SAC 2019-2021: I/4/a; SF-SAC 2022: I/4/a Census mapping: GENERAL, EIN (AND) Data sources: SF-SAC 2001-2003: I/5/c; SF-SAC 2004-2007: I/5/c; SF-SAC 2008-2009: I/4/c; SF-SAC 2010-2012: I/4/c; SF-SAC 2013-2015: I/4/c; SF-SAC 2016-2018: I/4/c; SF-SAC 2019-2021: I/4/c; SF-SAC 2022: I/4/c Census mapping: EIN INFO, EIN"
 charts_tables_captext = "Census mapping: CAPTEXT, CHARTSTABLES"
 charts_tables_findingstext = "Census mapping: FINDINGSTEXT, CHARTSTABLES"
+charts_tables_note = "Census mapping: "  # TODO: fill in census mapping here
```
danswick, line 105: Needs to be documented.

backend/dissemination/migrations/0003_note_contains_chart_or_table.py
```diff
@@ -0,0 +1,22 @@
...
+        ("dissemination", "0002_additionalein"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="note",
+            name="contains_chart_or_table",
+            field=models.TextField(
+                default="",
+                help_text="Census mapping: ",
```
danswick, line 17: I'm not entirely sure what our best practices should be for adding new columns / assigning defaults.

We'll also want to regenerate this migration once the help_text is updated.

tools/workbook-generator/main.py
danswick, line 1: @jadudm I'm not sure if this actually needs a map since the field names are the same?

backend/dissemination/migrations/0003_note_contains_chart_or_table.py
jadudm: This should not have a default, as it should be required by the schema, and there should always be a Y/N in this column. I think that would eliminate the need for a default?

tools/workbook-generator/main.py
jadudm: Ignore that file. It should be removed from `main` as part of bringing in the version of this that became the E2E code. So, don't stress about this code. It wants to go away.

jadudm (approved):
LGTM.

PR #2156, opened 2023-09-18
Initial rework of cognizant/oversight agency assignment.

tadhg-ohiggins:
Dissemination
Requires a lot of info.
Do we have it all?

-----

## Overview

The standard checklist isn’t really applicable here (but will be for the PR to `main`).

When we were reviewing #2065, we discovered that not all of the data was making it to dissemination. This works stems from trying to fix that and then to get the tests to pass.

I opted to move the dissemination step into a `SingleAuditChecklist` method, rather than having it be a signal that handles the `post_save` event on `SingleAuditChecklist`. It’s now necessary to call `sac.disseminate()` after `sac.save()`. In the application, this happens in `audit/views.py`, in the submission step view.

After that change and some others, submission tests failed, because if we initiated the dissemination code from our tests (which exercise the submission step view), that code would look for things that weren’t present in the test data. A number of changes address that.

Avoiding circular imports was an issue here, requiring moving code around.

I wasn’t testing the `run_2022` management command when I did the initial refactor, and I’m not sure I have the setup steps for it fully correct; right now, when I run it locally on this branch it seems to report many more cog assignment mismatches than it apparently should.

In addition, it had a number of errors around not being able to find fields when translating from `SingleAuditChecklist.general_information` to `General`; I don’t know why these only showed up after the changes I made, since I don’t see how my changes would alter the data arriving at that point. I made a few changes to that, primarily making a number of fields optional (or populating them with empty strings if they were absent); that’s probably the part of this that needs the most attention.
While making those changes I also refactored `audit.intake_to_dissemination.IntakeToDissemination.load_general`, but kept the same per-field approaches to getting data out of `SingleAuditChecklist.general_information` and `SingleAuditChecklist.audit_information`.

-----

### `audit/fixtures/json/federal-awards--test0001test--simple-pass.json`

I increased these just in case tests would fail for having too small a number in either or both of them.

### `audit/intake_to_dissemination.py`

Remove references to `SingleAuditChecklist` to prevent circular-import problems.

Previously-mandatory fields that are now optional:

#### `load_notes`

*   `notes_to_sefa`—will be an empty dict if missing in `load_notes()`.

#### `load_general`

*   `audit_information`—will be an empty dict if missing.
*   `auditee_certification`—will be an empty dict if missing.
*   `auditee_certify_name` and `auditee_certify_title`—will now fail one level deeper if `auditee_certification` is missing.
*   `cognizant_agency` and `oversight_agency`—will now be empty strings if missing.

### `audit/models.py`

*   Add `disseminate` and `assign_cog_over` methods to `SingleAuditChecklist`, moving that logic out of `support`.
*   Add `DISSEMINATED` status.
*   Add `transition_to_disseminated` method.

### `audit/test_views.py`

Various changes to provide sufficient data so that `sac.disseminate` doesn’t break tests when invoked.

### `audit/models.py`

Add explicit invocation of `sac.disseminate` after `sac.save` in submission step.

### `support/admin.py`

I removed the `sac` property from `CognizantAssignment`, so now things that previously relied upon that need to look up the SAC by `report_id`, as done here.o

### `support/cog_over.py`

*   Correct import order.
*   Alter `compute_cog_over` to take as arguments the fields from a `SingleAuditChecklist` that it needs, rather than the entire object.
*   Remove `propogate_cog_update`; this logic is now in-line in `support/signals.py`, primarily for circular import reasons.
*   Alter `record_cog_assignment` to take as arguments the fields from a `SingleAuditChecklist` that it needs, rather than the entire object.

### `support/management/commands/run_2022.py`

Add `sac.disseminate()` after `sac.save()`.

### `support/models.py`

Remove `sac` property of `CognizantAssignment`, primarily for circular import reasons.

### `support/signals.py`

*   Move what was `propogate_cog_update` into `post_cog_assignment`.
*   Remove `fac_post_process` now that we’re using explicit `sac.disseminate()` calls.

### `support/test_cog_over.py`

Alter tests to support all of the other changes.

### DIFF
--- a/backend/audit/models.py
+++ b/backend/audit/models.py
@@ -167,14 +167,14 @@ class SingleAuditChecklist(models.Model, GeneralInformationMixin):  # type: igno
         Cognizant/Oversight agency assignment followed by dissemination
         ETL.
         """
-        try:
-            self.assign_cog_over()
-            intake_to_dissem = IntakeToDissemination(self)
-            intake_to_dissem.load_all()
-            intake_to_dissem.save_dissemination_objects()
-            return None
-        except Exception as err:
-            return {"error": err}
+        # try:
+        self.assign_cog_over()
+        intake_to_dissem = IntakeToDissemination(self)
+        intake_to_dissem.load_all()
+        intake_to_dissem.save_dissemination_objects()
+        # TODO: figure out what exceptions to catch here
+        # except Exception as err:
+        #     return {"error": err}
 
     def assign_cog_over(self):
         """
--- a/backend/audit/models.py
+++ b/backend/audit/models.py
@@ -227,6 +227,7 @@ class SingleAuditChecklist(models.Model, GeneralInformationMixin):  # type: igno
         AUDITEE_CERTIFIED = "auditee_certified"
         CERTIFIED = "certified"
         SUBMITTED = "submitted"
+        DISSEMINATED = "disseminated"
 
     STATUS_CHOICES = (
         (STATUS.IN_PROGRESS, "In Progress"),
@@ -235,6 +236,7 @@ class SingleAuditChecklist(models.Model, GeneralInformationMixin):  # type: igno
         (STATUS.AUDITEE_CERTIFIED, "Auditee Certified"),
         (STATUS.CERTIFIED, "Certified"),
         (STATUS.SUBMITTED, "Submitted"),
+        (STATUS.DISSEMINATED, "Disseminated")
     )
 
     USER_PROVIDED_ORGANIZATION_TYPE_CODE = (
@@ -465,8 +467,6 @@ class SingleAuditChecklist(models.Model, GeneralInformationMixin):  # type: igno
         the appropriate privileges will done at the view level.
         """
 
-        from audit.intake_to_dissemination import IntakeToDissemination
-
         self.transition_name.append(SingleAuditChecklist.STATUS.SUBMITTED)
         self.transition_date.append(datetime.now(timezone.utc))
         if self.general_information:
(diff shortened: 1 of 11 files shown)


backend/audit/models.py
```diff
@@ -160,6 +162,40 @@ def save(self, *args, **kwargs):
...
+        # try:
+        self.assign_cog_over()
+        intake_to_dissem = IntakeToDissemination(self)
+        intake_to_dissem.load_all()
+        intake_to_dissem.save_dissemination_objects()
+        # TODO: figure out what exceptions to catch here
+        # except Exception as err:
+        #     return {"error": err}
+
+    def assign_cog_over(self):
```
JeanMarie-TTS, line 179: I don't see why it's better to move this logic here. But it should work for now since we won't have to deal with overrides forat least another month. Hope you looked into the case where a Cog Assignment is manually overridden  - we then need to update SAC and Dissemination then as well.

backend/support/management/commands/run_2022.py
```diff
@@ -26,6 +26,7 @@ def handle(self, *args, **kwargs):
             sac = self.make_sac(gen)
             sac.submission_status = sac.STATUS.SUBMITTED
             sac.save()
+            sac.disseminate()
```
JeanMarie-TTS, line 29: run_2022 is just a proram to check our logic. We are not actually determining cog/over for 2022 as these were done by census.  I don't think we need calll disseminate here. In fact, we need to make sure we don't run this code in any env other than local, or perhaps dev.

timoballard (approved):

PR #2183, opened 2023-09-20
Use fixture for IntakeToDissemination.load_general test.

tadhg-ohiggins:
Since we have fixtures
We should use them where we can:
Fewer truth sources.

-----

Bases the General Information data in the `IntakeToDissemination.load_general` test on the JSON test fixture; would have caught [this](https://github.com/GSA-TTS/FAC/pull/2177/files).

## PR checklist: reviewers

- [ ]   Test-only change. If the tests pass, this PR should be good to go.


danswick (approved):
Looks good as soon as tests pass

PR #2203, opened 2023-09-21
Unshares the backups bucket so share happens cleanly every time

asteel-gsa:
Last PR on this.. promise... _maybe..._

TLDR: `cf share-service` cannot share a service that is already shared. `cf share-service` also does not have have `-f`


tadhg-ohiggins (approved):
Looks worth a try!

Issue #2224, opened 2023-09-22
FAC - Bug: Do not disseminate in run_2022 command 


gsa-suk:
### Describe the bug

run_2022 command is used for local testing of cog_over.  Do not disseminate this test data. 

### Steps to reproduce the bug

Run 'fac run_2022'.

### Expected Behavior

No test data in dissemination.

### Code of Conduct

- [X] I agree to abide by the [Digital.gov Community Guidelines](https://digital.gov/communities/community-guidelines/) and the [TTS Code of Conduct](https://handbook.tts.gsa.gov/code-of-conduct/). Respect your peers, use plain language, and be patient.
- [X] I checked the [current issues](https://github.com/GSA-TTS/FAC/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

Issue #2134, opened 2023-09-14
FAC - Bug: Fine tuning required for historic table queries.


gsa-suk:
### Describe the bug

In the Census historic tables (gen19, cfda19, gen22, cfda22, ...), there are multiple submissions with the same EIN and there are multiple submissions with the same UEI.  Querying these tables with (EIN, UEI) pair will result in the least number of multiple submissions / the least number of multiple DBKEYs, resulting in lesser cog re-assignments / overrides. 
Reference: https://docs.google.com/document/d/1SaBRMdr7HsHqHuDZ98oxsJOvQGQPlYjEmk7Zhky8mXM/edit

(DBKEY, EIN) combination is unique in Cognizant_Agencies_2021_2025 table.  Querying this table with 'EIN only' will provide incorrect results.  Also querying with 'UEI only' will provide incorrect results, since UEI is not present in most cases.
Reference: https://github.com/GSA-TTS/FAC/blob/main/backend/support/fixtures/census_baseline.csv

### Steps to reproduce the bug

Query the tables.

### Expected Behavior

Update the queries.

### Code of Conduct

- [X] I agree to abide by the [Digital.gov Community Guidelines](https://digital.gov/communities/community-guidelines/) and the [TTS Code of Conduct](https://handbook.tts.gsa.gov/code-of-conduct/). Respect your peers, use plain language, and be patient.
- [X] I checked the [current issues](https://github.com/GSA-TTS/FAC/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

PR #2225, opened 2023-09-22
Sk/2134

gsa-suk:
Tickets: [2134](https://github.com/GSA-TTS/FAC/issues/2134), [2224](https://github.com/GSA-TTS/FAC/issues/2224)

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/support/cog_over.py
+++ b/backend/support/cog_over.py
@@ -103,7 +103,7 @@ def determine_hist_agency(ein, uei):
 def get_dbkey(ein, uei):
     try:
         dbkey = CensusGen22.objects.values_list("dbkey", flat=True).get(
-            Q(ein=ein), Q(uei=uei) | Q(uei=None)
+            Q(ein=ein), Q(uei=uei)
         )[:]
     except (CensusGen22.DoesNotExist, CensusGen22.MultipleObjectsReturned):
         dbkey = None
@@ -114,7 +114,7 @@ def lookup_baseline(ein, uei, dbkey):
     try:
         cognizant_agency = CognizantBaseline.objects.values_list(
             "cognizant_agency", flat=True
-        ).get(Q(is_active=True) & ((Q(ein=ein) & Q(dbkey=dbkey)) | Q(uei=uei)))[:]
+        ).get(Q(is_active=True) & (Q(ein=ein) & Q(dbkey=dbkey)))[:]
     except (CognizantBaseline.DoesNotExist, CognizantBaseline.MultipleObjectsReturned):
         cognizant_agency = None
     return cognizant_agency
@@ -123,7 +123,7 @@ def lookup_baseline(ein, uei, dbkey):
 def get_2019_gen(ein, dbkey):
     gens = CensusGen19.objects.annotate(
         amt=Cast("totfedexpend", output_field=BigIntegerField())
-    ).filter(Q(ein=ein), Q(dbkey=dbkey) | Q(dbkey=None))
+    ).filter(Q(ein=ein), Q(dbkey=dbkey))
 
     if len(gens) != 1:
         return (len(gens), 0)
@@ -134,7 +134,7 @@ def get_2019_gen(ein, dbkey):
 def get_2019_cfdas(ein, dbkey):
     cfdas = CensusCfda19.objects.annotate(
         amt=Cast("amount", output_field=BigIntegerField())
-    ).filter(Q(ein=ein), Q(dbkey=dbkey) | Q(dbkey=None))
+    ).filter(Q(ein=ein), Q(dbkey=dbkey))
 
     if len(cfdas) == 0:
         return None
--- a/backend/support/cog_over.py
+++ b/backend/support/cog_over.py
@@ -103,7 +103,7 @@ def determine_hist_agency(ein, uei):
 def get_dbkey(ein, uei):
     try:
         dbkey = CensusGen22.objects.values_list("dbkey", flat=True).get(
-            Q(ein=ein), Q(uei=uei)
+            Q(ein=ein), Q(uei=uei) | Q(uei=None)
         )[:]
     except (CensusGen22.DoesNotExist, CensusGen22.MultipleObjectsReturned):
         dbkey = None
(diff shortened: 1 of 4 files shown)


backend/support/cog_over.py
```diff
@@ -103,7 +103,7 @@ def determine_hist_agency(ein, uei):
 def get_dbkey(ein, uei):
     try:
         dbkey = CensusGen22.objects.values_list("dbkey", flat=True).get(
-            Q(ein=ein), Q(uei=uei) | Q(uei=None)
```
JeanMarie-TTS, line 106: This change appears incorrect. In G-FAC, we collect both UEI and EIN. In Census, UEI is optional. If we find a matching row based on EIN in Gen22 with UEI as nullm we should take it. We cannot restruct our search to both fields matching.

backend/support/cog_over.py
```diff
@@ -123,7 +123,7 @@ def lookup_baseline(ein, uei, dbkey):
 def get_2019_gen(ein, dbkey):
     gens = CensusGen19.objects.annotate(
         amt=Cast("totfedexpend", output_field=BigIntegerField())
-    ).filter(Q(ein=ein), Q(dbkey=dbkey) | Q(dbkey=None))
```
JeanMarie-TTS, line 126: This one too is incorrect for the same reason as the first one.

backend/support/cog_over.py
```diff
@@ -134,7 +134,7 @@ def get_2019_gen(ein, dbkey):
 def get_2019_cfdas(ein, dbkey):
     cfdas = CensusCfda19.objects.annotate(
         amt=Cast("amount", output_field=BigIntegerField())
-    ).filter(Q(ein=ein), Q(dbkey=dbkey) | Q(dbkey=None))
```
JeanMarie-TTS, line 137: Seems incorrect for the same reason.

backend/support/test_cog_over.py
```diff
@@ -342,6 +342,7 @@ def test_cog_assignment_with_uei_in_baseline(self):
         sac.general_information["auditee_uei"] = UEI_WITH_BASELINE
         baker.make(
             CognizantBaseline,
+            ein=sac.general_information["ein"],
```
JeanMarie-TTS, line 345: The intent here was to test the case where there is a matching UEI in the baseline regardless of the value of the ein. Adding an ein makes it the same as a different test case.

backend/support/cog_over.py
```diff
@@ -114,7 +114,7 @@ def lookup_baseline(ein, uei, dbkey):
     try:
         cognizant_agency = CognizantBaseline.objects.values_list(
             "cognizant_agency", flat=True
-        ).get(Q(is_active=True) & ((Q(ein=ein) & Q(dbkey=dbkey)) | Q(uei=uei)))[:]
```
JeanMarie-TTS, line 117: This change seems incorrect. When we lookup the baseline and find a matching UEI we should use it. This will be the case especially when we start processing AY24. Note that our goal is not to get run_2022 to better match census, but to get AY23 submissions and beyond right,

JeanMarie-TTS (changes requested):
I would not make the proposed change to cog_over and test_cog_over. These modules should be primarily for G-FAC data.
As for the changes to run_2022 itself, I will request @tadhg-ohiggins to review as he was involved with sacing  sac, etc and I have not studies this code indetail after the refactoring.

backend/support/management/commands/run_2022.py
```diff
@@ -26,7 +26,8 @@ def handle(self, *args, **kwargs):
             sac = self.make_sac(gen)
             sac.submission_status = sac.STATUS.SUBMITTED
             sac.save()
-            sac.disseminate()
+            if not sac.cognizant_agency and not sac.oversight_agency:
```
tadhg-ohiggins, line 29: This change is correct. We don't want to run the entire `sac.disseminate()` process, we just want to run `sac.assign_cog_over()` if appropriate.

backend/support/cog_over.py
gsa-suk: This is correct based on the findings and also the discussions in slack and meetings.
https://docs.google.com/document/d/1SaBRMdr7HsHqHuDZ98oxsJOvQGQPlYjEmk7Zhky8mXM/edit

backend/support/cog_over.py
gsa-suk: This is correct based on the findings and also the discussions in slack and meetings.
https://docs.google.com/document/d/1SaBRMdr7HsHqHuDZ98oxsJOvQGQPlYjEmk7Zhky8mXM/edit

backend/support/cog_over.py
gsa-suk: This is correct based on the findings and also the discussions in slack and meetings.
https://docs.google.com/document/d/1SaBRMdr7HsHqHuDZ98oxsJOvQGQPlYjEmk7Zhky8mXM/edit

backend/support/cog_over.py
gsa-suk: Based on all the findings and discussions, with the current Census tables, (EIN, DBKEY) is the only unique combination.  For AY23 and beyond also, UEI only cannot be used based on the findings.  When the unique identifier is defined for AY23 and beyond, the queries need to be updated.

https://docs.google.com/document/d/1SaBRMdr7HsHqHuDZ98oxsJOvQGQPlYjEmk7Zhky8mXM/edit

backend/support/cog_over.py
gsa-suk: Based on discussions, '| Q(uei=None)' will be used to select entities that have unique ein, but no uei.

backend/support/test_cog_over.py
gsa-suk: Test case is updated handle 'uei only' case.

backend/support/cog_over.py
gsa-suk: Based on discussions, for AY23 and beyond, (EIN, UEI) will be used as the unique identifier for an entity.  Query is updated accordingly.

JeanMarie-TTS (approved):
LGTM

Issue #2101, opened 2023-09-13
Create a form for tribal entities to indicate public/Fed-only status.


jadudm:

danswick:
Per chat with @Carley-Sullivan and @jadudm, it sounds like only **auditee certifying officials** should be able to submit this form. Maybe we should _also_ limit this form's visibility to auditee certifying officials to prevent confusion for viewers/editors with other account types.

danswick:
Next steps: validate against schema after it merges.

danswick:
Merged into main!

PR #2238, opened 2023-09-25
Tribal Audit Consent Form

jperson1:
# Tribal Audit Consent/Tribal Data Release Form

Issue: https://github.com/GSA-TTS/FAC/issues/2101

## Changes:
1. Tribal audit consent form
    * The URL `/audit/tribal-data-release/REPORTID`
    * The Django view (its own file in viewlib) that gets the object for display and saves the object on post
    * The HTML for the form
    * The form link in the submission checklist, which is static and does not rely on `submission_progress_check` or anything. 

## How to test:
1. Switch to this branch and run normally
3. Start a submission, or check out an old one
4. Go to the submission checklist and navigate to the tribal audit consent form
5. Fill it out and submit
6. Verify that the object is in the DB with proper variable names

## Screenshots:
Form:
image

Checklist:
image

## PR checklist: submitters

## PR checklist: reviewers


eazapata23 (approved):
LGTM, great work!

backend/audit/templates/audit/submission_checklist/submission-checklist.html
```diff
@@ -60,6 +60,18 @@
...
                     
+                    {% comment %} Tribal data release consent {% endcomment %}
+                    <li class="usa-icon-list__item padding-bottom-105" id="tribal_consent-completed">
+                        {% include './icon-list-icon.html' with completed=False %}
+                        <div>
+                            <p class="text-bold margin-0">
+                                <a class="usa-link"
+                                    href="{% url 'audit:TribalAuditConsent' report_id %}">Tribal data release</a>
+                            </p>
+                            <p class="margin-0">Auditees must opt in or opt out of making their reporting package publicly available.</p>
```
tadhg-ohiggins, line 71: @jperson1 Since this will show up as inactive (I assume) when non-tribal orgs are doing submissions, I think the language here should include something about how it only applies to tribal orgs, otherwise non-tribal submitters might think they need to access this form somehow.

backend/audit/templates/audit/submission_checklist/submission-checklist.html
danswick, line 1: I'm not sure if this is the right way to change a section's `inactive` status, but I started testing it out using a `section_block` here: https://github.com/GSA-TTS/FAC/pull/2285/files#diff-6db0c28c3db5698e21669a6603e3cdcf7d9d14d7d74c7ed99cdb723977932a4f

backend/audit/templates/audit/submission_checklist/submission-checklist.html
jadudm: Ideally:

1. This does not appear *at all* for non-tribal submissions.
2. It should appear inactive as part of tribal submissions, unless...
3. You are visiting the page as the auditee certifying official.

tadhg-ohiggins: Making it entirely hidden requires either adding a fourth state to the logic that drives the page or treating logic for the tribal submission form as entirely different from the other sections. Either are  possible but require more work and I suspect will make test coverage harder.
There are similar concerns for gating it on user role; I don't think we do that elsewhere on the page.

jadudm: I'd like to understand why it's hard, if we can do a short call at some point. But, showing this to 40K users when only 500 need it feels like a bad choice for 99% of our users.

tadhg-ohiggins: Per co-work conversation, the tribal data consent block should not determine its state using `submission_progress_check`, but rather have its own custom logic in the Django template.

jperson1:
Regression tests run:

https://github.com/GSA-TTS/FAC/actions/runs/6385288888

jperson1:
Once more, with feeling:
https://github.com/GSA-TTS/FAC/actions/runs/6395445013

PR #2246, opened 2023-09-25
Add Dev Backup and environment restores

asteel-gsa:
Since backups occur in prod, and we will be restoring `prod -> staging`, but not backing up `staging`, this has `if: ${{ inputs.environment == 'dev' }}` as a safeguard, incase someone changes the default value of `backup-database.yml`

This utilizes pass through vars in the bash scripts now, one for backing up and one for restoring.

`backup_database.sh` consumes a single value, and that is `['dev' or 'prod']`. This will make the call 
`./backup_database.sh ${{ env.space }}` where `env.space` == the enviroment name (dev or prod), 
making the name of the backup `$env-db-backup-$date.psql.bin` and `$env-media-backup-$date.tar` respectively.

`restore_database.sh` utilizes the same thing, but consumes two values. The name of the db backup, and the name of the media backup. you can obtain these values from `aws s3 ls s3://$BUCKET/static/` where `$BUCKET` is the cf name of the s3 bucket. (not the friendly name)

File names currently:
```bash
2023-09-25 19:16:11  356685603 dev-db-backup-2023-09-25-2314.psql.bin
2023-09-25 19:16:33  602050560 dev-media-backup-2023-09-25-2314.tar
```

`restore-database.yml` would expect `dev-db-backup-2023-09-25-2314.psql.bin` as `dbbackup-name`
and `dev-media-backup-2023-09-25-2314.tar` as `mediabackup-name`

TODO: Remove the sleep and replace with `cf curl v3/apps/guid/tasks` to check the status return from `RUNNING` to `SUCCEEDED` as a check before doing the `unbind-service`


.github/workflows/restore-database.yml
```diff
@@ -0,0 +1,59 @@
...
+      dbbackup-name:
+        required: true
+        type: string
+      mediabackup-name:
+        required: true
+        type: string
+
+jobs:
+  restore-database:
+    name: Restore ${{ inputs.environment }} database
```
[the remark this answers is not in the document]
asteel-gsa: We shouldn't have to target anything.

Meta shares the `backups` bucket to staging, so `prod and staging` have access to the bucket when you `bind-service`
Dev has its own bucket, which, we can share to `preview` so then `preview and dev` will have access to the bucket when you `bind-service`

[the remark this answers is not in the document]
JeanMarie-TTS: How do we specify that we are taking the backup from prod and applying to staging? Don't we need to say that the restore should actuallt run in the staging space?
One more thing. The dbbackup and mediabackup taken at a point in tims need to be restored together.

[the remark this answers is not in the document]
asteel-gsa: By the file name. `production-db-backup-2023-09-25-1714`

If you are in the `staging` space, and the backups bucket is shared to `staging` space, then you bind the bucket to the app, when you run `python manage.py dbrestore -i production-db-backup-2023-09-25-1714.psql.bin` it will look in the s3 for that file, and then apply it in that space.

The media and db are restored together in `restore_database.sh`

asteel-gsa:
I think we have settled on this for now.

This locks the `restore-database.yml` to myself and @JeanMarie-TTS for now, it will be further expanded to fac-admins after initial testing is done. Along with that, a codeowners entry was added for this file, as it is imperative, that if this file changes, we are alerted.

want to try and get the `cf curl` into this before I say its ready for review

.github/workflows/backup-database.yml
```diff
@@ -0,0 +1,57 @@
...
+          cf_username: ${{ secrets.CF_USERNAME }}
+          cf_password: ${{ secrets.CF_PASSWORD }}
+          cf_org: gsa-tts-oros-fac
+          cf_space: ${{ env.space }}
+          command: cf run-task gsa-fac -k 2G -m 2G --name pg_backup --command "./backup_database.sh ${{ env.space }}"
+
+      - name: Sleep for 5 minutes to ensure backup occurs
+        if: ${{ inputs.environment == 'dev' || inputs.environment == 'production' }}
+        run: sleep 5m
+        shell: bash
```
asteel-gsa, line 46:
```suggestion
      - name: Sleep for 5 minutes to ensure backup occurs
        if: ${{ inputs.environment == 'dev' || inputs.environment == 'production' }}
        run: sleep 5m
        shell: bash
```

This does need to be removed at some point, but for whatever reason, I am unable to use `cf auth` and then `cf curl /v3/apps/$(cf app gsa-fac --guid)/task` with my cf account. Doing so will return a json we can parse that retrieves the status of the task, and then that will loop until `status == 'SUCCEEDED'`

Taking this as an enhancement down the line

backend/backup_database.sh
```diff
@@ -1,7 +1,26 @@
...
+# Backup the database
+echo "Database Backup File: $1-db-backup-$date.psql.bin" >> "$log_file"
+python manage.py dbbackup -o "$1-db-backup-$date.psql.bin" >&1 >> "$log_file"
+
+# Backup the media contents
+echo "Media Backup File: $1-media-backup-$date.tar" >> "$log_file"
+python manage.py mediabackup -o "$1-media-backup-$date.tar" >&1 >> "$log_file"
+
+# Upload logfile to s3://backups/logs
+python manage.py fac_s3 backups --ls --tgt "$date" >> "$log_file"
```
JeanMarie-TTS, line 21: This ls command will list all objects that start with the specified date. Therefore, you should include this date as the first part of the names of the backup files, rather than starting with environment and the type of backup. You can check backup_local.sh  where the naming is different.

asteel-gsa: Ah! I see. I think we can remove this line then. If we want to ls, we can do it from cli, rather than in this backups file. The process of taking a backup doesn't really need to `ls` to see if there are backups there

JeanMarie-TTS: The intent is to record in the log some info about the nackup artifcat itself. Not sure why we would not do this. 
We should move the date to the front of the name.

JeanMarie-TTS (changes requested):

JeanMarie-TTS (approved):

asteel-gsa:
Up to date with `main` and ready for review

JeanMarie-TTS (commented):
Ready to merge. Need to do testing in `dev` asap so nothing breaks when we deploy to prod.

backend/backup_local.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "Backup date: $DATE" > $LOG_FILE
+python manage.py collect_metrics >> $LOG_FILE
+
+echo "DB Backup File: $DB_BACKUP_FILE" >> $LOG_FILE
+python manage.py dbbackup -o $DB_BACKUP_FILE >&1 >> $LOG_FILE
+# need to upload as my config backs up to local backup folder
+python manage.py fac_s3 $BUCKET --upload --src backup/$DB_BACKUP_FILE --tgt $DB_BACKUP_FILE
+
+echo "Media Backup File: $MEDIA_BACKUP_FILE" >> $LOG_FILE
+python manage.py mediabackup -o $MEDIA_BACKUP_FILE >&1 >> $LOG_FILE
```
danswick, line 19: Am i remembering correctly that there were problems with the third-party dependency `django-dbbackup` in the past? Are those no longer an issue?

asteel-gsa: No, we got it working. It was a disk/ram issue in the past. We have been using django dbbackup for a while now in prod

backend/br_ops/purge_backups.sh
```diff
@@ -0,0 +1,35 @@
...
+cfspace=$(cf target |tail -1|cut -d':' -f2|xargs)
+echo "Space: ${cfspace}"
+
+if [[ "$cfspace" != "dev" && "$cfspace" != "prodiction" ]]; then
+  echo "Error: Backups may only be purged fromr dev or production"
+  exit 1
+fi
+prefix=$(date '+%Y-%m-%d-%H%M')
+export PATH=/home/vcap/deps/0/apt/usr/lib/postgresql/15/bin:$PATH
+python manage.py fac_s3 backups --ls | awk '{print $1}' | grep "db\-backup" | sort -r > "$prefix"_backup_list.txt
```
tadhg-ohiggins, line 18: @danswick, @timoballard, and I think that we should move more of the logic here into the `fac_s3` management command, so that
* that command generates the backup list text file
* that command reads application config from settings.py in order to control the retention policy.

backend/br_restore.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Set basic runtime settings
+echo "Restoring from env: $1"
+echo "Restoring backup type: $2"
+echo "Restoring backup dated: $3"
+export PATH=/home/vcap/deps/0/apt/usr/lib/postgresql/15/bin:$PATH
+date=$(date '+%Y-%m-%d-%H%M')
+prefix="$date"_"$2"_"$1"
+log_file="$prefix"_restore_log.txt
+prefix="$3"_"$2"_"$1"
```
tadhg-ohiggins, line 11: Why are we redefining `prefix` here?

bin/ops/purge_backups.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
tadhg-ohiggins, line 1: Is this file the same as `backend/br_ops/purge_backups`? Does it need to be duplicated here for some reason?

tadhg-ohiggins (changes requested):
The main change we'd like to see is moving the retention policy logic into Python and putting the relevant configuration in `settings.py`.

bin/ops/purge_backups.sh
JeanMarie-TTS: Yes, the script in bin/ops needs to be removed

backend/br_restore.sh
JeanMarie-TTS: Looks like left-over code. Good catch. Will do another commit this afternoon.

backend/br_ops/purge_backups.sh
JeanMarie-TTS: I would do this in a later PR. This would involve some logic changes, and I'd rather see the current PR tested in `dev` first.

JeanMarie-TTS:
@asteel-gsa Since we are no longer planning to use `djangobackup` for backing up media files, we need to change some of the scripts included in this PR.

asteel-gsa:
@JeanMarie-TTS Feel free to modify how you see fit.

JeanMarie-TTS:
> @JeanMarie-TTS Feel free to modify how you see fit.

@alexbielen If you can give me the commands that we plan to use to backup and restore media files, I will update the scripts accordinglt.

Issue #2331, opened 2023-09-30
[HD]: ALN 27.011 no accepted?


jadudm:
### Zendesk link

https://fac-gov.zendesk.com/agent/tickets/50

### FAC components involved

- [ ] submission system (app.fac.gov)
- [X] workbooks
- [ ] static site (fac.gov)
- [ ] The helpdesk (zendesk)
- [ ] API (api.fac.gov)
- [ ] other

### What browser did the user report as using?

None

### What happened?

It looks like... we're not accepting ALNs with an agency prefix of 27?

![ALN Validation error](https://github.com/GSA-TTS/FAC/assets/38290/6ec18fb7-3f0b-4369-b28c-09967dfe1ae7)

jadudm:
Regarding the passthrough entity/direct award workbook:

```
'direct_or_indirect_award': {'is_direct': 'Y', 'entities': [{'passthrough_identifying_number': ' '}]}
```

There's a single space in the passthrough entity column. It's effectively invisible. But, it shows up.

jadudm:
See https://github.com/GSA-TTS/FAC/tree/2331-direct-indirect

PR #2380, opened 2023-10-04
2331 hd aln 27011 no accepted

sambodeme:
## Description
See #2331 
After investigation, it turned out that the application returns erroneous error reports due to the validation logic that is unable to properly capture the invalid field names from the error schema.

- Updated the code to allow for processing various error schemas.
- For testing purpose, launch the application in local and upload the federalAwards workbook attached to this ticket. Confirm that your error report is similar to the screenshot below. 

image

## PR checklist: submitters

## PR checklist: reviewers


backend/audit/excel.py
```diff
@@ -1,824 +1,863 @@
...
+    return None
+
+
+def _extract_validation_field_name(error):
+    try:
+        # Parse the input data as JSON
+        data_dict = error.schema
+        # Check if the input data is in the format { ... 'not': {'required': ['field_name']}}
+        if "not" in data_dict and "required" in data_dict["not"]:
+            field_name = data_dict["not"]["required"][0]
```
danswick, line 754: Are there any future cases where we'll have a list of not-required entries? It doesn't look like there are any cases where we have more than one field name in `{... 'not': {'required': [...]`, but we might want to keep this in mind just in case.

danswick (approved):
Tested locally and was able to get the error table listed in the PR description. 

Left one comment inline, but it's not a blocker for merging.

backend/audit/excel.py
sambodeme: We could be aggressive here and update the logic to account for a list of `not required` fields

Issue #2422, opened 2023-10-05
FAC - Bug: date_assigned column is not loading properly in CognizantBaseline table, change source to CENSUS / GSAFAC


gsa-suk:
### Describe the bug

1. date_assigned column in CognizantBaseline table is always showing current date after seed_cog_baseline.
2. Replace 'Census' with 'CENSUS' in source column of CognizantBaseline table.
3. Replace 'GSA' with 'GSAFAC' for GSA cog assignments.

### Steps to reproduce the bug

Run seed_cog_baseline and check data in CognizantBaseline table.

### Expected Behavior

There will be various values in the date_assigned column.
source column will contain 'CENSUS'. 

### Code of Conduct

- [X] I agree to abide by the [Digital.gov Community Guidelines](https://digital.gov/communities/community-guidelines/) and the [TTS Code of Conduct](https://handbook.tts.gsa.gov/code-of-conduct/). Respect your peers, use plain language, and be patient.
- [X] I checked the [current issues](https://github.com/GSA-TTS/FAC/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

PR #2445, opened 2023-10-10
Sk/2422 cog date source

gsa-suk:
https://github.com/GSA-TTS/FAC/issues/2422

## PR checklist: submitters

## PR checklist: reviewers


backend/support/management/commands/seed_cog_baseline.py
```diff
@@ -71,7 +71,7 @@ def update_cogbaseline_w_csv(filename):
                 & (df["uei"] == cogbaseline_inactive.uei)
             )
         ]
-    save_df_to_cogbaseline(df, "Census")
+    save_df_to_cogbaseline(df, "CENSUS")
```
JeanMarie-TTS, line 74: We should remove the second parameter in the function call as we are passing the same value in all calls to this method. The constant can be directly embedded inside the method.

backend/support/models.py
```diff
@@ -28,7 +29,7 @@ class CognizantBaseline(models.Model):
     date_assigned = models.DateTimeField(
         # help_text = "Time when the cog agency was assigned to the entity",
         null=True,  # allow nulls in case history has nulls
-        auto_now_add=True,
+        default=timezone.now,
```
JeanMarie-TTS, line 32: Since this is addressing a bug that was discovered, we should include a test case to make sure the fix is resolving the bug.

backend/support/models.py
```diff
@@ -28,7 +29,7 @@ class CognizantBaseline(models.Model):
     date_assigned = models.DateTimeField(
         # help_text = "Time when the cog agency was assigned to the entity",
```
JeanMarie-TTS, line 30: Not sure why or when this was commented out. I would prefer to keep it in.

backend/support/signals.py
```diff
@@ -23,7 +23,10 @@ def post_cog_assignment(sender, instance, created, **kwargs):
             baseline.is_active = False
             baseline.save()
         CognizantBaseline(
-            ein=ein, uei=uei, cognizant_agency=cognizant_agency, source="GSA"
+            ein=ein,
+            uei=uei,
+            cognizant_agency=cognizant_agency,
+            source="GSAFAC",
```
JeanMarie-TTS, line 29: GSAFAC is the default. So, we should not be explicitly assigning it.

JeanMarie-TTS (changes requested):

backend/support/models.py
gsa-suk: run_2022 command confirms that this fix is working.

gsa-suk: test_cog_over already has test cases.  The CognizantBaseline table is updated in signals.py.  The signal event is asynchronous and hence the date cannot be checked in test_cog_over.

JeanMarie-TTS (approved):

PR #2456, opened 2023-10-11
Tadhg/ssh access issue template

tadhg-ohiggins:
“Open the blast doors!”
Then do things until they say:
“Quick! Close the blast doors!”

-----

We need a quick way to log that we enabled production ssh access for a limited period; this issue template is meant for that purpose.


danswick (approved):
This is great. Thanks, @tadhg-ohiggins!

PR #2483, opened 2023-10-12
Jmm/resolve cog baseline bug

JeanMarie-TTS:
## PR checklist: submitters

## PR checklist: reviewers

## What does the PR fix
When running `run_2022` locally, noticed that a new entry was being created in CognizantBaseline for every assugnment of a cognizant_agency. We should not beadding a new entry if there is already a valid entry in this table.
The test cases to test this scenario were not doing the right thing and hence tests were passing.
Updated the tst case to test this consition. Re-ran `run_2022` and it creates fewer entries now as it should be.
Made minor tweaks to admin.py to be able to see the data better on the admin screen. This change is not required to fix he bug, but was useful in resolving it.


backend/support/signals.py
```diff
@@ -14,23 +14,41 @@ def post_cog_assignment(sender, instance, created, **kwargs):
     have on other tables.
     """
     if created:
+        cognizant_agency = instance.cognizant_agency
         sac = SingleAuditChecklist.objects.get(report_id=instance.report_id)
-        cognizant_agency = sac.cognizant_agency
+        sac.cognizant_agency = cognizant_agency
+        sac.save()
```
gsa-suk, line 20: Lines 19 and 20 (saving cognizant_agency to sac) had to be moved into assign-cog_over function in audit/models.py to handle asynchronous signal so disseminate and run_2022 will work.

gsa-suk: @tadhg-ohiggins is more familiar with the disseminate related issue.

JeanMarie-TTS: I don't think there is a need to update the cog agency in models.py, but I did not remove it as it won't break anything. 
Will request @tadhg-ohiggins  to review the PR.

tadhg-ohiggins: `sac.disseminate()`, which is called when a submission is complete on the submission side, includes a call to `sac.assign_cog_over()`, which:

*   Calls `support.cog_over.compute_cog_over()`.
*   Saves the SAC instance after adding the computed info.
*   Calls `support.cog_over.record_cog_assignment()`.

`support.cog_over.record_cog_assignment()` then creates a new `CognizantAssignment` instance… which then invokes this code via signal and assigns the value of the cognizant agency again, and saves the SAC again.

In the case where we’re assigning cog/over as a result of hitting the final part of the submission process, this is redundant; we’re passing the same value to `record_cog_assignment` as gets passed to this signal, so the SAC already has that cognizant agency value and has already been saved.

In the case where we are not coming from the submission side but are handling the creation of a new `CognizantAssignment` from elsewhere, this may work as intended and not be redundant.

I don’t think we should do the redundant `sac.save()` because that could cause hard-to-track-down problems. Other options include:

*   Send a keyword argument of some kind to `CognizantAssignment` when it’s created that we can look for in this function to determine whether or not we need to invoke `sac.save()`.
*   Remove the signal and instead override the `save_model` [Django admin method](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#modeladmin-methods) in `support.models.CognizantAssignment`. That way we know we’re in the scenario where the assignment is being made via the admin interface and not as part of submission.

I suggest the second option.

tadhg-ohiggins (changes requested):

backend/support/signals.py
JeanMarie-TTS: @tadhg-ohiggins  The problem with the second option is that it leads to curcular import dependencies.
I would keep the save in the signal and remove it from other places.

I think a meeting may be a better way to talk about this.

danswick:
@JeanMarie-TTS can we catch this up with main so @tadhg-ohiggins's PR is more reviewable?

backend/support/signals.py
jadudm: Was this conversation resolved? Or, is this part of the work that y'all might do in terms of debt payment/cleanup?

JeanMarie-TTS:
@danswick  Merfed with `main` as requested. All tests passed.

backend/support/signals.py
tadhg-ohiggins: @jadudm I think my PR resolves this; see https://github.com/GSA-TTS/FAC/pull/2616#issuecomment-1800887220

danswick (approved):

tadhg-ohiggins (approved):

PR #2517, opened 2023-10-16
Remove stray print statements

tadhg-ohiggins:
Debugging is hard
And `print` is necessary
But not when we’re live.

-----

Remove some stray `print` statements.


jadudm (approved):
LGTM.

Issue #2499, opened 2023-10-13
Activate Wrap Text in Workbook


sambodeme:
### Tasks

```[tasklist]
- [ ] Activate Wrap Text in all workbooks
- [ ] Increase column height in all workbooks
- [ ] Update workbook versions
```

PR #2518, opened 2023-10-16
2499 activate wrap text in workbook

sambodeme:
For details, see #2499 

## PR checklist: submitters

## PR checklist: reviewers


backend/schemas/scripts/render.py
```diff
@@ -499,6 +501,46 @@ def set_open_range_and_single_cell_widths(ws, sheet):
...
+        if ws.title in ("Form", "AdditionalNotes"):
+            for named_range_name in wb.defined_names:
+                activate_wraptext_for_named_range(wb, ws, named_range_name)
+
+
+def activate_wraptext_for_named_range(wb, ws, named_range_name):
+    """Activate wrapText for all cells within a named range in the worksheet."""
+    if named_range_name in wb.defined_names:
+        named_range = wb.defined_names[named_range_name]
+        if named_range.attr_text.startswith(
```
tadhg-ohiggins, line 535: Where are these string values, with the single quotation marks and the exclamation marks, coming from?

tadhg-ohiggins (approved):

backend/schemas/scripts/render.py
sambodeme: These are sheets named in the workbooks. For instance, for audit findings text, it is defined [here](https://github.com/GSA-TTS/FAC/pull/2518/files#diff-a5302ef12de61ad2603bca96749d4e35a95169b68e02115ee16cac8a4ef1df3bR5) and passthrough [here](https://github.com/GSA-TTS/FAC/pull/2518/files#diff-a5302ef12de61ad2603bca96749d4e35a95169b68e02115ee16cac8a4ef1df3bR104). Similarly, for CAP, it is[ here](https://github.com/GSA-TTS/FAC/pull/2518/files#diff-a5302ef12de61ad2603bca96749d4e35a95169b68e02115ee16cac8a4ef1df3bR104)

sambodeme: Thinking of a way to mitigate situations where the sheet name gets changed? We could store these specific sheet names in Sheets.libsonnet and read them from there.

jadudm (approved):
Looks good.

PR #2551, opened 2023-10-19
New Relic Debugging

asteel-gsa:
No core changes, remove port from full host and set explicitly as `NEW_RELIC_PROXY_PORT` based on NR support recommendation, and enable debug logging to send back to NR support.

**NOTE**: This does need to be actively monitored with `cf logs gsa-fac | grep "newrelic"` when a deployment occurs


jadudm (commented):
LGTM.

jadudm (approved):
LGTM.

sambodeme (approved):

PR #2554, opened 2023-10-19
Remove lighthouse and pa11y from test workflows

tadhg-ohiggins:
Warnings are useful
But constant failures less so.
Scratch Lighthouse for now.

-----

The Lighthouse and pa11y failures due to upstream problems are getting in the way, so disable Lighthouse and pa11y for now.

-----

This does nothing but remove the a11y test step, other checks should be unnecessary for this PR.


asteel-gsa (approved):
LGTM

PR #2563, opened 2023-10-20
Jadudm fix workbook generator for new checks

jadudm:
1. Fixes the workbook generator. 
    * Correctly sets timestamps (storing them in the `sac` as well as disseminating them) 
    * Creates the `cfda_key` column in the Federal Awards workbook, so that generated workbooks pass the new evals.
    * Sets the auditor cert object, so workbooks correctly move to the full "Disseminated" state and appear in the correct table when done.
 2. Fixes the cypress tests with the new frontend changes on the disseminated table.
 3. Removes `test_load_fixtures_federal_awards` from `test_commands.py`. This test uses a workbook that is not in good sync with the validations. Two other reasons: first, the `end_to_end` code does *more* than this test, exercising the same code, but more thoroughly. Second, the `test_should_pass` and `test_should_fail` code all runs many workbooks through, again providing better coverage against the new validation structures. This has, I think, become redundant.

e2e

## PR checklist: submitters

## PR checklist: reviewers


backend/audit/fixtures/single_audit_checklist.py
```diff
@@ -210,8 +210,8 @@ def _load_single_audit_checklists_for_user(user):
         if sac is None:
             # need to make this object
             sac = _create_sac(user, auditee_name)
-        if "post_create_callable" in item_info:
-            item_info["post_create_callable"](sac, user)
+        # if "post_create_callable" in item_info:
+        #     item_info["post_create_callable"](sac, user)
```
phildominguez-gsa, line 214:
```suggestion
```
Why not delete it?

backend/cypress/support/full-submission.js
```diff
@@ -155,8 +155,9 @@ export function testFullSubmission(isTribal, isPublic) {
     // The report ID should be found in the Completed Audits table
     cy.get('.usa-table').contains(
       'caption',
-      'The audits listed below have been submitted to the FAC for processing and may not be edited.',
-    ).siblings().contains('td', reportId);
+      // 'The audits listed below have been submitted to the FAC for processing and may not be edited.',
+      /audits are complete/
+      ).siblings().contains('td', reportId);
```
phildominguez-gsa, line 160:
```suggestion
      /audits are complete/
    ).siblings().contains('td', reportId);
```

backend/audit/test_commands.py
```diff
@@ -48,23 +47,31 @@ def test_load_fixtures(self, mock_scan_file):
...
-        self.assertTrue(files.first().sac.federal_awards)
+    # This is loading a workbook that does not validate.
+    # However, we now run tests that load and validate (or intentionally
+    # fail to validate) many workbooks.
+    #
+    # Those tests supercede this test.
+    # Therefore, I'm taking this test out, as it is more fragile
+    # than the `pass` and `fail` many-workbook-tests.
+
+    # @patch("audit.validators._scan_file")
```
phildominguez-gsa, line 58: My vote would be to delete this and document it in the PR desc (as opposed to commenting it out) but not a big deal.

phildominguez-gsa (changes requested):

phildominguez-gsa (approved):

backend/dissemination/workbooklib/federal_awards.py
```diff
@@ -213,10 +213,13 @@ def generate_federal_awards(dbkey, year, outfile):
     addls = _fix_addl_award_identification(Cfda, cfdas, dbkey)
     set_range(wb, "additional_award_identification", addls)
 
-    (prefixes, extensions) = _fix_pfixes(cfdas)
+    (prefixes, extensions, full_cfdas) = _fix_pfixes(cfdas)
     set_range(wb, "federal_agency_prefix", prefixes)
     set_range(wb, "three_digit_extension", extensions)
 
+    # We need a `cfda_key` as a magic column for the summation logic to work/be checked.
+    set_range(wb, "cfda_key", full_cfdas, conversion_fun=str)
```
sambodeme, line 221: In federal awards workbooks we use this formula `=IF(OR(B{0}="",C{0}=""),"",CONCATENATE(B{0},".",C{0}))`  to build `cfda_key` but according to the logic [here](https://github.com/GSA-TTS/FAC/pull/2563/files#diff-ff2dabbd25da7c9de0277ab7292bd1acd3afc8d8541e68a377bf9e736e659628R127) an item in `full_cfdas` is not garanteed to always be the concatenation of item in `prefixes` and item in `extensions` due to `else "000"`. This could lead to random bugs

PR #2573, opened 2023-10-20
Use a separate PG DB for Census data

JeanMarie-TTS:
## PR Summary

- Create a separate Postgres RDS to house data that we will get from Census
- Implement tests for settings.py to make sure that Census tables are from the new database
- Implement DBRouter to route reads and writes to the correct database
- Call `manage.py migrate` for the new datavase in the many places where we run `migrate`

## PR checklist: submitters

## PR checklist: reviewers


backend/config/settings.py
```diff
@@ -240,6 +244,9 @@
     MIDDLEWARE.append("whitenoise.middleware.WhiteNoiseMiddleware")
     DEFAULT_FILE_STORAGE = "report_submission.storages.S3PrivateStorage"
 
+    # Private C2F bucket
+    AWS_C2F_BUCKET_NAME = "fac-c2g-s3"  # TODO need to fix for CG
```
asteel-gsa, line 248:
```suggestion
    AWS_C2F_BUCKET_NAME = "fac-c2g-s3"  # TODO need to fix for CG
```

This falls under `if ENVIRONMENT not in ["DEVELOPMENT", "PREVIEW", "STAGING", "PRODUCTION"]:` as it is now, and i believe it will need its own elif statement for cgov environments:
```
        elif service["instance_name"] == "fac-c2g-s3":
            # What is the bucket for?
            s3_creds = service["credentials"]

            AWS_PRIVATE_ACCESS_KEY_ID = s3_creds["access_key_id"]
            AWS_PRIVATE_SECRET_ACCESS_KEY = s3_creds["secret_access_key"]
            AWS_PRIVATE_STORAGE_BUCKET_NAME = s3_creds["bucket"]

            AWS_S3_PRIVATE_REGION_NAME = s3_creds["region"]
            AWS_S3_PRIVATE_CUSTOM_DOMAIN = f"{AWS_PRIVATE_STORAGE_BUCKET_NAME}.s3-{AWS_S3_PRIVATE_REGION_NAME}.amazonaws.com"
            AWS_S3_PRIVATE_OBJECT_PARAMETERS = {"CacheControl": "max-age=86400"}

            AWS_S3_PRIVATE_ENDPOINT = s3_creds["endpoint"]
            AWS_S3_ENDPOINT_URL = f"https://{AWS_S3_PRIVATE_ENDPOINT}"

            AWS_PRIVATE_LOCATION = "static"
            AWS_PRIVATE_DEFAULT_ACL = "private"
            # If wrong, https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl

            MEDIA_URL = (
                f"https://{AWS_S3_PRIVATE_CUSTOM_DOMAIN}/{AWS_PRIVATE_LOCATION}/"
            )
```

backend/config/settings.py
```diff
@@ -382,16 +389,10 @@
     # Will not be enabled in cloud environments
     DISABLE_AUTH = False
 
-# Remove once all Census data has been migrated
```
jadudm, line 385: These are being removed before all Census data has been migrated. Is this correct?

backend/docker-compose.yml
```diff
@@ -18,6 +18,20 @@ services:
       timeout: 5s
       retries: 10
 
+  c2g-db:
```
jadudm, line 21: I think this will want to be in compose-web and the M1 dockerfile as well?

danswick:
Some of these changes may already be in main. @JeanMarie-TTS to confirm.

JeanMarie-TTS:
> Some of these changes may already be in main. @JeanMarie-TTS to confirm.

@danswick This PR was cloned and deployed earlier, but the new RDS functionality was not working. The current version of this PR fixes this bug.

terraform/shared/modules/env/env.tf
```diff
@@ -8,6 +8,16 @@ module "database" {
   rds_plan_name    = var.database_plan
 }
 
+module "database-c2g" {
```
phildominguez-gsa, line 11: I think we're moving away from "c2g" in favor of something more descriptive

phildominguez-gsa (changes requested):

backend/config/settings.py
danswick: These are left over from the old data distro work. We should be able to delete them now without trouble, but could also wait until we do the full data_distro cleanup.

danswick (approved):
Reviewed while pairing. Looks good 👍

Issue #1111, opened 2023-05-11
Create a Django command to insert a few placeholder audits


jadudm:
We often have to go through the Pre-SAC and General Info page to get things started.

We could have a Django command that takes an email address, and goes through and creates some objects in the DB for a few audits. This would make things easier for manual testing.

jadudm:
Done.

PR #2583, opened 2023-10-21
Adjust user setting when generating workbooks

danswick:
fixes https://github.com/GSA-TTS/FAC/issues/1111

## PR checklist: submitters

## PR checklist: reviewers


backend/dissemination/workbooklib/workbook_creation.py
```diff
@@ -61,7 +60,12 @@ def setup_sac(user, test_name, dbkey):
     else:
```
sambodeme, line 60: The reason  for this `else` statement is because `setup_sac` is called [here](https://github.com/GSA-TTS/FAC/blob/main/backend/dissemination/management/commands/generate_workbook_files.py#L200) with `user=None`. However the `sac` object created by this call is the passed to [workbook_loader](https://github.com/GSA-TTS/FAC/blob/main/backend/dissemination/management/commands/generate_workbook_files.py#L201) which makes no use of this object because `user=None`  ([see here](https://github.com/GSA-TTS/FAC/blob/main/backend/dissemination/workbooklib/workbook_creation.py#L84). In the duplicated version of the code, the [else statement](https://github.com/GSA-TTS/FAC/blob/main/backend/census_historical_migration/workbooklib/workbook_creation.py#L59) has been removed and the call to [workbook_loader ](https://github.com/GSA-TTS/FAC/blob/main/backend/census_historical_migration/management/commands/historic_workbook_generator.py#L199) simplified. At some point, we will refactor workbook_loader to separate the workbook creation from the load process which will prevent this kind of [call](https://github.com/GSA-TTS/FAC/blob/main/backend/census_historical_migration/management/commands/historic_workbook_generator.py#L200)

sambodeme (approved):
I added a comment.

backend/dissemination/workbooklib/workbook_creation.py
danswick: We still need a `User` in order to create a test SAC in the first place, right? Agreed we should refactor at some point, though. For now, I'm happy to get this change in so we can stop the dependency errors from the `Baker` import.

sambodeme: Yes, we need user in order to create a test SAC (and this sac object is not used for workbook creation :))

PR #2585, opened 2023-10-22
Add Super Linter [POST SEARCH]

asteel-gsa:
Adds super linter to workflows, which will detect any new files coming into the codebase and lint them. While we can [Limit the linter to specific languages](https://github.com/marketplace/actions/super-linter#environment-variables) our current implementation already does a pretty good job of that, this just adds another layer of linting to the process.


jadudm (commented):
Does the super linter have a local equivalent?

asteel-gsa:
> Does the super linter have a local equivalent?

It does, yes. [running via docker](https://github.com/github/super-linter/blob/main/docs/run-linter-locally.md) want me to add it into the makefile and try and get it running locally? @jadudm

asteel-gsa:
Went ahead and removed these since it does drastically increase the runtime, and well, we already use more lightweight versions:
```
VALIDATE_PYTHON
VALIDATE_PYTHON_BLACK
VALIDATE_PYTHON_FLAKE8
VALIDATE_PYTHON_MYPY
VALIDATE_PYTHON_PYLINT
```

Current process:
Local: Lints Everything as specified in `super-linter.env` (can be updated in the future)
GHA: Lints **ONLY NEW** entries into the code base, but lints everything

PR #2616, opened 2023-10-25
Tadhg/cogassign no signals

tadhg-ohiggins:
Try the direct way
And not the indirect way;
Let’s see what happens.

-----

I [commented](https://github.com/GSA-TTS/FAC/pull/2483#discussion_r1366368647) on [2483](https://github.com/GSA-TTS/FAC/pull/2483) suggesting something similar to this approach. This removes reliance on signals entirely and instead makes any `save()` of `CognizantAssignment` check to see if this is new object creation, and if it is, do the things that were formerly in `support.signals`.


JeanMarie-TTS:
@tadhg-ohiggins Too many changes to fully check.
But I'm good with moving the code from signals.py to models.py.
I don't know where the call to cog is being made upon submission - hopefully, you've removed sac.sace() from there.
If all tests are passing, you have my approval to merge.

tadhg-ohiggins:
@JeanMarie-TTS now that `main` has been merged in, this PR consists of two changed files, one of which is `signals.py` being deleted entirely.

The change this branch makes is to [override the `save` method](https://github.com/GSA-TTS/FAC/pull/2616/files#diff-e68b5017f5119dc5424001bea25e6eccca1ab5f123904307336681721d8d24a3R88) of `CognizantAssignemt` so that whenever it’s invoked, it checks to see if the save is an instance being created or updated. If it’s being created, it then updates the corresponding `SingleAuditChecklist` appropriately and saves it, then also runs `reset_baseline` and attempts to update the corresponding `dissemination.General` entry as well (if one exists).

The logic is intended to be the same as was happening in `support/signals.py`. With this version, the code is synchronous and the code path more linear.

If the logic in this PR matches what you expect, you can approve it and merge it into `jmm/resolve-cog-baseline-bug`, and then if the tests on https://github.com/GSA-TTS/FAC/pull/2483 pass that should be good to merge into `main`.

PR #2636, opened 2023-10-27
Media Backup Hotfix

asteel-gsa:
apparently, didn't like the export, so using full path after aws install


jadudm (approved):
LGTM. Hotfix.

PR #2638, opened 2023-10-27
Update requirements moving model_bakery to primary requirements out of dev

tadhg-ohiggins:
“The bakery’s open.”
“Bake it until you make it.”
“Bread: not just for devs.”

-----

Since we’re now using workbook generation in more places, and specifically via [backend/dissemination/management/commands/end_to_end_test_data_generator.py](https://github.com/GSA-TTS/FAC/blob/main/backend/dissemination/management/commands/end_to_end_test_data_generator.py), we need to move `model_bakery` out of `dev-requirements` and into `requirements`

If the tests pass, this PR should be fine.


tadhg-ohiggins:
Apparently we weren't using `baker` heavily enough to need to bring it in, closing.

PR #2642, opened 2023-10-27
Search: inexact matches on name fields

### LINKED ISSUE #2637: Make name search queries use LIKE semantics

timoballard:
Resolves #2637 

https://github.com/GSA-TTS/FAC/assets/1425377/eae8ecee-f704-42a6-b248-d561187e6fc9

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/dissemination/search.py
+++ b/backend/dissemination/search.py
@@ -17,7 +17,10 @@ def search_general(
     # TODO: use something like auditee_name__contains
     # SELECT * WHERE auditee_name LIKE '%SomeString%'
     if names:
-        names_match = Q(Q(auditee_name__in=names) | Q(auditor_firm_name__in=names))
+        names_match = Q()
+        for name in names:
+            names_match.add(Q(auditee_name__search=name), Q.OR)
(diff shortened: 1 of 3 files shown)


backend/dissemination/search.py
```diff
@@ -17,7 +17,10 @@ def search_general(
     # TODO: use something like auditee_name__contains
     # SELECT * WHERE auditee_name LIKE '%SomeString%'
     if names:
-        names_match = Q(Q(auditee_name__in=names) | Q(auditor_firm_name__in=names))
+        names_match = Q()
+        for name in names:
+            names_match.add(Q(auditee_name__search=name), Q.OR)
```
jadudm, line 22: do we downcase/normalize anywhere?

timoballard: not explicitly, because this (`__search`) uses [psql's full text search](https://www.postgresql.org/docs/9.4/textsearch-intro.html), which is case-insensitive by default as far as I can tell (and it does in fact behave as such in our tests)

tadhg-ohiggins (approved):

PR #2684, opened 2023-11-01
New Relic Deployment Blips

asteel-gsa:
Adds a workflow to tell new relic when a deployment is happening. This will allow us to see (via NR) that our deploys are going through, and when downtime occurs, and will help finish the new relic service setup


jadudm (approved):
LGTM

Issue #2693, opened 2023-11-02
FAC-D-2023-009 - Scanning at rest


jadudm:
```[tasklist]
### Tasks
- [x] ClamAV needs to scan files at rest
- [x] Schedule to run on a periodic basis
- [x] Capture events of failed scans in NR and remediate
```

jadudm:
https://github.com/GSA-TTS/FAC/tree/jadudm/scan-files-at-rest-for-viruses

danswick:
@asteel-gsa will test in preview, monitor performance, check results with a collection of historical test data.

jadudm:
@danswick , @asteel-gsa lets put in a GH Action/cron that runs this every six months, scheduling it to run in April and... whatever six months from April is. 

We're going to have to discuss the strategy as a whole w/ GSA, but for now, we need to complete the ticket.

jadudm:
Hm. We may also want to think about how it reports things, how it keeps track of what it has run, etc... :/

asteel-gsa:
I am fine merging it as is, just to say it is running. However, I am not keen on the idea of watching it for x months 😆 

it would probably be nice to have a `logger.info(f"SCAN PASS: {object}")` statement as well, though that may clutter the logs, so... 

the `excel` bucket is fine however, I see a lot of `Read timed out. (read timeout=15)"}` errors, and not a true failed scan. I don't see anything in the code to `exit 1` either, so, I am rerunning `singleauditreport` scan in `preview` atm, but, considering it will run for 2 days, we probably shouldn't deploy anything else to preview, so I can check the logs as to why it failed the task. (at this point, my only guess is the task runner crashed)

asteel-gsa:
@jadudm I think we need to look at this a bit further. Even in `preview` this operation is so long running that when I realize the task has failed and check the logs, they are gone.

It has a lot of trouble combing through `${bucket}/singleauditreport/` with an incredible amount of failed scans, so we might want to adjust the read timeout to 30s or more

jadudm:
It is probably worse than it appears.

It may be the smart thing is to just have the convo about scanning inbound *and* outbound, as opposed to trying to solve this problem. Otherwise, we're going to need a whole database/logging framework for tracking how often we've scanned all 700K+ files.

:/ 

More TBD.

danswick:
[...]

PR #2705, opened 2023-11-02
A Django command that scans files at rest

jadudm:
https://github.com/GSA-TTS/FAC/issues/2693

This is a Django command that can do one of two things:

1. Given a --bucket and --object, it will scan that object.
2. Given a --bucket and --path, it will scan all the files at that path

Common calling patterns will be:

* `fac scan_bucket_files_for_viruses --bucket gsa-fac-private-s3 --path singleauditreport`
* `fac scan_bucket_files_for_viruses --bucket gsa-fac-private-s3 --path excel`

This should be testable locally, and once merged, we can test it in `dev`. If it can be run manually in `dev`, we can then build an associated GH Action to trigger it. 

This does not write to any files, and therefore is not a risk to data. It does generate log messages.

https://github.com/GSA-TTS/FAC/assets/38290/70148ec7-4754-4c3b-9df2-30220bbdd481

## PR checklist: submitters

## PR checklist: reviewers


danswick:
Some questions that came up during cowork:
* Is there any additional risk to doing this scan from inside the Django app? If a potentially risky file exists, aren't we guaranteeing that it touches the app by scanning every file?
* How should we handle inevitable ClamAV timeouts? 
* Once we start getting a lot more audits, ~how much compute/memory/etc are we going to need? If it takes 1 second to scan a file (which is likely a vast underestimate), and there are 2.5 million seconds in a month, we would outgrow a "monthly scan" cadence once we hit 2.5 million media files in S3 (which includes ~500k historic PDF files). 
  * Maybe we need to consider purging workbook files either annually or when a submission is complete.
* What would ingress/egress costs look like to scan every file?
* Could scanning files _on the way out_, every time, address the threat vector in question? 
* Could we just scan any time ClamAV or its definitions are updated? Would this qualify as a "periodic" scan?

jadudm:
Q: Is there any additional risk to doing this scan from inside the Django app? If a potentially risky file exists, aren't we guaranteeing that it touches the app by scanning every file?
A: We run this in an Action, so it is not running in our production app space. If the file is a problem for us to load into a BytesIO object and pass to ClamAV, then we were in trouble when the file came in in the first place.
A: I'm not sure what tooling would be "safer" to load a file and pass it to ClamAV via HTTPS POST. No matter how you slice it, the file has to pass through RAM on its way to the "remote" ClamAV instance.
A: I may be missing something here.

Q: How should we handle inevitable ClamAV timeouts?
A: It's a failed scan. Either we add a backoff and retry, or we call the file "no good" and quarantine it?
A: If the conn doesn't timeout naturally, we need to add a timeout.
A: This was a first step. We need something that can run  and log, and that we can catch in NR. A documented mitigation plan, for the moment, would suffice.

Q: Once we start getting a lot more audits, ~how much compute/memory/etc are we going to need? If it takes 1 second to scan a file (which is likely a vast underestimate), and there are 2.5 million seconds in a month, we would outgrow a "monthly scan" cadence once we hit 2.5 million media files in S3 (which includes ~500k historic PDF files).
A: Hm.

```
(define pdfs 700000)
(define xlsx (* pdfs 4))
(define files (+ pdfs xlsx))
(define time-per-file 4)
(define total-time (* files time-per-file))
(define hours (/ (/ total-time 60) 60))
(define days (/ hours 24))
(define months (/ days 30))
```
gives me about ~5 months. 

A1: We can do that twice a year!
A2: I have no idea how we address this.

Q: Maybe we need to consider purging workbook files either annually or when a submission is complete.
A: Or, monthly. 
A: Or, zip them all every month, so there's only one file to scan per month.
A: Or, yep.

Q: What would ingress/egress costs look like to scan every file?
A: $0? Ingress/egress is to/from the internet? So, I think this would be S3 to our app, both of which are within the same GovCloud region. If this is true, then I think there is no ingress/egress between the app (running as an action) and the S3 storage.
A: We would want to check with cloud.gov.

Q: Could scanning files on the way out, every time, address the threat vector in question?
A: We could try and make the case. We'd have to convince GSA leadership/our CISO that this is the preferable solution to running virus scans all year long.

Q: Could we just scan any time ClamAV or its definitions are updated? Would this qualify as a "periodic" scan?
A: That would likely be more often. Meaning... we couldn't run the scans fast enough to keep up with this. It would count as periodic, I think, but we couldn't run all the files.

backend/dissemination/management/commands/scan_bucket_files_for_viruses.py
```diff
@@ -0,0 +1,145 @@
...
+                logger.error(f"SCAN FAIL: {object}")
+        if path:
+            results = scan_files_at_path_in_s3(bucket, path)
+            if results:
+                if all(map(check_scan_ok, results)):
+                    logger.info(f"SCAN OK: COUNT {len(results)}")
+                else:
+                    for r in results:
+                        if is_stringlike(r):
+                            logger.error(f"SCAN FAIL: {r}")
```
asteel-gsa, line 141: As this is right now, since this is just a django command, and has no associated workflow, (and even if it did, question still applies when run with cf tasks) how do we get the output back to us for consumption? 

Could we write fails to a file and send it off somewhere?

jadudm: I thought that the `logger` actions, as a Django command, would be picked up by `logshipper`/NR. 

Is that incorrect? 

If so, then yes: I need to bundle up results and send them somewhere.

tadhg-ohiggins: According to https://docs.cloudfoundry.org/devguide/using-tasks.html#-task-logging-and-execution-history :

> Any data or messages the task outputs to stdout or stderr is available in the firehose logs of the app. A syslog drain attached to the app receives the task log output. The task execution history is retained for one month.

So it looks like that `logger.info` line would be picked up by `logshipper`/NR.

backend/dissemination/management/commands/scan_bucket_files_for_viruses.py
```diff
@@ -0,0 +1,145 @@
...
+    file.seek(0)
+    return file
+
+
+def scan_file_in_s3(bucket, object_name):
+    try:
+        io_obj = load_file_from_s3(bucket, object_name)
+        bytes_blob = io_obj.read()
+        simple_obj = SimpleUploadedFile(object_name, bytes_blob)
+        return _scan_file(simple_obj)
```
danswick, line 79: In order for this to log positives, we need to check the result status code and log any found viruses here. Ref: https://github.com/GSA-TTS/FAC/blob/main/backend/audit/validators.py#L56.

backend/dissemination/management/commands/scan_bucket_files_for_viruses.py
```diff
@@ -0,0 +1,145 @@
...
+def scan_files_at_path_in_s3(bucket, path):
+    s3 = get_s3_client()
+    objects = s3.list_objects(Bucket=bucket, Prefix=path)
+    if objects:
+        results = []
+        if "Contents" in objects:
+            for object_summary in objects["Contents"]:
+                object_name = object_summary["Key"]
+                result = scan_file_in_s3(bucket, object_name)
+                results.append(result)
```
danswick, line 94: This is likely to get bogged down by the volume of files. We should log each time ClamAV finds a suspicious file instead of creating one large results object.

backend/dissemination/management/commands/scan_bucket_files_for_viruses.py
```diff
@@ -0,0 +1,145 @@
...
+        simple_obj = SimpleUploadedFile(object_name, bytes_blob)
+        return _scan_file(simple_obj)
+    except Exception as e:
+        logger.error(f"SCAN SCAN_FILE_IN_S3 {e}")
+        return f"{object_name}"
+
+
+def scan_files_at_path_in_s3(bucket, path):
+    s3 = get_s3_client()
+    objects = s3.list_objects(Bucket=bucket, Prefix=path)
```
danswick, line 87: `list_objects` can only return up to 1000 objects in a single request, so we'll need to handle pagination or figure something else out.

[Boto3 list_objects docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_objects_v2.html).

[Boto3 built-in paginator docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/paginator/ListObjectsV2.html#)

danswick (commented):
adding some inline comments following conversation with @tadhg-ohiggins.

tadhg-ohiggins (approved):

tadhg-ohiggins:
Contents of this PR moved to https://github.com/GSA-TTS/FAC/pull/3285 due to Git commit signing issues; closing.

PR #2728, opened 2023-11-03
Merge product decision records into architecture decision records

tadhg-ohiggins:
Some categories
Only increase confusion;
Stick with one for now.

-----

Move Product Decision Records into Architecture Decision Records so that all of our decision records are in one place.

Before anyone asks, yes, I do plan to write an ADR that mentions this move, but no, I don’t think we should wait for that before doing this moving and renaming.

-----

This is a documentation/team practices change only.


jadudm (approved):
LGTM, and thank you.

Issue #2672, opened 2023-11-01
Epic: improve testing around core/critical E2E functions


jadudm:
# At a glance

[comment]: # "Begin with a short summary so intent can be understood at a glance."
[comment]: # "In order to: some objective or value to be achieved"
[comment]: # "as a: stakeholder"
[comment]: # "I want: some new feature"

**In order to** have confidence in the workbook migration code 
**as a** product owner
**I want** more testing around what was previously exploratory/nice-to-have app code.

## `workbook_creation.py`

https://github.com/GSA-TTS/FAC/blob/main/backend/dissemination/workbooklib/workbook_creation.py

```[tasklist]
### Tasks
- [ ] https://github.com/GSA-TTS/FAC/issues/2673
- [ ] https://github.com/GSA-TTS/FAC/issues/2674
- [ ] https://github.com/GSA-TTS/FAC/issues/2675
- [ ] https://github.com/GSA-TTS/FAC/issues/2833
- [ ] https://github.com/GSA-TTS/FAC/issues/2890
```

## `excel_creation.py`

```[tasklist]
### Tasks
- [ ] https://github.com/GSA-TTS/FAC/issues/2676
- [x] E2E improve: Remove `test_pfix()`
- [ ] https://github.com/GSA-TTS/FAC/issues/2677
- [ ] https://github.com/GSA-TTS/FAC/issues/2678
- [ ] https://github.com/GSA-TTS/FAC/issues/2679
- [ ] https://github.com/GSA-TTS/FAC/issues/2680
```

The code around test tables is problematic, and [I speak about it in the E2E documentation](https://docs.google.com/document/d/12jcupPwqTc1muIaCsMxKgC33cOPOQQ6V4EoH9OmZGo0/edit?disco=AAAA_SMW678). There are ways to improve this dramatically, but it may not be worth it. In which case, despite how this code looks... *not touching it* might be the right thing to do.

## `sac_creation.py`

Probably all of it. At the least, we need a way to map

historic field -> named range -> api_field

as well as provide a migration function. In essence, it would be good to have a dictionary of `FieldMap` structures (just like the workbooks) that do the work of creating the general info object. However, we would then take the FieldMap structures and, instead of filling a workbook, use them to create a JSON object that represents the `general_info` and `audit_info` objects in the SAC.

That would be the best way to clean this whole thing up. And, it would then be easy to integrate the general info form into the API end-to-end testing.

PR #2790, opened 2023-11-13
Removing `test_pfix()` from `workbooklib`

phildominguez-gsa:
Addresses https://github.com/GSA-TTS/FAC/issues/2672

In this PR:
- Removes `test_pfix()` from `workbooklib`, as it doesn't do anything
- Replacing some instances with `str` if a submission fails without it

Testing:
- `docker compose run web python manage.py historic_data_migrator --email <your email>`
  - If you get a login error, log into the app as normal and retry

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/census_historical_migration/workbooklib/notes_to_sefa.py
+++ b/backend/census_historical_migration/workbooklib/notes_to_sefa.py
@@ -4,7 +4,6 @@ from census_historical_migration.workbooklib.excel_creation import (
     set_single_cell_range,
     map_simple_columns,
     generate_dissemination_test_table,
-    test_pfix,
 )
 from census_historical_migration.workbooklib.templates import sections_to_template_paths
 from census_historical_migration.workbooklib.excel_creation import (
@@ -22,9 +21,8 @@ import logging
 logger = logging.getLogger(__name__)
 
 mappings = [
-    FieldMap("note_title", "title", "title", None, test_pfix(3)),
-    FieldMap("note_content", "content", "content", None, test_pfix(3)),
-    # FieldMap("seq_number", "seq_number", "note_seq_number", 0, int),
+    FieldMap("note_title", "title", "title", None, str),
+    FieldMap("note_content", "content", "content", None, str),
 ]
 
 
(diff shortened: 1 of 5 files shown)


JeanMarie-TTS (commented):
LGTM

JeanMarie-TTS (approved):

backend/census_historical_migration/workbooklib/secondary_auditors.py
```diff
@@ -24,20 +23,6 @@
     FieldMap("secondary_auditor_name", "cpafirmname", "auditor_name", None, str),
     FieldMap("secondary_auditor_contact_phone", "cpaphone", "contact_phone", None, str),
     FieldMap("secondary_auditor_address_state", "cpastate", "address_state", None, str),
-    FieldMap(
```
sambodeme, line 27: Why are these removed ?

sambodeme (changes requested):

backend/census_historical_migration/workbooklib/secondary_auditors.py
phildominguez-gsa: Oops, fixed

backend/census_historical_migration/workbooklib/notes_to_sefa.py
```diff
@@ -4,7 +4,6 @@
     set_single_cell_range,
     map_simple_columns,
     generate_dissemination_test_table,
-    test_pfix,
```
JeanMarie-TTS, line 7: Does this not have to be replaced with str, rather than merely removing test_pfix?

backend/census_historical_migration/workbooklib/secondary_auditors.py
```diff
@@ -3,7 +3,6 @@
     set_uei,
     map_simple_columns,
     generate_dissemination_test_table,
-    test_pfix,
```
JeanMarie-TTS, line 6: replace with str?

backend/census_historical_migration/workbooklib/notes_to_sefa.py
phildominguez-gsa: Nope, `str` is a built-in Python [method](https://www.w3schools.com/python/ref_func_str.asp)

backend/census_historical_migration/workbooklib/secondary_auditors.py
phildominguez-gsa: Nope, see [other comment](https://github.com/GSA-TTS/FAC/pull/2790/files#r1393062897)

sambodeme (approved):

Issue #2674, opened 2023-11-01
E2E improve: `setup_sac()`


jadudm:
[setup_sac()](https://github.com/GSA-TTS/FAC/blob/main/backend/dissemination/workbooklib/workbook_creation.py#L53)

This code essentially builds a SAC object as well as loads the `general_info` form.

This code could be broken out better, especially around form population. There are multiple places in this code where *wrong* data is generated to complete the task of generating a submission.

A consistent approach to a "test user" for submitting/loading test data should be taken.

Either way, this is essentially a data load function, but it looks different than the workbook data loads. A dictionary of fields, how they map to Census data, and a place to add migration functions that map the old to the new feels like a necessary part of this refactoring.

PR #2794, opened 2023-11-14
2674 e2e improve setup sac

sambodeme:
## Description
This PR introduces a series of refactoring changes around the setup_sac logic:
- Updated the names of various functions, variables, and named tuples for improved clarity and consistency.
- Removed sections of code that were no longer in use to streamline the codebase.
- Segmented the existing code into separate files. This enhances readability and maintainability by organizing related functionalities more logically.
- Reorganized the structure of the existing code to improve flow and comprehension.
- Ticket is #2674 

For Testing:

- To generate the necessary workbooks, run the following command:
```docker compose run --rm web python manage.py historic_workbook_generator --year 22 --output data --dbkey 171944```
  This command will create a folder named data inside your project and populate it with the generated workbooks.

- To create and process an audit report, use this command:
```docker compose run --rm web python manage.py historic_data_migrator --years 22 --email your_email_here --dbkeys 182926```
This process will create an audit report, submit it, and then disseminate it. Make sure that you have logged into the application with the specified email address before running this command. 

## PR checklist: submitters

## PR checklist: reviewers

### CODE REVIEW

File: backend/census_historical_migration/base_field_maps.py [hunk 1]
```diff
@@ -0,0 +1,17 @@
+from collections import namedtuple as NT
+
+FormFieldInDissem = WorkbookFieldInDissem = 1000
+
+# This provides a way to map the sheet in the workbook to the
+# column in the DB. It also has a default value and
+# the type of value, so that things can be set correctly
+# before filling in the XLSX workbooks.
+# Define a new named tuple with an additional field
+SheetFieldMap = NT("SheetFieldMap", "in_sheet in_db in_dissem default type")
```
jadudm, line 10: Strictly as a comment, not a proposed revision:

These are currently named tuples, or "records." If turning these into classes would make them more useful, it is completely reasonable to do so. FYI.

backend/census_historical_migration/sac_general_lib/audit_information.py
```diff
@@ -0,0 +1,150 @@
...
+    CensusCfda22 as Cfda,
+    CensusFindings22 as Finding,
+)
+from census_historical_migration.base_field_maps import FormFieldMap, FormFieldInDissem
+from census_historical_migration.sac_general_lib.utils import (
+    _create_json_from_db_object,
+)
+import audit.validators
+
+DOLLAR_THRESHOLD = 750000
```
jadudm, line 12: It would be completely reasonable to move this out to the app-level `config`, for a variety of reasons. Perhaps not part of this commit, but we are going to need this number elsewhere.

backend/census_historical_migration/sac_general_lib/audit_information.py
```diff
@@ -0,0 +1,150 @@
...
+
+mappings = [
+    FormFieldMap(
+        "dollar_threshold", None, FormFieldInDissem, DOLLAR_THRESHOLD, int
```
jadudm, line 16: There is. 

The draft dictionary (which needs to get published...)

https://federalist-35af9df5-a894-4ae9-aa3d-f6d95427c7bc.sites.pages.cloud.gov/preview/gsa-tts/fac-transition-site/jadudm/dictionaries/developers/v1_0_2/dictionary/

may help?

backend/census_historical_migration/sac_general_lib/audit_information.py
```diff
@@ -0,0 +1,150 @@
...
+        "reportablecondition",
+        FormFieldInDissem,
+        None,
+        bool,
+    ),
+    FormFieldMap(
+        "is_low_risk_auditee", None, FormFieldInDissem, False, bool
+    ),  # FIXME: There is no in_db mapping?
+    FormFieldMap(
+        "agencies", None, "agencies_with_prior_findings", [], list
```
jadudm, line 56: This needs to be fixed in the dictionary. I think this is `PYSCHEDULE`. Maybe.

backend/census_historical_migration/sac_general_lib/audit_information.py
```diff
@@ -0,0 +1,150 @@
...
+    ),  # FIXME: There is no in_db mapping?
+]
+
+
+def _get_agency_prefixes(dbkey):
+    agencies = {}
+    cfdas = Cfda.select().where(Cfda.dbkey == dbkey)
+    cfda: Cfda
+    for cfda in cfdas:
+        agencies[int((cfda.cfda).split(".")[0])] = 1
```
jadudm, line 66: It might be this can be a bit cleaner, and perhaps use a `set` instead of the weird use of a hash table to simulate a set. Or, you can leave it. I know I wrote that code, and perhaps it can be clearer as to its purpose. Up to you.

backend/census_historical_migration/sac_general_lib/audit_information.py
```diff
@@ -0,0 +1,150 @@
...
+    return agencies.keys()
+
+
+def _get_gaap_results(dbkey):
+    findings = Finding.select().where(Finding.dbkey == dbkey)
+    gaap_results = {}
+    # FIXME: How do we retrieve gaap_results from the historic data? I could not find corresponding fields in Census tables.
```
jadudm, line 73: It has a different name. We might ask @Carley-Jackson .

backend/census_historical_migration/sac_general_lib/auditee_certification.py
```diff
@@ -0,0 +1,45 @@
...
+#     FormFieldMap("is_complete_and_accurate", None, FormFieldInDissem, False, bool),
+#     FormFieldMap("has_engaged_auditor", None, FormFieldInDissem, False, bool),
+#     FormFieldMap("is_issued_and_signed", None, FormFieldInDissem, False, bool),
+#     FormFieldMap("is_FAC_releasable", None, FormFieldInDissem, False, bool),
+# ]
+
+auditee_signature_mappings = [
+    FormFieldMap("auditee_name", "auditeename", FormFieldInDissem, None, str),
+    FormFieldMap("auditee_title", "auditeetitle", FormFieldInDissem, None, str),
+    # FIXME: Do we need this ?
```
jadudm, line 25: I don't think we disseminate it. So, possibly/probably not. I'm working from memory, though.

backend/census_historical_migration/sac_general_lib/auditee_certification.py
```diff
@@ -0,0 +1,45 @@
...
+def _auditee_certification(dbkey):
+    gobj: Gen = Gen.select().where(Gen.dbkey == dbkey).first()
+    certification = {}
+    # certification["auditee_certification"] = _create_json_from_db_object(
+    #     gobj, auditee_certification_mappings
+    # )
+    certification["auditee_signature"] = _create_json_from_db_object(
+        gobj, auditee_signature_mappings
+    )
+    # FIXME: Do we need this ?
```
jadudm, line 41: When I hacked this in, I added *just enough data* to pass the cross-validations. I'm... on one hand, it might be nice to be complete, but I don't think we have the historic data to populate the checkboxes. Or: _by definition_, they had to accept the terms in order to submit the FAC. So, _by definition_, our reconstruction would just "automatically" check all the boxes.

Use your judgement here. This is an object that is far, far lower in the integrity chain than, say, `federal_awards`. If the critical information (signature, role, etc.) are present, then building the full object with the booleans for the checkmarks is probably not needed.

If I understand the "do we need this?" question correctly. Happy to discuss.

backend/census_historical_migration/sac_general_lib/auditor_certification.py
```diff
@@ -0,0 +1,42 @@
+# import audit.validators
+from census_historical_migration.workbooklib.census_models.census import (
+    CensusGen22 as Gen,
+)
+from census_historical_migration.base_field_maps import FormFieldMap, FormFieldInDissem
+from census_historical_migration.sac_general_lib.utils import (
+    _create_json_from_db_object,
+)
+
+# FIXME: Do we need this ?
```
jadudm, line 10: See previous comment. Probably not.

backend/census_historical_migration/sac_general_lib/general_information.py
jadudm, line 1: An overall comment: the restructuring of the forms (having not even finished reading through) is wonderful. Thank you.

backend/census_historical_migration/sac_general_lib/general_information.py
```diff
@@ -0,0 +1,147 @@
...
+    FormFieldMap("auditor_address_line_1", "cpastreet1", FormFieldInDissem, None, str),
+    FormFieldMap("auditor_city", "cpacity", FormFieldInDissem, None, str),
+    FormFieldMap("auditor_contact_name", "cpacontact", FormFieldInDissem, None, str),
+    FormFieldMap("auditor_contact_title", "cpatitle", FormFieldInDissem, None, str),
+    FormFieldMap("auditor_country", "cpacountry", FormFieldInDissem, None, str),
+    FormFieldMap("auditor_ein", "auditor_ein", FormFieldInDissem, None, str),
+    FormFieldMap("auditor_ein_not_an_ssn_attestation", None, None, True, bool),
+    FormFieldMap(
+        "auditor_email", "cpaemail", FormFieldInDissem, "noemailfound@noemail.com", str
+    ),  # FIXME: Do we want to keep `noemailfound@noemail.com`?
```
jadudm, line 43: No. We should pick something, though.

`fac-no-email-in-census-data@gsa.gov` or similar.

Whatever we use, if we use it, should be descriptive as to why it is being inserted.

jadudm: I think our process would actually want to treat this as a failure in the first instance? This might be where we want a mapping function.

backend/census_historical_migration/sac_general_lib/general_information.py
```diff
@@ -0,0 +1,147 @@
...
+)
+from census_historical_migration.base_field_maps import FormFieldMap, FormFieldInDissem
+from census_historical_migration.sac_general_lib.utils import (
+    _create_json_from_db_object,
+    _boolean_field,
+)
+
+mappings = [
+    FormFieldMap(
+        "auditee_fiscal_period_start", "fyenddate", "fy_start_date", None, str
```
jadudm, line 17: It is possible (perhaps not in this commit, but soon) that we want an xform that replaces all foundational Python types.

```
def string_to_string(...):
    # error checking, normalization
    # throw an exception if some condition happens
    return str(input)
```

or similar.

This way, we have a centralized xform for all strings, all bools, etc.

I'm particularly thinking about the email question, below. It might want its own special transform, to handle the missing case. (Or, perhaps the missing case/exception/error is part of a standard string-to-string transform).

backend/census_historical_migration/sac_general_lib/general_information.py
```diff
@@ -0,0 +1,147 @@
...
+    ),  # FIXME: There is no in_db mapping ?
+    FormFieldMap("secondary_auditors_exist", "multiple_cpas", None, None, bool),
+]
+
+
+def _period_covered(s):
+    return {"A": "annual", "B": "biennial", "O": "other"}[s]
+
+
+def _census_audit_type(s):
```
jadudm, line 67: I'm assuming a future version of this will incorporate some kind of error handling for when things are not present in the dictionary?

Again, flagging, but not for this PR. I think the cleanup is excellent, and gives us a staging ground for next steps.

backend/census_historical_migration/sac_general_lib/sac_creator.py
```diff
@@ -0,0 +1,92 @@
...
+        user=user,
+        email=user.email,
+        role="editor",
+    )
+
+    # We need these to be different.
+    Access.objects.create(
+        sac=sac,
+        user=user,
+        email="bob_the_auditee_official@auditee.org",  # user.email, FIXME - Confirm this is the right value?
```
jadudm, line 59: Like other values, we might need this to be more descriptive? Not sure what we should do here.

Actually, I do.

For now, I think we clean it up a bit. However, this needs a ticket and some discussion.

We're going to either want to insert a clearly defined placeholder, or we're going to want to pull values from the General form for the auditor and auditee contact, but... perhaps label them as choices we made.

What I'm concerned about is that these are _access objects_, and therefore they'll end up in our system. We're creating them because they get accessed as part of the migration. 

Hence... using one set of "fake" access objects makes sense.

But, they should still be `fac-census-migration-auditee-official` or similar.

backend/census_historical_migration/sac_general_lib/sac_creator.py
```diff
@@ -0,0 +1,92 @@
...
+    Access.objects.create(
+        sac=sac,
+        user=user,
+        email="bob_the_auditor_official@auditor.org",  # user.email, FIXME - Confirm this is the right value?
+        role="certifying_auditor_contact",
+    )
+
+    sac.auditee_certification = _auditee_certification(dbkey)
+    sac.auditor_certification = _auditor_certification(dbkey)
+    sac.data_source = "CENSUS"  # FIXME - Confirm this is the right value?
```
jadudm, line 71: We're using `GSAFAC` and `CENSUS` as keys, so yes. If these want to be broken out to a "constant" somewhere, that's fine.

backend/census_historical_migration/workbooklib/end_to_end_core.py
```diff
@@ -191,19 +191,19 @@ def api_check(json_test_tables):
...
 
 
-def generate_workbooks(user, email, dbkey, year):
+def generate_workbooks(user, dbkey, year):
     entity_id = "DBKEY {dbkey} {year} {date:%Y_%m_%d_%H_%M_%S}".format(
         dbkey=dbkey, year=year, date=datetime.now()
     )
     sac = setup_sac(user, entity_id, dbkey)
     if sac.general_information["audit_type"] == "alternative-compliance-engagement":
         print(f"Skipping ACE audit: {dbkey}")
```
jadudm, line 200: Definitely could become a log statement, and perhaps should become a logged "failure to migrate" when we see ACEE audits (for now).

backend/census_historical_migration/workbooklib/excel_creation_utils.py
```diff
@@ -62,7 +57,7 @@ def set_uei(Gen, wb, dbkey):
     if g.uei:
         set_range(wb, "auditee_uei", [g.uei])
     else:
-        g.uei = "BADBADBADBAD"
+        g.uei = "BADBADBADBAD"  # FIXME: Shouldn't this raise an exception here?
```
jadudm, line 60: Now it should, yes. Although, it will be 100% true for all audits before 2022. So, we have to have a strategy.

We can discuss. It should not be `BADBADBADBAD`, as you suspect.

backend/census_historical_migration/workbooklib/excel_creation_utils.py
```diff
@@ -92,43 +87,32 @@ def map_simple_columns(wb, mappings, values):
...
-    dt = _census_date_to_datetime(g.fyenddate)
-    return f"{g.audityear}-{dt.month:02}-TSTDAT-{dbkey.zfill(10)}"
+def add_hyphen_to_zip(zip):
+    strzip = str(zip)
+    if len(strzip) == 5:
+        return strzip
+    elif len(strzip) == 9:
+        return f"{strzip[0:5]}-{strzip[5:9]}"
+    else:
+        logger.info("ZIP IS MALFORMED IN WORKBOOKS E2E / SAC_CREATION")
```
jadudm, line 97: Possibly another place that should fail? Dunno.

backend/census_historical_migration/workbooklib/post_upload_utils.py
```diff
@@ -0,0 +1,162 @@
...
+        # nothing to do here
+        return
+
+    with open(pdf_filename, "rb") as f:
+        content = f.read()
+    file = SimpleUploadedFile(pdf_filename, content, "application/pdf")
+    print(file.__dict__)
+    pdf_file = PDFFile(
+        file=file,
+        component_page_numbers={  # FIXME - How do we want to handle these values?
```
jadudm, line 90: They might actually exist somewhere in the historic data. For now, I don't know. We should start a ticket of questions like this. Or something.

backend/census_historical_migration/base_field_maps.py
sambodeme: Agreed, actually I took that approach in my initial commit but I can remember why I diverted path.

backend/census_historical_migration/sac_general_lib/audit_information.py
sambodeme: Thanks for sharing this. I was struggling searching in googlesheets

backend/census_historical_migration/sac_general_lib/auditee_certification.py
sambodeme: We don't disseminate, I simply added this comment so that we can make a group decision.

backend/census_historical_migration/sac_general_lib/general_information.py
```diff
@@ -0,0 +1,153 @@
...
+    ),
+    FormFieldMap(
+        "user_provided_organization_type", None, "entity_type", "unknown", str
+    ),  # FIXME: There is no in_db mapping ?
+    FormFieldMap("secondary_auditors_exist", "multiple_cpas", None, None, bool),
+]
+
+
+def _period_covered(s):
+    period_dict = {"A": "annual", "B": "biennial", "O": "other"}
```
phildominguez-gsa, line 64: Could this be a const defined somewhere so it's not recreated every time this is called?

backend/census_historical_migration/sac_general_lib/general_information.py
```diff
@@ -0,0 +1,153 @@
...
+        raise KeyError(f"Key '{s}' not found in period coverage mapping")
+    return period_dict[s]
+
+
+def _census_audit_type(s):
+    audit_type_dict = {
+        "S": "single-audit",
+        "P": "program-specific",
+        "A": "alternative-compliance-engagement",
+    }
```
phildominguez-gsa, line 75: Similar here, could we define a const outside of the method?

backend/census_historical_migration/sac_general_lib/general_information.py
```diff
@@ -0,0 +1,153 @@
...
+    ),  # FIXME: There is no in_db mapping ?
+    FormFieldMap("secondary_auditors_exist", "multiple_cpas", None, None, bool),
+]
+
+
+def _period_covered(s):
+    period_dict = {"A": "annual", "B": "biennial", "O": "other"}
+    if s not in period_dict:
+        raise KeyError(f"Key '{s}' not found in period coverage mapping")
+    return period_dict[s]
```
phildominguez-gsa, line 67: I feel a `try... except` block would be more straightforward here

backend/census_historical_migration/sac_general_lib/general_information.py
```diff
@@ -0,0 +1,153 @@
...
+
+def _census_audit_type(s):
+    audit_type_dict = {
+        "S": "single-audit",
+        "P": "program-specific",
+        "A": "alternative-compliance-engagement",
+    }
+    if s not in audit_type_dict:
+        raise KeyError(f"Key '{s}' not found in census audit type mapping")
+    return audit_type_dict[s]
```
phildominguez-gsa, line 78: Similar here, `try... except`?

backend/census_historical_migration/sac_general_lib/sac_creator.py
```diff
@@ -0,0 +1,92 @@
...
+    sac.save()
+
+    logger.info("Created single audit checklist %s", sac)
+    return sac
+
+
+def setup_sac(user, test_name, dbkey):
```
phildominguez-gsa, line 78: While we're on this ticket I think it's a good time to put a doc string on this

backend/census_historical_migration/sac_general_lib/general_information.py
sambodeme: I created a [ticket](https://github.com/GSA-TTS/FAC/issues/2833) for this to limit the size of the PR

backend/census_historical_migration/sac_general_lib/auditee_certification.py
sambodeme: I opted to build the entire object so that we can utilize the validators.

backend/census_historical_migration/sac_general_lib/auditor_certification.py
sambodeme: Similar to my previous comment

backend/census_historical_migration/sac_general_lib/auditee_certification.py
sambodeme: I decided to keep the field so that we don't run into issue with the validator.

backend/census_historical_migration/sac_general_lib/audit_information.py
sambodeme: [Ticket ](https://github.com/GSA-TTS/FAC/issues/2832) created for follow up.

backend/census_historical_migration/sac_general_lib/general_information.py
sambodeme: We should treat this as a failure

backend/census_historical_migration/sac_general_lib/sac_creator.py
sambodeme: Using `fac-census-migration-auditee-official@auditee.org` for now but also added this to the [ticket](https://github.com/GSA-TTS/FAC/issues/2832)

backend/census_historical_migration/workbooklib/post_upload_utils.py
sambodeme: https://github.com/GSA-TTS/FAC/issues/2832

backend/census_historical_migration/workbooklib/excel_creation_utils.py
sambodeme: json-schema should catch this.

backend/census_historical_migration/sac_general_lib/general_information.py
sambodeme: We want to raise an exception here to fail the migration process

backend/census_historical_migration/sac_general_lib/general_information.py
sambodeme: Similar to my previous comment

backend/census_historical_migration/sac_general_lib/general_information.py
sambodeme: Actually, I saw a need for it and implemented this method and few others as part of the current PR. The ticket mentioned above will be use to improve that `string_to_string` method and add others (if necessary)

backend/census_historical_migration/sac_general_lib/general_information.py
sambodeme: This dictionary is small and unlikely to expand, and since it's only called once per submission (unlike other sections), keeping it within the function is better for readability and maintenance. However, I'll move it out due to its values being duplicated in `Base.libsonnet`, aligning with other code areas where we've adopted a similar approach for duplicated sources. This will help consolidate it in future.

backend/census_historical_migration/sac_general_lib/general_information.py
sambodeme: Similar to my comment above.

danswick (approved):
Discussed with @sambodeme 👍

backend/census_historical_migration/workbooklib/excel_creation_utils.py
jadudm: For the future/future work: yes, but I think *all* Census zips lack the hyphen. So, we may end up wanting a xform that takes 9-digit zips and adds the hyphen, or lots of data won't migrate. In this case, it isn't that the data is "bad," it's just we might be expecting a different form. (Totally may be mis-remembering.)

I can't remember what we expect. I know this merged, but noting it here @sambodeme, @phildominguez-gsa .

sambodeme: It looks like what you're describing, `... that takes 9-digit zips and adds the hyphen,` is what line 95 achieves. So either you meant to pull this into a separate xform for consistency, or you meant something else. I will ticket this and follow up with you.

PR #2807, opened 2023-11-15
Search: Permissioned tribal audit access

### LINKED ISSUE #2768: Permission model for tribal data access
Implement the permission model described in our [recent ADR around tribal data access](https://github.com/GSA-TTS/FAC/blob/main/docs/architecture/decisions/0030-tribal-audit-permissions.md).

We'll know we're done when...
- [ ] We can assign a permission that allows tribal data access via the Django admin site
- [ ] Authenticated users holding the tribal data access permission can see tribal audit data in their web search results
- [ ] Authenticated users not holding the tribal data access permission cannot see tribal audit data in their web search results
- [ ] Unauthenticated users cannot see tribal audit data in their web search results
- [ ] There are tests covering the above scenarios

timoballard:
Resolves #2768 

- Adds `Permission` and `UserPermission` models & migration (migration populates `Permission` table with `read-tribal` permission)
- Adds post-login step to claim all `UserPermissions` associated with the current user's email addresses
- Updates search, PDF and XLSX download views to conditionally allow access to private reports if the user holds the `read-tribal` permission

Screen recording demos:
- private report (INCA COMMUNITY SERVICES, INC.) not visible in search for anonymous user
- private report not visible to authenticated but unpermissioned user
- private report visible to permissioned user
- permissions managed through django admin

https://github.com/GSA-TTS/FAC/assets/1425377/4c61b3ab-8848-4967-847f-59b39059d4b7

## PR checklist: submitters

## PR checklist: reviewers


backend/dissemination/search.py
```diff
@@ -12,46 +12,30 @@ def search_general(
...
+    query.add(_get_end_date_match_query(end_date), Q.AND)
+    query.add(_get_cog_or_oversight_match_query(agency_name, cog_or_oversight), Q.AND)
+    query.add(_get_audit_years_match_query(audit_years), Q.AND)
 
     results = General.objects.filter(query).order_by("-fac_accepted_date")
 
     return results
 
 
+def _get_is_public_query(include_private):
```
tadhg-ohiggins, line 33: Would it make more sense to have an `if` clause around this, rather than having this clash between the function name and the variable name?

tadhg-ohiggins (approved):
Looks good apart from the one minor comment I made.

backend/dissemination/search.py
timoballard: 👍 moved the `is_public` query clause up to the root function for clarity

tadhg-ohiggins (approved):

PR #2811, opened 2023-11-16
Start run_migration command

JeanMarie-TTS:
## run_migration
Implement a django management command to run migrations.
It takes submissions from the raw Census tables in PG and attempts to load them one at a time into the FAC application using existing code from wirkbooklib.
This is intended to be a starting point for us to do a continuous cycle of maing changes including refactoring as we work through all the requirements laid out in the ADR.
Running this command on the existing codebase will throw errors, but it allows us to see what they are and work on them.
`manage.py run_migration --year 2022`

## PR checklist: submitters

## PR checklist: reviewers


backend/census_historical_migration/loader.py
```diff
@@ -0,0 +1,28 @@
+from audit.models import SingleAuditChecklist
+from .models import ELECAUDITHEADER as Gen
+from .make_submission import load_historic_data
+
+
+def load_data(audit_yaer):
+    SingleAuditChecklist.objects.all().delete()
```
jadudm, line 7: I know this is still in draft, but this absolutely cannot go to production as-is. This line is terrifying.

jadudm (changes requested):
See my comment on line 7.

I'm happy for us to do some pairing on this code. I would be happy to collaborate on improving it. But we cannot afford to have code going in that has something that might delete all of our data.

Perhaps I'm misreading this line, but that really is what it looks like?

backend/census_historical_migration/loader.py
```diff
@@ -0,0 +1,28 @@
+from audit.models import SingleAuditChecklist
+from .models import ELECAUDITHEADER as Gen
+from .make_submission import load_historic_data
+
+
+def load_data(audit_yaer):
```
eazapata23, line 6: `audit_yaer` should be `audit_year`

JeanMarie-TTS:
> See my comment on line 7.
> 
> I'm happy for us to do some pairing on this code. I would be happy to collaborate on improving it. But we cannot afford to have code going in that has something that might delete all of our data.
> 
> Perhaps I'm misreading this line, but that really is what it looks like?

@jadudm The delete line was temporary and is now remove.

jadudm:
I believe all of this work has been pulled into other PRs. Closing. It can always be reopened.

PR #2860, opened 2023-11-24
Administrative API and improved tribal data API access controls

jadudm:
This PR does a few things.

1. It removes old dissemination API versions (1.0.0, 1.0.1, and 1.0.2)
2. It changes tribal data access checks in the dissemination API. These were using a `role` set in api.data.gov. They will now use a statically-defined table of api.data.gov key UUIDs. Currently, there is one mock key present for demonstration purposes. In short: no one has tribal data access via API at this time. 
3. It introduces the administrative API.

The administrative API has several pieces.

1. A statically defined table (`create_access_tables`) of users whose API keys should be allowed to use the API. That table includes permission levels of SELECT, INSERT, and DELETE. (Why is below.)
2. A set of views over internal tables. These can be expanded. They initially include `audit_access`, which is useful for determining who has access to an audit, `singleauditchecklist`, which is the full internal table, `tribal_access`, which lets us see who has access to view tribal audits on the WWW, and `admin_api_events` lets us view the log of administrative events on the API. these all check against the UUID of the key and makes sure the user has SELECT permissions.
3. A set of API endpoints (`/rpc`) for adding and removing tribal access email addresses.

The `create_access_tables` code determines who has access to the administrative API. The API itself is used to determine who has access to tribal audits via the WWW. There are a handful of people (on the team) who fall into the former category. Many Federal users fall into the latter.

NOTE that `run.sh` is touched. Why? Because it was refactored at one point to run migrations before the API code. The API must be torn down first, then migrations run, then stood back up. This is because Postgrest cannot execute properly without the correct models in place. So, this commit puts the call to make migrations back in the middle of the sequence, as opposed to pulling it out. (If this can actually be run either way, then it should be demonstrated under the conditions of a model change with the API. I'm pretty confident that if a dissemination model changes, the API breaks if we don't leave the call in the middle.)

**HEADS UP**: There is a migration, and I'm... not 100% certain it is clean. I'd appreciate eyes on that.

## PR checklist: submitters

## PR checklist: reviewers


backend/dissemination/api/api_v1_0_3/create_functions.sql
```diff
@@ -13,48 +13,61 @@
...
 end;
 $get_jwt_claim$ language plpgsql;
 
-create or replace function get_header(item text) returns text
+create or replace function api_v1_0_3.get_header(item text) returns text
 as $get_header$
 begin 
-	raise info 'request.header % %', item, getter('request.header', item);
-	return getter('request.header', item);
+	-- raise info 'request.header % %', item, getter('request.header', item);
```
danswick, line 33:
```suggestion
```

backend/dissemination/api/api_v1_0_3/create_functions.sql
```diff
@@ -13,48 +13,61 @@
...
 -- https://api-umbrella.readthedocs.io/en/latest/admin/api-backends/http-headers.html
--- I'd like to go to a model where we provide the API keys. 
--- However, for now, we're going to look for a role attached to an api.data.gov account.
--- These come in on `X-Api-Roles` as a comma-separated string.
-create or replace function has_tribal_data_access() returns boolean
+-- Previously, we were using a role attached to the key.
+-- This now uses an explicit table to determine if a key has tribal data access.
+-- What this means is that we must go through a commit/PR process in order to use
+-- keys that will have access to tribal data. This should be supported by a 
+-- ticketing process.
```
danswick, line 43: It would be good to document the process now so it doesn't get lost.

backend/dissemination/api/api_v1_0_3/create_functions.sql
```diff
@@ -13,48 +13,61 @@
...
+            ELSE 0::BOOLEAN
+            END 
+        INTO key_exists;
+    
+    RETURN key_exists;
+END;
 $has_tribal_data_access$ LANGUAGE plpgsql;
 
-create or replace function has_public_data_access_only() returns boolean
+create or replace function api_v1_0_3.has_public_data_access_only() returns boolean
```
danswick, line 65: Is this function used anywhere?

backend/dissemination/api/api_v1_0_3/create_access_tables.sql
```diff
@@ -0,0 +1,22 @@
+-- This is explicitly not a Django managed table.
+-- In order to grant an API key access to read tribal data
+-- we need to go through a PR.
```
danswick, line 3: This comment isn't quite correct, right? An admin user can also grant access via API?

danswick (commented):
Added a few comments inline, but still have a few files to get through.

backend/dissemination/api/api_v1_0_3/create_functions.sql
jadudm: I'll add a ticket template.

backend/dissemination/api/api_v1_0_3/create_functions.sql
jadudm: Possibly not. We can remove it.

backend/dissemination/api/api_v1_0_3/create_access_tables.sql
jadudm: No. The only way to get access tot he API is to have your key's UUID added to the access table. The text says "grant an API key access," not "grant a user access."

To gain access as a user, we use the API. To use the API, we commit to this table.

backend/dissemination/api/api_v1_0_3/create_functions.sql
jadudm: A ticket template has been added; comment updated.

backend/dissemination/api/api_v1_0_3/create_functions.sql
jadudm: No. Removed.

danswick (approved):
We reviewed as a group and this should be good to go once tests pass.

PR #2863, opened 2023-11-26
2861 update logic to use elecaudits model

sambodeme:
## PR checklist: submitters

## PR checklist: reviewers


backend/census_historical_migration/workbooklib/excel_creation_utils.py
```diff
@@ -61,11 +69,63 @@ def set_range(wb, range_name, values, default=None, conversion_fun=str):
...
+                    )
+                ),
+            }
+        )
+    return ranges
+
+
+# FIXME-MSHD: Remove this function once we switch all workbook generators to using census models
 def set_uei(Gen, wb, dbkey):
     g = Gen.select().where(Gen.dbkey == dbkey).get()
```
JeanMarie-TTS, line 130: Please use django ORM

backend/census_historical_migration/sac_general_lib/report_id_generator.py
```diff
@@ -1,8 +1,13 @@
-from census_historical_migration.sac_general_lib.utils import _census_date_to_datetime
+from census_historical_migration.sac_general_lib.utils import (
```
JeanMarie-TTS, line 1: Please use relative imports within the same app

backend/census_historical_migration/workbooklib/excel_creation_utils.py
sambodeme: This code is not being used as part of this PR. I added a comment on top to signal that it will be removed (but not as part of this PR). It is now be replaced by `set_workbook_uei`

backend/census_historical_migration/test_excel_creation.py
```diff
@@ -143,3 +152,83 @@ def test_set_range_ws_missing(self):
         wb.defined_names.add(defn)
 
         self.assertRaises(KeyError, set_range, wb, self.range_name, ["bar"])
+
+
+class TestApplyConversionFuncTion(TestCase):
```
danswick, line 157:
```suggestion
class TestApplyConversionFunction(TestCase):

```

danswick (approved):
Left one suggestion and let's handle the relative imports discussion later all at once. If we want to enforce relative imports as a matter of style, we can probably add that to one of the linters as a rule.

PR #2868, opened 2023-11-27
Command to run census migration by year

phildominguez-gsa:
In this PR:
- Splitting out code from https://github.com/GSA-TTS/FAC/pull/2811 to create a command that allows you to run submissions for a given year. This does not include any of the db/model changes.
- `dynamic_import()` works with years given with more than 2 digits
- I did some renaming, docstrings, etc.

Testing:
- `docker compose run --rm web python manage.py run_migration_for_year --year 2022`
  - This should iterate over submissions for 2022 (which is a lot, so it won't complete any time soon)

## PR checklist: submitters

## PR checklist: reviewers


backend/census_historical_migration/make_submission.py
```diff
@@ -0,0 +1,44 @@
...
+    return result
+
+
+def create_or_get_user(result, gen):
+    """Returns the user for the submission if found, or a default user otherwise"""
+    user_email = gen.AUDITEEEMAIL
+    user_name = gen.AUDITEECONTACT.split()[0] + "_generated"
+
+    if not user_email or len(user_email) == 0:
+        user_email = "loader_generated_email@history_data.org"
```
jadudm, line 26: For the future, this needs to be an impossible address. Something `@fac.gsa.gov` should be safe from being able to be claimed via `login.gov`.

backend/census_historical_migration/make_submission.py
```diff
@@ -0,0 +1,44 @@
...
+    gen = ELECAUDITHEADER.objects.get(AUDITYEAR=audit_year, DBKEY=dbkey)
+    user = create_or_get_user(result, gen)
+    run_end_to_end(user.email, dbkey, audit_year)
+
+    return result
+
+
+def create_or_get_user(result, gen):
+    """Returns the user for the submission if found, or a default user otherwise"""
+    user_email = gen.AUDITEEEMAIL
```
sambodeme, line 22: We decided to use a default user access for the migration process. We should have an ADR on this soon. For now, you can use [this](https://github.com/GSA-TTS/FAC/blob/main/backend/census_historical_migration/sac_general_lib/sac_creator.py#L59) so we remain consistent. Also instead of `auditee.org` use `fac.gsa.gov`. I have a follow up PR to do the same change on the other side

backend/census_historical_migration/make_submission.py
sambodeme, line 26: Same as above

backend/census_historical_migration/historic_data_loader.py
```diff
@@ -0,0 +1,30 @@
...
+
+def load_historic_data_for_year(audit_year):
+    """Iterates over and processes submissions for the given audit year"""
+    result_log = {}
+    total_count = error_count = 0
+
+    gens = Gen.objects.filter(AUDITYEAR=audit_year)
+    for gen in gens:
+        dbkey = gen.DBKEY
+        result = load_historic_data(audit_year, dbkey)
```
sambodeme, line 13: We may want to add a comment here since it is the entry point for the process. So, the rest of the code ignore the audit_year parameter because it is assumed that we do migration year-by-year. That is at no point there will be audit data for more than a year in the historic database. Let's capture this here. We should probably have something in place to ensure this assumption remains true before each migration cycle (this last part is not necessary related to your PR).

backend/census_historical_migration/make_submission.py
```diff
@@ -0,0 +1,44 @@
...
+
+from django.contrib.auth import get_user_model
+
+
+User = get_user_model()
+
+
+def load_historic_data(audit_year, dbkey):
+    """Create a submission for the given audit year and dbkey"""
+    result = {"success": [], "errors": []}
```
sambodeme, line 12: It does not seem that `errors` is being used in create_or_get_user and run_end_to_end.

sambodeme: It is probably okay to have a light weight error check here. Later in the process we'll craft a more refined mechanism to catch errors and record in the database

jadudm: There's a tough design decision that always plagues this kind of software: throw exceptions, or return errors?

In Go, the practice is to return `error` objects. In Java, it's exceptions. Python... could go either way.

It is possible that an exception-based mechanism will allow y'all to short-circuit a migration that is going to be a "failure," but it might also be that we want to attempt a full migration of every record, and then record *all* the ways that record failed, so that we have more data for the next round. 

Something to noodle on.

backend/census_historical_migration/historic_data_loader.py
JeanMarie-TTS: > Per conversation, we decided that we'll thread the year through, because loading historical data year-by-year (e.g. remove 2022, then load 2021...) would be annoying in production. It might be best to assume all of the historical data is loaded, and we select what we need from it. Whether that is for this PR or a future was left for the group to decide as to what is best.

Agree!

backend/census_historical_migration/workbooklib/census_models/census.py
```diff
@@ -14,7 +14,7 @@ def model_module_path(model, year):
```
JeanMarie-TTS, line 15: Do we still need census_models? The models are already present in models.py. 
Besides, we want to avoid dynamic imports if we can.

sambodeme: Not needed but it can't be removed as part of this PR.

backend/census_historical_migration/historic_data_loader.py
sambodeme: This [ticket ](https://github.com/GSA-TTS/FAC/issues/2879) is to revisit the entire migration logic and enforce audit_year in all queries.

backend/census_historical_migration/make_submission.py
phildominguez-gsa: Done!

backend/census_historical_migration/make_submission.py
phildominguez-gsa: Done!

backend/census_historical_migration/make_submission.py
phildominguez-gsa: Done!

backend/census_historical_migration/historic_data_loader.py
phildominguez-gsa: Renamed `gens`/`gen` and added a comment

sambodeme (approved):
This is a good starting point. We'll continue refining it alongside the overall migration code.

PR #2933, opened 2023-12-06
2879 enforce audit year in all queries

sambodeme:
## PR checklist: submitters

## PR checklist: reviewers


backend/census_historical_migration/management/commands/historic_data_migrator.py
```diff
@@ -1,62 +1,65 @@
...
+                    continue
+
+                run_end_to_end(user, audit_header, result)
+                logger.info(result)
+
+    def handle(self, *args, **options):
+        dbkeys_str = options["dbkeys"]
+        years_str = options["years"]
 
         if ENVIRONMENT in ["LOCAL", "DEVELOPMENT", "PREVIEW", "STAGING"]:
```
JeanMarie-TTS, line 59: I don't think we should have STAGING in this list. STAGING should be treated like PRODUCTION.

backend/census_historical_migration/management/commands/historic_data_migrator.py
```diff
@@ -1,62 +1,65 @@
...
+)
+from census_historical_migration.workbooklib.excel_creation_utils import (
+    get_audit_header,
+)
 from census_historical_migration.historic_data_loader import create_or_get_user
-
-from config.settings import ENVIRONMENT
-from django.core.management.base import BaseCommand
 from census_historical_migration.workbooklib.end_to_end_core import run_end_to_end
+from config.settings import ENVIRONMENT
```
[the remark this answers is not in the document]
sambodeme: @JeanMarie-TTS I reviewed your PR, once it gets merged, it will clear this one (after the branch is updated).

backend/census_historical_migration/management/commands/historic_data_migrator.py
sambodeme: I believe this list was originally created when the command was used for end-to-end testing, with the rationale that one should not run tests in prod. However, the command is now being used for data migration, and we might need to use it in staging. While the command itself will likely evolve in the future, I am not sure if we should exclude it from staging at this point.

sambodeme: I believe this list was created when the code was used for end_to_end testing and the idea was one should not run end_to_end tests in production but now the command is being used for data migration and we may want to use that in staging. The command itself will probably change in the furture but we I am not sure if we should exclude this from staging.

gsa-suk:
LGTM.

JeanMarie-TTS (dismissed):
Let's queue up a discussion on how to treat `STAGING`. Okay to proceed for now.

PR #3016, opened 2023-12-18
Set head_ref to an env var

asteel-gsa:
Based on the following [example](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable)

I believe that this will work. Github suggests using the event payload in an env var, instead of directly referencing it. This _should_ patch the allstart issue. _I think_


jadudm (approved):
LGTM.

Issue #2924, opened 2023-12-05
Add export button to the search results page


timoballard:
Add an export button to the search results page. Clicking the button invokes the export view, triggering a download of a spreadsheet containing the disseminated data for all reports in the search result set.

PR #3025, opened 2023-12-18
Search Results - Summary Report Generation

jperson1:
# Search Results - Summary Report Generation

Best Fit Issue: https://github.com/GSA-TTS/FAC/issues/2924

## Changes:
1. Summary report generation for search results.
    1. There's a button, and it resubmits the form to a new URL & view.
    2. The new view re-runs the search and uses the results to build the summary spreadsheet. Then the user is given the file.
    3. The user clicks the button and receives a file. There's no redirection, so it may seem as though nothing is happening for a few seconds.
    4. The summary report generation is not terribly fast, so it's limited to 1000 results for now. That may still be too high.
2. Minor change for summary report generation efficiency (less total queries). 
3. Summary reports come with a coversheet note about the limit and a link to the API, in the same style as the pre-dissem sheets. 
4. Tests on the `MultipleSummaryReportDownloadView`
    * It mocks the file upload and download portions. 
    * It doesn't test the actual file, just some of the expected outputs of the view.
6. A lot of comments, formatting, and refactoring of the search view.
    * The call of `search_general` and the form cleanup have moved into their own functions.

The record limit on downloads comes from `settings.SUMMARY_REPORT_DOWNLOAD_LIMIT`. That value is used in the views and HTML. 

## How to test:
2. Ensure you have data in the dissemination tables.
    * There are sql dumps available on GDrive.
3. Make a search. Verify the "Export to XLSX" button looks right.
7. Verify the "Export to XLSX" button works.
    * Small result sets should come back in under a few seconds.
    * Large result sets are capped at 1000 records.
    * A max size result set takes about 15 seconds to generate on my machine. 

## Screenshot:
image

## PR checklist: submitters

## PR checklist: reviewers


jperson1:
Draft-ing for a bit for conversations on copy and communicating download limits.

backend/dissemination/views.py
```diff
@@ -225,10 +279,35 @@ def get(self, request, report_id, file_type):
...
+        1. Get the report_ids from a fresh search with the provided search parameters
+        2. Generate a summary report with the report_ids, that goes into S3
+        3. Redirect to the download url of this new report
+        """
+        form = SearchForm(request.POST)
+
+        cleaned_data = clean_form_data(form)
+        include_private = include_private_results(request)
+        results = run_search_general(cleaned_data, include_private)
+        results = results[:1000]  # Hard limit XLSX downloads to 1000 records
```
timoballard, line 307: can we wrap this 1000 in a const (e.g. `SUMMARY_REPORT_MAX`) that we use here and pass into the `search.html` context for deciding whether or not to enable the download button, so that we can adjust that number in a single place?

backend/dissemination/views.py
```diff
@@ -225,10 +279,35 @@ def get(self, request, report_id, file_type):
...
+        redirect to its download link.
+        """
         sac = get_object_or_404(General, report_id=report_id)
         filename = generate_summary_report([sac.report_id])
         download_url = get_download_url(filename)
 
         return redirect(download_url)
+
+
+class MultipleSummaryReportDownloadView(View):
```
timoballard, line 295: I think it'd be good to have a couple of basic tests around this View - don't think we need to go so far as to inspect the resultant xlsx file in the test, but we should have some coverage around the various cases (what should this view return if no reports are found, some reports but less than the max, more than the max, etc)

jadudm:
The screenshots may be out of sync with what is in the code (haven't looked). We'll want to use the words "summary reports" instead of "bulk download." These were summary reports in the previous system; "bulk download" is language that we might want to reserve for database dumps in the future.

We also need to be careful about referencing the API here. It is *not* end-user friendly. The language might want to be "Programmers can also access the data via the FAC API," or something similar. While there are ways to access it via RPA and some Excel extensions, the API is not something that can casually be used by just anyone.

I'd suggest (at the least) standardizing on "summary report" and not "bulk download," and consider removing the API reference here, or being clearer as to whom the API serves.

lauraherring:
Sounds good, @jadudm 

So to clarify, @jperson1: change the subheading in the info box from "Bulk download" to "Summary reports" and then delete the last sentence entirely, that refers to the API.

jperson1:
@jadudm @lauraherring I've updated the relevant copy, and an updated screenshot is included in the description.

@timoballard I've pushed a bundle of changes that include making the download limit a settings variable (`settings.SUMMARY_REPORT_DOWNLOAD_LIMIT`) and a bunch of light testcases.

jperson1:
Per convo with @lauraherring and @Leighdiddy, the infobox info is all in one box, all the time. Further changes post-break.

image
image

timoballard (approved):
Looks great - nice work @jperson1!

lauraherring:
Looks great from a content perspective!

PR #3071, opened 2023-12-21
Update end-to-end dependencies and workflow, refactor upload specs

danswick:
This branch:

* removes a third party GH action that was unexpectedly running tests Node 16, while the other third party action was running cypress on Node 20.
* Rebuilt package-lock.json using Node 20.
* Lightly refactors the upload POST request assertions.
* Updates cypress and cypress-adjacent dependencies. 

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 20
 
       - name: Create .env file
         run: touch .env
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -38,7 +38,7 @@ jobs:
       CYPRESS_LOGIN_TEST_OTP_SECRET_AUDITEE: ${{ secrets.CYPRESS_LOGIN_TEST_OTP_SECRET_AUDITEE }}
       DISABLE_AUTH: False
       # useful for debugging but otherwise complete overwhelm
-      DEBUG: cypress:server:util:process_profiler
+      # DEBUG: cypress:server:util:process_profiler
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -38,23 +38,25 @@ jobs:
...
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
       - name: Create .env file
         run: touch .env
 
       - name: Run Cypress full submission test
-        uses: Wandalen/wretry.action@master
```
asteel-gsa, line 50: Keep in mind, this action was added to assist with the 60000ms timeout errors that persisted through the cypress workflow. If that is no longer an issue, it can be removed, but this is just a loop that runs 5 times.

danswick: I think the retries may have been hiding some of the underlying flakiness in our e2e tests, which has now gotten to the point where even retrying 5 times doesn't seem to help. 

I initially removed the retry action so I could test a Node compatibility issue: because the cypress tests are nested under `wretry.action`, they're kicked off using wretry's version of Node (16), which also installs npm dependencies. But the _cypress_ action is hardcoded to use Node 20, which is unfortunate and resulted in some missing dependency errors in the logs.

Anyway, consistifying all of the Node versions does seem to have resolved some warnings, but the tests are still flaky in new and exciting ways.

Short term, I think we should just simplify the cypress test that confirms files were uploaded: catch the 302 response from the server, but don't bother checking for corresponding page updates.

Medium term, we need to refactor the Cypress tests to speed them up and make sure each test is actually testing what we want. Cypress also supports more granular retry functionality directly, but I'm not sure it would work well with the current full_submission spec.

asteel-gsa: Yup! Makes sense to me. If we aren't seeing the timeouts we were, then get rid of it. Thanks for the explanation.

eazapata23:
Looks good, only linting issues

tadhg-ohiggins (approved):

jperson1 (approved):

Issue #2950, opened 2023-12-07
Record transformations in MigrationChangeRecord table


gsa-suk:
When a migration completes successfully,
update MigrationChangeRecord table with information about the transformations performed.

```[tasklist]
### Tasks
- [ ] https://github.com/GSA-TTS/FAC/issues/3040
- [ ] https://github.com/GSA-TTS/FAC/issues/3046
- [x] https://github.com/GSA-TTS/FAC/issues/3047
- [ ] https://github.com/GSA-TTS/FAC/issues/3048
- [ ] https://github.com/GSA-TTS/FAC/issues/3056
```

PR #3082, opened 2023-12-26
Census migration: Making `result` a module var

phildominguez-gsa:
Contributes to https://github.com/GSA-TTS/FAC/issues/2950

In this PR:
- This is for merging changes into `sk/2950_migration_change_record`, NOT into `main`
- Changes `result` to be a module variable that can be imported wherever needed, instead of repeatedly passing it around as a parameter
- Organizing some imports

Testing:
- Performing a migration should still yield results in the `migrationchangerecord` table

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/census_historical_migration/historic_data_loader.py
+++ b/backend/census_historical_migration/historic_data_loader.py
@@ -1,6 +1,6 @@
 from .models import ELECAUDITHEADER as AuditHeader
 from .workbooklib.end_to_end_core import run_end_to_end
-from census_historical_migration.migration_result import result
+import census_historical_migration.migration_result as migration_result
 
 from django.contrib.auth import get_user_model
 from django.core.paginator import Paginator
@@ -28,14 +28,15 @@ def load_historic_data_for_year(audit_year, page_size, pages):
         )
 
         for submission in page.object_list:
-            result = {"success": [], "errors": [], "transformations": []}
+            migration_result.result = migration_result.DEFAULT_RESULT
+
             # Migrate a single submission
             run_end_to_end(user, submission)
 
-            result_log[(submission.AUDITYEAR, submission.DBKEY)] = result
+            result_log[(submission.AUDITYEAR, submission.DBKEY)] = migration_result.result
             total_count += 1
 
-            has_failed = len(result["errors"]) > 0
+            has_failed = len(migration_result.result["errors"]) > 0
             if has_failed:
                 error_count += 1
             if total_count % 5 == 0:
--- a/backend/census_historical_migration/historic_data_loader.py
+++ b/backend/census_historical_migration/historic_data_loader.py
@@ -31,7 +31,9 @@ def load_historic_data_for_year(audit_year, page_size, pages):
             # Migrate a single submission
             run_end_to_end(user, submission)
 
-            result_log[(submission.AUDITYEAR, submission.DBKEY)] = MigrationResult.result
+            result_log[
+                (submission.AUDITYEAR, submission.DBKEY)
+            ] = MigrationResult.result
             total_count += 1
 
             has_failed = len(MigrationResult.result["errors"]) > 0
--- a/backend/census_historical_migration/historic_data_loader.py
+++ b/backend/census_historical_migration/historic_data_loader.py
@@ -1,6 +1,6 @@
 from .models import ELECAUDITHEADER as AuditHeader
 from .workbooklib.end_to_end_core import run_end_to_end
-import census_historical_migration.migration_result as migration_result
+from .migration_result import result
 
 from django.contrib.auth import get_user_model
 from django.core.paginator import Paginator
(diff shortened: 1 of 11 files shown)


backend/census_historical_migration/historic_data_loader.py
```diff
@@ -1,5 +1,6 @@
 from .models import ELECAUDITHEADER as AuditHeader
 from .workbooklib.end_to_end_core import run_end_to_end
+from census_historical_migration.migration_result import result
```
sambodeme, line 3:
```suggestion
from .migration_result import result
```

sambodeme (commented):
Not sure why I cannot find `migration_result.py`

backend/census_historical_migration/historic_data_loader.py
```diff
@@ -27,14 +28,15 @@ def load_historic_data_for_year(audit_year, page_size, pages):
         )
 
         for submission in page.object_list:
-            result = {"success": [], "errors": [], "transformations": []}
```
sambodeme, line 30: We want to reset the result fields here otherwise, it will keep appending data to the previously run submission results. It could also be done in `end_to_end_core.py` at the beginning of run_end_to_end (which will make more sense)

backend/census_historical_migration/workbooklib/end_to_end_core.py
```diff
@@ -297,7 +298,7 @@ def run_end_to_end(user, audit_header, result):
             audit_header.DBKEY,
             sac.report_id,
             result["transformations"],
-            )
+        )
```
sambodeme, line 301: Another place were we could reset the result fields before processing the next submission.

sambodeme: We may want to reset here regardless of previous reset to ensure the last submission processed does not leave unnecessary data in memory

phildominguez-gsa:
> Not sure why I cannot find `migration_result.py`

Yep forgot to add it 😅

backend/census_historical_migration/historic_data_loader.py
phildominguez-gsa: `migration_result.result = migration_result.DEFAULT_RESULT` below is the new reset, but I'll try moving it to run_end_to_end

backend/census_historical_migration/historic_data_loader.py
phildominguez-gsa: Done

backend/census_historical_migration/historic_data_loader.py
phildominguez-gsa: Moved the reset to e2ec

backend/census_historical_migration/workbooklib/end_to_end_core.py
phildominguez-gsa: I'm not opposed but right now the commands use `result` to print stuff after it's done running. For now I'm just resetting at the start.

backend/census_historical_migration/workbooklib/end_to_end_core.py
```diff
@@ -250,10 +251,11 @@ def api_check(json_test_tables):
     return combined_summary
 
 
-def run_end_to_end(user, audit_header, result):
+def run_end_to_end(user, audit_header):
     """Attempts to migrate the given audit"""
     try:
-        sac, result = setup_sac(user, audit_header, result)
+        migration_result.result = migration_result.DEFAULT_RESULT
```
sambodeme, line 257: Does this work ? I would expect these two objects to point to the same memory location due to `result = DEFAULT_RESULT` in migration_result.py and therefore here instead doing a reset it just reaffirms that `migration_result.result` is pointing to the now-modified `DEFAULT_RESULT`

backend/census_historical_migration/migration_result.py
```diff
@@ -0,0 +1,5 @@
+"""Stores the result for the current migration"""
+
+DEFAULT_RESULT = {"success": [], "errors": [], "transformations": []}
+
+result = DEFAULT_RESULT
```
sambodeme, line 5: How about something like 
```
class MigrationResult:
    DEFAULT_RESULT = {"success": [], "errors": [], "transformations": []}
    result = DEFAULT_RESULT.copy()

    @staticmethod
    def reset():
        MigrationResult.result = MigrationResult.DEFAULT_RESULT.copy()

    @staticmethod
    def update_transformation(data):
        MigrationResult.result["transformations"].append(data)
    ....................an update function for errors and success .... or something similar

backend/census_historical_migration/workbooklib/end_to_end_core.py
phildominguez-gsa: Yep, now using copy.deepcopy()

backend/census_historical_migration/migration_result.py
phildominguez-gsa: Done (I called them `append_*` to make sure it's clear)

gsa-suk (commented):
LGTM.

PR #3194, opened 2024-01-12
Resolve new User Accesses immediately, if possible

### LINKED ISSUE #3142: Improvement: Resolve user accesses
## User story

A user (Alice) modifies someone's access via "Manage users." E.g. they change someone (e.g. Bob) from an editor to a certifying official. (Or, they attach a certifying official where it had previously been left blank.)

If Bob is currently logged in, they will still receive 403 errors when they attempt to hit the auditor certification page. Why? Because their accesses are only refreshed at login. So, Bob would need to log out and log back in again.

## Possible fixes

We would prefer that when Alice changes Bob's status, the effect is immediate.

1. We could update Accesses on every page load. This feels expensive.
2. We could have signals, and do it async via a middlware. This feels complex.
3. We could update on the save of the accesses on the manage users page. This feels just right.

We should explore/spike #3 and determine whether this eliminates user confusion around the updating of user accesses.

timoballard:
Resolves #3142 

Small update to user the `create` method rather than `save` when creating `Access` objects via the manage submission access page. Using `save` meant that we were not utilizing our overridden `create` method that attempts to immediately map the `Access` object to a `User` object, if one exists with a matching email address.

To test locally:
1. Create a new submission
2. Login to the FAC with a different email address in a private browser window
3. In the original window, go to Manage access, add the alternate email address as an editor
4. In the private window, refresh the My submissions page --> should see the new submission listed without having to logout/login

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/audit/models/access.py
+++ b/backend/audit/models/access.py
@@ -30,7 +30,6 @@ class AccessManager(models.Manager):
         event_type = obj_data.pop("event_type", None)
 
         # try to pair this Access with an actual User object if we have one for this email address
-        print(obj_data)
         if obj_data["email"]:
             try:
                 acc_user = User.objects.get(email__iexact=obj_data["email"])
--- a/backend/audit/models/access.py
+++ b/backend/audit/models/access.py
@@ -30,9 +30,10 @@ class AccessManager(models.Manager):
         event_type = obj_data.pop("event_type", None)
 
         # try to pair this Access with an actual User object if we have one for this email address
+        print(obj_data)
         if obj_data["email"]:
             try:
-                acc_user = User.objects.get(email=obj_data["email"])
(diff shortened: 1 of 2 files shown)


tadhg-ohiggins (approved):
Looks good.

I think there's an edge case here where there's more than one match, because a user has abandoned a login.gov account that matches the email address and created a new one, and effectively created multiple accounts in our system that have the same email address.

timoballard:
@tadhg-ohiggins agreed - I think in that case we have a `1 / <number_of_accounts>` chance of automatically assigning to the correct `User`, which is kind of the best we can do, since we have no real way of knowing which of their accounts is the "active" one. In that case, it should sort itself out the next time they login

PR #3199, opened 2024-01-15
Fix Job Names

asteel-gsa:
Copy/Paste gone wrong


jadudm (approved):
LGTM. Would be good to test in `dev` when it goes in.

asteel-gsa:
> LGTM. Would be good to test in `dev` when it goes in.

its going to fix this, thats all. doesnt impact anything else
image

PR #3256, opened 2024-01-23
Hack 16 seconds off federal awards validation time by adding it to app startup time

tadhg-ohiggins:
So it’s really slow  
To validate all the things—  
Do it at startup  

-----

We currently load our templates into openpyxl and into our intermediate representation every time we run `has_all_the_named_ranges`, which we run on every uploaded Excel file. We do this just to get the names of the ranges. However, these names do not change between user uploads.

This PR changes the logic to go through the above process on app startup instead.

There is almost undoubtedly a better way to do this by reading the JSON instead, but this approach already cuts about 16 seconds off of uploads, so it’s worth doing it this crude way first and improving on it later.

This will cut validation time for _all_ workbooks; it's just that I only timed federal awards.

This also significantly increases application startup time, by probably 40 seconds, but that's a reasonable trade for the benefits and we will improve it later.

## PR checklist: submitters

## PR checklist: reviewers


danswick:
I'm testing out the jsonschema approach we discussed over in this branch/PR: https://github.com/GSA-TTS/FAC/pull/3257

backend/audit/intakelib/checks/check_has_all_the_named_ranges.py
```diff
@@ -6,46 +7,68 @@
...
-# 3. Compare the template named ranges to the IR
+# 1. Extract list of expected named ranges from our rendered jsonschema.
+# 2. Compare the jsonschema named ranges to the IR
 #
 # If we're missing anything, we need to bail immediately. They have mangled the workbook.
 # We should not accept the submission.
+TEMPLATES = {}
+for section in map_section_to_workbook:
+    template = load_workbook(map_section_to_workbook[section], data_only=True)
+    template_ir = extract_workbook_as_ir(template)
```
danswick, line 64:
```suggestion
for section in map_section_to_jsonschema:
    template = load_workbook(map_section_to_jsonschema[section], data_only=True)
    template_ir = extract_workbook_as_ir(template)
```

danswick (approved):

danswick:
Closing in favor of https://github.com/GSA-TTS/FAC/pull/3274

Issue #3252, opened 2024-01-22
Epic: Database performance


jadudm:
## Background

We have approximately 12K audits in the dissemination table. This appears to be enough data to slow down:

1. The application
2. API queries
3. Web queries

We have suspicions that the DB is a bottleneck; that is, when the API is under load, it appears that other parts of the system slow down. This is commesurate with our `medium` database instance being heavily loaded. (We saw this in the DB migration work: the `medium` instance is 8GB of RAM on 1 VPU.) The performance DB is also a `medium` instance, and therefore we have no reason to believe it performs better than the `preview` instance doing migration work.

In `preview`, we did two things:

1. Upgrade to an `xlarge-redundant` instance. This provides 16GB of RAM and 2 VPUs to the DB.
2. Index critical columns or sets of columns in the dissemination tables. 

```[tasklist]
### Recommended actions
- [x] If possible, determine where the bottlenecks are.
- [ ] Identify and apply indicies to relevant columns
- [ ] Upgrade DB instance (during a low period) from `medium-redundant` to `xlarge-redundant`
- [x] Fix search so it does not crash on most/all queries.
```

The research needs to be timeboxed; we may, because of the cloud.gov architecture, have poor visibility into the DB server's performance bottlenecks.

Upgrading should be straight-forward, and involves a short downtime while the migration happens at the AWS end.

Relevant columns are at least:

1. Columns used in the WWW search interface
2. Columns that seem to be used heavily in the API queries

#1 is easy to determine by inspection of our search code.

#2 might be determinable from logs. If not, `report_id` is obvious in all tables.  

## We'll know we're done when: API test

Currently, it is not possible to reliably request 100 elements from the `passthrough` table via API. E.g.

```
/passthrough?select=report_id&limit=100&offset=100000
```

will either succeed (after a 120s+ delay) or, sometimes, fail with a 502. 

It is possible, but unlikely, that the issue is the PostgREST container not having enough RAM, but given the performance issues seen across the application, it is probably the DB.

## We'll know we're done when: WWW test

With 120K audits in the system (migrated data in `preview`), we see that a search for ALNs crashes the app.

```
93
45
```

as two agency numbers is enough to bring down the app after a timeout. This should be performant, and always succeed.

jperson1:
Some quick notes for findings on Search:

[...]

PR #3283, opened 2024-01-25
Speed up search queries

jadudm:
## Background

Addresses https://github.com/GSA-TTS/FAC/issues/3252.

As we are about to load historical data (~160K more audits), our query performance is now a bottleneck. Here, a "bottleneck" means "most searches crash the app."

This PR is *necessary* in order to load the historical data. Things to note:

1. We have no E2E/Cypress coverage on searches. We have no way of automatically confirming, for a given set of search parameters, and a known DB, that a known set of results will come back.
    1. We have no way of loading a repeatable set of data for testing, locally or in our production environments, at this time. So, #1 is difficult. 
2. This branch passes `make docker-test` and Cypress E2E, but that does not strictly address changes to search implementation.

We have tested manually, and we believe that we are returning the correct results still. 

* The `general` search did not change, so we have high confidence there.
* Much of the work around ALN search had to be rewritten. It is now clearer, better commented.
* We manually checked ~10 audits for correctness. We can confirm that the "my ALN" and "other ALN" fields are being annotated correctly.

The correct way to test this would be to write code against the API that checks these annotations. (E.g. Given an ALN, retrieve all the report IDs, calculate the Y/N fields, and then integrate that into Cypress as a confirmation of what comes back in the WWW view.) We have not done this at this time, as it would represent a complete reimplementation of those annotations in Python, without the ORM. 

## Performance

With this change, a search against the complete historical data (locally) for ALN 93 will take ~7.5 seconds (before rendering, exact value may vary based on local machine), and yield ~71K results. We consider this good performance, as it is returning half the DB in under 10 seconds, with annotations.

## Indexing

We commented out possible indexes. We left them in the code for further investigation.

Our initial testing (using the API) as well as wall clock tests suggest that indexing apparently relevant columns in `General` and `FederalAward` achieved *no speedup*. This feels *very wrong*. However, our performance without them seems to be acceptable, and it can be the subject of further investigation.

For this reason, there are no migrations in this work. Indexes would introduce migrations.

## Snaps or it didn't happen

image

## PR checklist: submitters

## PR checklist: reviewers


backend/dissemination/search_general.py
```diff
@@ -0,0 +1,165 @@
...
+
+    ##############
+    # Initialize query.
+    # This is where tribal/is_public is set by default.
+    q_base = _initialize_query(params.get("include_private", False))
+    r_base = General.objects.filter(q_base)
+
+    ##############
+    # Audit years
+    # query.add(_get_audit_years_match_query(audit_years), Q.AND)
```
jadudm, line 23:
```suggestion
```

backend/dissemination/views.py
```diff
@@ -140,25 +144,42 @@ def post(self, request, *args, **kwargs):
...
+        time_starting_post = time.time()
+
         form = SearchForm(request.POST)
         results = []
         context = {}
 
         form_data = clean_form_data(form)
+
+        # If no years are checked, check 2023.
+        logger.info(form_data)
```
jadudm, line 156:
```suggestion
```

tadhg-ohiggins (approved):
Reviewed with @jadudm and @jperson1

PR #3333, opened 2024-02-01
Context for 500s, Resolve Pre-cert Report 500s

jperson1:
# Context for 500s, Resolve Pre-cert Report 500s

## Background:
Users have indicated via the help desk that pre certification SF-SAC downloads send them to a 500 error page. The page in question also displays the test site banner, despite the interaction occurring in production. 

Issues:
1. The test site banner appears on all 500 error pages. There are other things wrong too, because Django 500s default to containing no context at all. See [here](https://docs.djangoproject.com/en/5.0/ref/views/#the-500-server-error-view). 
2. The SF-SAC downloads are failing. Upon generation, it attempts to add data to a row from all columns in a given list. Some of these columns don't exist pre-certification, and are generated from disseminated data. So, the SF-SAC generator tries to fill in data that doesn't exist and breaks. 

## Changes:
1. Include a custom error handler for 500s that does very little. It exists to ensure that certain context processors run. This will mean that 500 errors pages properly include login info, the OMB values, and the test site banner. 
3. Update `gather_report_data_pre_certification()` to ensure it does not try to access/generate underscored fields. Underscored fields exist to add additional columns to workbooks for the benefit of certain users post-submission. Therefore, they're not necessary here.

## How to test:
First, verify that the issues are present in the first place.
1. Make sure the Django `settings.py` DEBUG variable is set to False. We want to see the error page. 
2. Run on main normally, and navigate to a pre-certification SF-SAC download. Try to download it. 
4. Verify that download fails and sends you to the 500 page.
5. Verify that the 500 page has the test banner, has no OMB values, and displays as though you are not signed in.

Second, switch to this branch and run normally.
1. Make sure the Django `settings.py` DEBUG variable is still set to False. We want to see any error pages. 
2. Run on this branch normally, try the download again. Verify that it works. 
3. To verify that the 500 page changes are working, we want to force a 500 to happen locally. You can do this by:
    a. Undoing the last commit to `dissemination/summary_reports.py`, essentially undoing the fix in this branch.
    b. Quickly implementing a test view that raises a 500. See [here](https://stackoverflow.com/a/24660486). 
    c. Modifying an existing view to raise a 500. 

## Screenshot:
500 error page, now:
image

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/config/error_handlers.py
+++ b/backend/config/error_handlers.py
@@ -1,0 +1,8 @@
+from django.shortcuts import render
+
+
+def handler500(request, template_name="500.html"):
+    """
+    Custom handler for 500s. Ensures the 500 template loads with proper context (environment, login info, OMB values).
+    """
+    return render(request, template_name, {})
--- a/backend/config/urls.py
+++ b/backend/config/urls.py
@@ -18,6 +18,8 @@ schema_view = get_schema_view(
     renderer_classes=[JSONOpenAPIRenderer],
 )
 
+handler500 = "config.error_handlers.handler500"
+
 urlpatterns = [
     path("api/schema.json", schema_view),
     path(
--- a/backend/dissemination/summary_reports.py
+++ b/backend/dissemination/summary_reports.py
@@ -424,11 +424,12 @@ def gather_report_data_pre_certification(i2d_data):
         for obj in dissemination_data[model_name]:
             row = []
             for field_name in field_names:
-                value = getattr(obj, field_name)
-                # Wipe tzinfo
-                if isinstance(value, datetime):
-                    value = value.replace(tzinfo=None)
-                row.append(value)
+                if not field_name.startswith("_"):
+                    value = getattr(obj, field_name)
+                    # Wipe tzinfo
+                    if isinstance(value, datetime):
+                        value = value.replace(tzinfo=None)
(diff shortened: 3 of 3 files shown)


tadhg-ohiggins (approved):
We should come back later to make sure the 500 change doesn't do anything we don't expect, but for now this should go out.

backend/config/error_handlers.py
```diff
@@ -0,0 +1,8 @@
+from django.shortcuts import render
+
+
+def handler500(request, template_name="500.html"):
+    """
+    Custom handler for 500s. Ensures the 500 template loads with proper context (environment, login info, OMB values).
+    """
+    return render(request, template_name, {})
```
timoballard, line 8: if we're passing an empty context dict here, where/how does the mentioned proper context come into play?

jperson1: Django's default 500 view explicitly renders the `500.html` template with no variables. By overriding it here (even if we don't give it anything in particular), we allow the template to render with the global context added by the functions in `context_processors.py`.

timoballard: makes sense- thanks @jperson1!

timoballard (approved):

PR #3348, opened 2024-02-06
Tadhg/dissemination race condition

tadhg-ohiggins:
Good indentation  
Isn’t just a nice-to-have—  
It’s fundamental.  

-----

**We do not report any error to the user in the case where we catch the `TransactionManagementError`**.

We are making the assumption that seeing that error where we do means that this specific race condition has occurred and that the first write has succeeded. We do check that the `dissemination.General` table has an entry with the corresponding `report_id`, but that may not be comprehensive enough.

-----

While investigating this error, we also found other issues, the most serious of which is that for quite some time [all of the `SubmissionViewTests` test cased below `test_unlock_after_certification` were incorrectly indented and thus inactive](https://github.com/GSA-TTS/FAC/pull/3348/files#diff-55b93de2a8465d6cceb8ddb886d49769971543a3fa95c06dc23152471d57d5adL430-R604).

This PR corrects that, and adds a number of other test corrections that catch those tests up to where other parts of the application have changed during the time they were inactive.

In addition, this PR changes an instance of `logging.warning` to `logger.warning` in [`intake_to_dissemination.py`](https://github.com/GSA-TTS/FAC/pull/3348/files#diff-6e5e23e622de844b8095e89f6e8b4aaedfb2e926e2f8d5770c500fabd2feaa6bL72-L74)

-----

I know that @jadudm, at least, has a setup for testing this bug with two local instances running; it would be a good idea to test this branch with that setup in case the tests in it don’t cover it as well as we hope they do.

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/audit/intake_to_dissemination.py
+++ b/backend/audit/intake_to_dissemination.py
@@ -35,7 +35,7 @@ class IntakeToDissemination(object):
         self.audit_year = int(audit_date.split("-")[0])
         self.loaded_objects: dict[str, list] = {}
         self.mode = mode
-        self.errors = []
+        self.errors: list[str] = []
 
     def load_all(self):
         load_methods = {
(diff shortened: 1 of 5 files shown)


jadudm (approved):
LGTM. Tested in a multi-instance environment. Only stores one set of data, avoiding the race at time of dissemination.

PR #3355, opened 2024-02-07
Don't create ever-expanding results in create_views 

jadudm:
This does not have a ticket.

Users have been reporting increasingly poor performance in the API. It was unclear what or why it was behaving badly.

Recently, a query to just return one record, offset 60K records into `federal_awards`, was reported as timing out by a Federal partner. Further investigation, with a full database, revealed that we had, in every view, something that looked like this:

```
    where
        (award.report_id = gen.report_id
         and
         gen.is_public = true)
        or (gen.is_public = false and api_v1_0_3_functions.has_tribal_data_access())
```

The second half of the `OR` clause is a huge problem.

It means the `SELECT` comes back for every condition under which the `OR` is `TRUE`. As a result, we are getting records back:

1. everywhere that the `report_id` is equal in both `federal_awards` AND
2. everywhere that `is_public` is `FALSE` and we are granted tribal access.

Somehow... and, to be honest, I'm still not fully clear on the combinatorics of this query... we were getting back 40M rows on a 2.5M row table. (A full count would time out; an estimated count, from the planner, would throw up a number in the range of 40M.)

The change was to shift all conditionals on the views to this:

```
    where
        gen.report_id = ein.report_id
         and
        (gen.is_public = true
         or 
        (gen.is_public = false and api_v1_0_3_functions.has_tribal_data_access()))
```

which now says 

"I would like the report ID to match, and I want the general table to either say the data is public, or I want to have permission to view it when the data is not public."

We will revise that further with Tribal API keys, but still... this is a critical change.

**This is likely part of our live DB performance issue**... we're crushing the DB engine every time someone hits the API. This may be part of where our overall DB/app performance penalties are coming from---every hit on the DB is competing with API queries in the worst possible way.

Note that this would _not_ be a load/performance hit on PostgREST, because the API server is able to quickly turn the URL into a query, and it sits and waits for the DB to come back. The DB, however, takes a lifetime trying to build up a result against tens of millions of (virtual?) rows.

Experimentation with indexing demonstrates that we can get at least a factor of 2 speedup, and there may be more work we can do with caching function results (but we have to be careful how we do that). However, once this fix is in place (and it needs more tests---I want this to leverage work that was done by PP and TB in that regard), the performance goes from "the query takes minutes to resolve" to queries likely taking a fraction of a second to complete.

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/dissemination/api/api_v1_0_3/create_views.sql
+++ b/backend/dissemination/api/api_v1_0_3/create_views.sql
@@ -16,10 +16,11 @@ create view api_v1_0_3.findings_text as
         dissemination_findingtext ft,
         dissemination_general gen
     where
-        (ft.report_id = gen.report_id
+        ft.report_id = gen.report_id
          and
-         gen.is_public = true)
-        or (gen.is_public = false and api_v1_0_3_functions.has_tribal_data_access())
+        (gen.is_public = true
+         or 
+        (gen.is_public = false and api_v1_0_3_functions.has_tribal_data_access()))
     order by ft.id
 ;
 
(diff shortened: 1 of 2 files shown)


jadudm:
From dev huddle, it appears that this has been fixed.

@timoballard , @purvinptl , close this as desired when your work comes through if your PR addresses it.

timoballard (dismissed):

timoballard (approved):

Issue #3441, opened 2024-02-22
Web search should properly disseminate Tribal data


timoballard:
Epic: #3331 

### Background
Because [only some](https://docs.google.com/document/d/1RNMOLAcgXk9NGX_05GMElqVR0-2c5HS4DPMKALlUB6I/edit) of the SF-SAC fields should be hidden from public view, we should be displaying tribal/private submissions in the web search results, even when the user is not authenticated or otherwise lacking permission to access Tribal data.

All of the tabular data in the search results is public, so nothing needs to be omitted there.

The summary view for a submission ([example](https://app.fac.gov/dissemination/summary/2023-09-GSAFAC-0000024841)) does contain protected data which will need to be omitted when accessed by unpermissioned users.

The PDF report is also restricted, so the download link should only be available for permissioned users.

### We're done when
- [ ] Tribal submissions appear in search results for all users
- [ ] PDF downloads work for users with the `read-tribal` permission
- [ ] PDF downloads do not work for users without the `read-tribal` permission
- [ ] Summary workbook downloads work, and are unabridged, for users with the `read-tribal` permission, and the workbook includes a disclaimer on the cover page
- [ ] Summary workbook downloads work, and omit [all protected fields](https://docs.google.com/spreadsheets/d/1DXkWcu3kuAOeQh4B3YrdiiHnIIc04QdLyzJwDLyjneg/edit#gid=0), for users without the `read-tribal` permission
- [ ] We have tests verifying all of the above

### Design considerations (@Leighdiddy):
- ~~How should we omit the PDF download link in cases where the user does not have tribal access permissions? Should we omit the link altogether, keep it there but present an access forbidden error if clicked, etc~~
  - Per @Leighdiddy's [comment](https://github.com/GSA-TTS/FAC/issues/3441#issuecomment-1965380828), we'll omit the link from the search results table

### Content considerations (@lauraherring):
- How should we word the disclaimer that we'll insert into the cover sheet of the unabridged summary workbook for permissioned users?

Leighdiddy:
@timoballard I think it would be best to fully omit the PDF download button from the summary view for users that don't have permission to access it, because showing it but then throwing an error message could confuse users. I could also see this (showing it but using an error message) resulting in more helpdesk tickets vs. not showing it. I'd like to make the difference as clear as possible for showing restricted data vs. regular public view data, which is tricky with using the same page layout so anything we can do to help visually differentiate between the two is a win.

[...]

PR #3466, opened 2024-02-29
3441 web search should properly disseminate tribal data

sambodeme:
## Description
This PR is to:
- Allow tribal submissions to appear in search results for all users
- Allow PDF downloads to work for users with the read-tribal permission
- Remove PDF download(s) button for users without the read-tribal permission
- See #3441 
- For testing purposes: Submit two reports, then make one private by unchecking the is_public flag in the dissemination_general table. Confirm that you can view both reports through the search interface, regardless of whether you are logged in. Also, verify that you can access the summary page for both reports but are unable to use the download button for the PDF of the private report.
- The change related to summary workbook will be part of a separate PR

-----
In this PR (Phil's stuff):
- Summary workbook downloads are unabridged for users with the read-tribal permission and omit [all protected fields](https://docs.google.com/spreadsheets/d/1DXkWcu3kuAOeQh4B3YrdiiHnIIc04QdLyzJwDLyjneg/edit#gid=0) for users without it
- Disclaimers added to all tribal-related cover sheets

Testing:
- Submit two reports, then make one private by unchecking the is_public flag in the dissemination_general table
- As a non-privileged user:
  - Go to http://localhost:8000/dissemination/search/ and perform a search that includes both reports
  - Click the Download all button
  - The resulting spreadsheet should have the tribal report_id rows missing from the `notes`, `captext`, and `findingtext` sheets, but all other (public) rows remaining
  - View the tribal audit and click the SF-SAC download button. The same rows should be removed here as well.
     - The cover sheet disclaimer should be as described [here](https://github.com/GSA-TTS/FAC/issues/3441#issuecomment-1973504324)
- As a privileged user
  - Repeat the steps above. All rows (including tribal report_id ones) should be there.
  - View the tribal audit and click the SF-SAC download button. All rows should be here as well.
    - The cover sheet disclaimer should be as described [here](https://github.com/GSA-TTS/FAC/issues/3441#issuecomment-1973504324)
 
## PR checklist: submitters

## PR checklist: reviewers


timoballard:
👏 

I think I underspecified the bit about the summary view page in #3441 -- I believe we need to omit the private data from that view (assuming no `read-tribal` privs) as well as from the actual summary report that gets downloaded. The screenshot below shows a private submission from my local testing -- I'm not able to download the SF-SAC summary report (👍), but on the summary view I'm still able to see that they have findings text, notes to sefa (0 in this case), and CAP. I believe the counts for those items are themselves private, so they should be conditionally removed from this view -- is that right @Carley-Jackson ?

image

jadudm:
Carley is OOO today; suppress the counts. 

If it turns out we don't have to, we can come back to it in a future iteration. Err on the side of caution.

sambodeme:
I was actually going to bring this up. My initial concern was about the fact that we are passing the records to the rendering template, even though we are not displaying them and are just showing the count. I am more inclined towards an approach where we just do the count on the backend side and pass through the result instead. However, given that the count related to private records could be deemed as private information as well, it may not be necessary to update this approach.

phildominguez-gsa (dismissed):

backend/dissemination/views.py
```diff
@@ -244,13 +248,13 @@ def get_audit_content(self, report_id):
...
         if audit_findings.exists():
             data["Audit Findings"] = [x for x in audit_findings.values()]
-        if audit_findings_text.exists():
+        if audit_findings_text.exists() and include_private_and_public:
             data["Audit Findings Text"] = [x for x in audit_findings_text.values()]
-        if corrective_action_plan.exists():
+        if corrective_action_plan.exists() and include_private_and_public:
             data["Corrective Action Plan"] = [
                 x for x in corrective_action_plan.values()
             ]
```
timoballard, line 260: the `include_private_and_public` checks here will result in each of these rendering with a count of 0 when the user lacks `read-tribal` permissions. Wondering if we need to differentiate between the actual 0 case (i.e. you're allowed to see the count and the count is 0) and the omission case (i.e. you're not allowed to see the count, which may or may not actually be 0). Put another way, is it a problem that in some cases the 0 shown here when the user lacks permission may incidentally be the actual value we're trying to hide? Do we need to show something like N/A when the user lacks permissions?

sambodeme: To your question, `Is it a problem that in some cases the 0 shown here, when the user lacks permission, may incidentally be the actual value we're trying to hide?` my answer is no. The only time a report would have 0 findings and 0 captext is when there are no findings associated with the report. However, the findings workbook is public for all reports (including private reports). That is, even if we display N/A instead of 0, people can still tell when 0 findings and 0 captext are expected just by seeing the findings workbook. Regarding `Note` records, this workbook is required for all reports and cannot be submitted empty (i.e., an empty note workbook will not pass validation). So, 0 notes is not a normal case. That said, using N/A instead of 0 is not a heavy lift, so if N/A is preferable here, I can make the update.

sambodeme: Made the change to display N/A instead of 0 for private counts.

jadudm:
Just for my own peace of mind:

"However, the findings workbook is public for all reports (including private reports)."

Are we absolutely 100% certain about that? (This is just me asking from a place of an abundance of caution.)

sambodeme:
> Just for my own peace of mind:
> 
> "However, the findings workbook is public for all reports (including private reports)."
> 
> Are we absolutely 100% certain about that? (This is just me asking from a place of an abundance of caution.)

Yes, findings workbook is supposed to be public based on the available to public column here https://docs.google.com/spreadsheets/d/1iS2B35gWE6j2EzmD2OfSEVU6y8xlQb5EL_ezNLzYzXo/edit#gid=507595660
FYI, the change in this PR is based on that `available to public column`

PR #3511, opened 2024-03-14
Jadudm/materialized views

tadhg-ohiggins:
If, with the literate, I quest  
To submit a pull request  
I never seek to take the credit  
We all assume that others‡ led it.  

-----

I reviewed this and didn’t actually write it (not my way of saying _nie mój cyrk, nie moje małpy_), but anyway, this:

*   Creates a new view in SQL to combine the General, Federal Wards, Findings, and Passthrough dissemination tables in one.
*   Creates a new unmanaged model in Django for the materialized view.
*   Adds Django management commands to create, delete, and regenerate the materialized view.
*   Uses the materialized view to search against (instead of the involved and slow queries we were using previously).

Effectively, this acts as a kind of cache, where we’re doing all the work required to combine those tables in advance, and then querying against the synthesized table for our searches in the web application, rather than doing the work to combine the tables every time a user searches.

‡: In this case, @jadudm and @sambodeme

```[tasklist]
### Next steps: 
- [ ] 1. Get concurrent refresh working on Preview. This requires a unique index on the created MV table.
- [ ] 2. (Probably) remove the automated create-on-deploy workflow. It will likely be best to create the MV post-deploy with an action. 
- [ ] 3. (Probably) remove / comment out the cron job in this PR. 
- [ ] 4. Create separate commands and actions for Drop/Create and Concurrent refresh.
- [ ] 5. Deploy to production and run Create during off-hours.
- [ ] 6. If the above is working as expected:
  - [ ] a. Enable refresh cron.
  - [ ] b. Create another PR with two search pages: Advanced, which looks like the current filter-ful search and is backed by the MV and Basic, which is backed by the non-MV portion of the Dissemination tables and only supports a limited set of parameters.
``` 

## PR checklist: submitters

## PR checklist: reviewers


backend/dissemination/search.py
```diff
@@ -30,12 +48,18 @@ def search(params):
...
     # GENERAL
-    results = search_general(params)
-    results = _sort_results(results, params)
-    results = search_alns(results, params)
-    results = search_findings(results, params)
-    results = search_direct_funding(results, params)
-    results = search_major_program(results, params)
+
+    if is_only_general_params(params):
+        results = search_general(DisseminationCombined, params)
```
timoballard, line 53: should this be

```python
results = search_general(General, params)
```

?

jadudm: I think, under `two`, we no longer need the branch between the two tables. And, in fact, I think the `if` can go away? @sambodeme and I saw this earlier, and wonder. I think this was a brief period of time where we thought we could improve things by keeping the general things on general... but, it doesn't matter for the use cases we're chasing?

jadudm: But... I'd be happy to pair and discuss. I'm not 100% certain. I *think* it wants to become `DisseminationCombined`, however. Because, otherwise, you're using a table that is "in the future" for some of your query, and a table that is "in the past" for other parts. 

Which is part of why we'll need to include a doc update (and the banner at the top of search) to say "search is updated once per hour," or similar. Someone coming to check if their audit was submitted won't know until the top of the hour (for example).

timoballard: yes that last point is what I had in mind when I asked -- not that querying `General` in this case would be better, but that it would always give up to date results. All things being equal, I'd prefer to not do that and have all of the search paths run against the same dataset, but wasn't sure how flexible that use case is, where people want to be extra sure that their submission really was accepted

jadudm: Unless you have a better idea, I think search has to live "one hour in the past."

We can, if we want, do several things:

1. Create a "is it submitted?" page. This is a UEI-only search against `general`. People can have an option to go to this page after they are done. (This could be a good solution, for all I know. While "single use," it addresses a point of stress that comes after submitting a big document of significance.)
2. We leave the site structure as-is, and document in multiple places that search updates once... per hour? Every 2 hours? (I don't know how long the refresh will take, or the load it will cause on the DB while rebuilding at this point.) We update the landing page where completed audits are shown to make clear "an audit listed here is considered complete and submitted to the FAC." 
3. ?

Either way, I agree search should run against one model/dataset only. I think we're currently in a world where we have to run against the `MATERIALIZED VIEW` and not the live models/dissemination tables.

jadudm: I'm essentially brainstorming in a GH thread. 

1. Users seem confused/unclear about their audit status on the internal landing page. We have struggled to make that text clear enough that they believe it represents a submitted audit.
2. I don't know if my #1 above is a use case. It's... a brainstorm in response to something you said, I think. Should we do it? I don't know. Possibly worth discussing, since it's a simple enough page... but it might not be worth the effort right now. Hence...

For the moment, I think we're in the category of #2: just use the MV. There's a delay for them to be able to verify their submission is in dissemination. :shrug:

danswick:
There is some inconsistency in the data deployed to preview: some items appear in search results (via the MV), but their corresponding summary views (Dissemination General) throw a 404. It looks like we aren't actually generating MV on deploy because these lines are commented out in the profile: https://github.com/GSA-TTS/FAC/pull/3511/files#diff-b716fce0dfec76b1c5b1c51c4abedfc2e21ce20343214b120dffe8fff2e20f32R26-R41. I believe these were uncommented because of slow deploys / RDS disk space limitations, which are possibly resolved now.

We should either trigger a manual MV regeneration or deploy to preview after uncommenting those lines.

backend/dissemination/management/commands/materialized_views.py
```diff
@@ -0,0 +1,26 @@
+from django.core.management.base import BaseCommand
+from dissemination import api_versions
+
+refresh_views = """
+REFRESH MATERIALIZED VIEW dissemination_combined;
```
danswick, line 5: I think we need to use `CONCURRENTLY` here or the refresh will rebuild the _entire_ MV, which blocks all interactions until it's done. https://www.postgresql.org/docs/15/sql-refreshmaterializedview.html

backend/run.sh
```diff
@@ -37,6 +38,17 @@ gonogo "migrate_app_tables"
...
+#####
+# MATERIALIZED VIEWS
+# These are unmanaged from Django's POV.
+# So, we don't have strictly have to drop them.
+# But, we might change them between deploys.
+# In which case, we should drop and recreate.
+drop_materialized_views
+gonogo "drop_materialized_views"
+create_materialized_views
+gonogo "create_materialized_views"
```
danswick, line 50: We might not want to drop-and-create on every deploy. The DB is completely locked while creating the MV, which means no one can interact while the (long-running) create command is running. We should consider building once, then running a **concurrent** refresh on deploy and on a schedule. There will still be some downtime during the refresh, but it _should_ be substantially shorter.

jadudm: I hadn't gotten that far, but having something in here that checks to see if it exists, and then *does nothing* on deploy, is possible.

I have an example, in `bash`, of this over in... the code that loads the historic data.

backend/run.sh
```diff
@@ -13,6 +13,7 @@ source tools/api_teardown.sh
 source tools/migrate_app_tables.sh
 source tools/api_standup.sh
 source tools/seed_cog_baseline.sh
+source tools/materialized_views.sh
```
danswick, line 16:
```suggestion
```

danswick (approved):
Reviewed in cowork and tested manually both locally and in preview.

Issue #3548, opened 2024-03-22
Create a new search page for General-only queries


danswick:
Follows discussion here: https://github.com/GSA-TTS/FAC/pull/3511/files#r1529245522. 

We should route General-only searches (those that don't require a join) to a simpler search interface users can use to, for example, verify their submission went through. A second, "advanced" search page would handle searches that require the Materialized Views table. 

```[tasklist]
### Tasks
- [x] Create basic, advanced search pages.
- [x] Update search documentation on static site.
```

PR #3552, opened 2024-03-25
Search, Advanced Search, Oh My!

jadudm:
In reference to https://github.com/GSA-TTS/FAC/issues/3548

Per the docs at the top of `search.html`:

```
To undo this action, complete the following steps. 
Search filters:
    1. Uncomment the search filters
Warnings:
    1. Remove search-alert-performance.html and its imports.
    2. Remove the warning banner in header.html
Summary report downloads:
    1. Uncomment the summary report download button here
    2. Uncomment the SF-SAC download button in summary.html
    3. Remove the forced Http404 in SingleSummaryReportDownloadView (dissemination/views.py)

See https://github.com/GSA-TTS/FAC/issues/3526 for more information.
```

1. The search filters in `search.html` were largely left commented out; the filters that apply to the `General` were uncommented, as they do not represent a performance issue for a "basic" search.
2. Alerts were *commented out*. 
3. Alert banner was commented out.
4. Summary report button was uncommented both in search results and on the summary page.
5. I do not know where the forced `404` is. 

To be done by people more knowledgeable than me:

1. I created a new URL pathway, `search/advanced`. This felt better than creating a new top-level URL.
2. I created a new view, `AdvancedSearch` and associated machinery. It is a copy of `Search`, with the requisite changes to route correctly.
3. I added code to bounce back-and-forth between the basic and advanced search from each page. It is in the info box.

I think the two pieces of code could be refactored nicely, so that the basic search only touches `General`, and does not accept any advanced search params. Basically, someone could get rid of my copy-pasta. My Django front-end skills are wanting here, and the goal is to get this live. So, this was the simplest, safest first step. 

To do:

* See if I can make sure that we hit `General` for the basic search, as opposed to `DisseminationCombined`.   

## PR checklist: submitters

## PR checklist: reviewers


backend/dissemination/search.py
```diff
@@ -16,22 +16,45 @@
...
-            "start_date",
-            "end_date",
-            "agency_name",
-            "cog_or_oversight",
-        ]
-    )
-    return params_set.issubset(gen_set)
+    only_general_params = True
+
+    if "advanced_search_flag" in params_dict:
```
danswick, line 40:
```suggestion
    if params_dict.get("advanced_search_flag", False):
```

tadhg-ohiggins (approved):

Issue #3485, opened 2024-03-06
Page Object Model for Cypress Test suite


eazapata23:
# Overview:
Based off of issue #3458 The idea was brought up to look into implementing Page Object Model(POM) starting with Search Page as a starting point.

# Objective:
Implement the Page Object Model(POM) for the Search Page to enhance code maintainability and reusability.

# Scope: 
Create page objects for web elements and actions related to the Search Page. After that is done, do the same for the rest of cypress test suite. 

# Acceptance Criteria:

1. Page Objects are successfully created for all relevant elements on the Search Page.
2. Implement POM for rest of Cypress test suite.
3. Methods for common actions are implemented and tested.
4. Code is reviewed and approved by a team member.

### Tasks

```[tasklist]
- [ ] https://github.com/GSA-TTS/FAC/pull/3574
- [x] Create Page Object class for the Search Page.
- [x] Implement methods for interacting with the search input field.
- [x] Implement methods for clicking the search button.
- [x] Implement methods for verifying search results
- [x] add summary report page + tests for downloading SF-SAC and Single Audit report
- [ ] https://github.com/GSA-TTS/FAC/issues/3590
- [ ] https://github.com/GSA-TTS/FAC/issues/3589
- [ ] https://github.com/GSA-TTS/FAC/issues/3591
- [ ] https://github.com/GSA-TTS/FAC/issues/3592
- [ ] https://github.com/GSA-TTS/FAC/issues/3593
- [ ] https://github.com/GSA-TTS/FAC/issues/3628
- [ ] https://github.com/GSA-TTS/FAC/issues/3629
- [ ] https://github.com/GSA-TTS/FAC/issues/3630
- [ ] https://github.com/GSA-TTS/FAC/issues/3631
- [ ] https://github.com/GSA-TTS/FAC/issues/3632
- [ ] https://github.com/GSA-TTS/FAC/issues/3633
- [ ] https://github.com/GSA-TTS/FAC/issues/3634
- [ ] https://github.com/GSA-TTS/FAC/issues/3636
- [ ] https://github.com/GSA-TTS/FAC/issues/3637
```

eazapata23:
[...]

PR #3574, opened 2024-03-28
DRAFT - WIP - POM for Search Page + Cypress tests

### LINKED ISSUE #3590: POM for gen info

### LINKED ISSUE #3591: POM for check access
linked to #3485

eazapata23:
This PR is linked to #3485 
This PR will close #3590 
This PR will close #3591 

In this PR:

- I added a plugin `downloadFile` for cypress and added it as a task in `cypress.config.js`
- Under `support` in Cypress directory, in the `commands.js` file the line `require(cypress-downloadfile)` was added.
- `Cypress-downloadfile` was added to `package-lock.json` & `package.json`
- Added `search-tests.cy.js` to e2e, which has two tests 1) basic search test & 2) advance search test
- Added data for search page to fixtures under `search_page_data` this holds all data for search parameters in one single json file vs before where I originally separated each field into it's own separate json files.
- I created a `page_objects` folder w/ file `search-page.js` this holds all web elements for search page + summary report and method/functions for interacting with web element.
- POM for `gen-info `added to `Page Objects` + deleted old `gen-info` cypress test.
- I modified the `login-gov.js` to include an `if else` statement due to the fact that login.gov started to ask if you wanted to add your PIV card as an authentication, I added the `if else` to skip this step.
- I modified the `initialize-audit.js` to import from the `page_objects` folder.
- I added `check-access.js` to the `page_objects` folder and deleted it from the `support` folder.
- I also added some new dependencies to both the `package-lock.json` & `package.json`.

# How to test:

1. Pull this branch, now since we're using a plugin for cypress, we need to make sure to download it locally on our machine or the tests won't work for search. Make sure you're on the `backend` directory of the project and run `npm i --save-dev cypress-downloadfile`. Once installed, you should see in your IDE a new folder titled `downloads` appear in the cypress directory, it should appear just under cypress but before e2e. 

2. Run docker as you normally would and once docker is running on another tab run this command `docker compose exec web /bin/bash` this will get you into docker shell. After that, run this command `python manage.py materialized_views --create` .  (I had to do this since I didn't have materialized view and my search tests would fail but after running command they passed)

3. Now in your terminal, enter your cypress env variables and run npx cypress open 

4. Once cypress is open, run the `full-submission.cy.js` tests should pass may take one or two tries for them but once they are completed run the other test for `search-tests.cy.js` -- Side 
Note: BEFORE running search tests, you'll notice that in your cypress downloads folder you won't have anything in there but AFTER running search tests, a `sf-sac-summary-report.xlsx` and/or a `single-audit-report.pdf` will show up, indicating that the download did work.
# Before Running Search Tests:
Screenshot 2024-04-04 at 12 09 16 PM

5. Now once you run `search-tests.cy.js` you should able to run a basic search that populates results, selecting a an audit from the table and viewing your summary report and downloading each document successfully.

https://github.com/GSA-TTS/FAC/assets/99770283/65d9f0c9-5872-4efa-b386-15805ab5e3d2

# After Running Search Tests:
Screenshot 2024-04-04 at 12 30 02 PM

6. The same applies for testing `advance search fields` but these are commented out for now since basic search returns results.

7. You can also verify search results with dbeaver by checking in `Materialized Views`:
Screenshot 2024-04-04 at 12 36 35 PM

## PR checklist: submitters

## PR checklist: reviewers


backend/cypress/pageObjects/general-info.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+    
+
+    selectAuditTypeAndPeriod() {
+        cy.get(this.auditTypeSingle).click();
+
+        cy.get(this.auditPeriodAnnual).click();
+    }
+
+    fillAuditeeInformation() {
```
timoballard, line 63: Wondering if there's a way to separate the page objects from the fixture data, such that the fixture data gets passed into these `fill` functions rather than loaded internally. Relatedly, is there a reason we can't have a single `json` file per page (e.g. `genInfo.json`, `search.json`, etc) rather than one file per field per page. That way, these functions could look more like:
```
fillAuditeeInformation(genInfo) {
  cy.get(this.auditeeNameField).type(genInfo.auditeeNameField);
  ...
}
```

thoughts?

purvinptl: 1. We can seperate the fixture data. I was following the search POM but it had its own search but gen-info is just being passed into other tests.
2. We can have a single json file. Again I was trying to make it consistent with Search POM fixtures.

I can start to implement those changes and request re-review.

eazapata23: We now have one file now for genInfoData and also separated page objects from the fixture data.

backend/cypress/e2e/search-tests.cy.js
```diff
@@ -0,0 +1,156 @@
...
+  reportId = searchData.reportId;
+
+
+  search = new searchPage();
+});
+
+//basic search test
+describe('Test Basic Search Fields', () => {
+
+  it('checks Audit Years', () => {
```
jperson1, line 47: We may want a more descriptive name for these tests, since they fill out every box

jperson1 (changes requested):
The tests look great! One thing, and two notes:

1. We should rename the `it` blocks, or break them into several `it`'s. 

1. Documentation on running advanced search locally is lacking. A quick blurb on what the materialized view is and how to get it running/how it relates to search would be good. It's not terribly pertinent to this PR, however.
2. The report ID is something that will change with every run of these tests, so we'll want a good way to make sure the correct one gets pulled for use when we run these in the higher environments. Maybe an API call to grab the most recent test submissions report_id, or something like that.

backend/cypress/e2e/search-tests.cy.js
eazapata23: That's a oversight on my end, you're right they do need to be more descriptive.

jperson1 (approved):
🎉

eazapata23:
Needs further revision based off questions/concerns not addressed here.

danswick (changes requested):
Per discussion with @eazapata23, this branch/PR is going to be restructured a bit before landing.

eazapata23:
This PR is now tracked and will close by PR #3574

PR #3607, opened 2024-04-04
Materialized Views Patch

asteel-gsa:
- Changes Wording
- Re-adds the `if:` block so that the action doesnt run over itself [like so](https://github.com/GSA-TTS/FAC/actions/runs/8550001910)


jadudm (approved):
LGTM.

Issue #3571, opened 2024-03-28
Handle accidental uploads of findings-associated workbooks 


danswick:
There are cases where a user:

1. Uploads an awards workbook with _some number of findings_ listed.
2. Uploads the other findings-associated workbooks.
3. Realizes they made a mistake and that there should be _no_ findings associated with the report.

Once in this state, it's impossible to remove the findings-associated data from an in-progress report, even by uploading empty workbooks in their place. We simply ignore the empty workbooks and retain the old data.

This is a bit of an edge case, but also creates a scenario where the app is behaving in an opaque way, potentially disseminated _incorrect_ information that looked perfectly correct to the submitter.

danswick:
This might be less of a user experience issue and more of a data integrity one. Since this may be resolved as part of the resubmisison work, we might not want to go down a _removal_ path, but instead block off the workaround path described above just to prevent incorrect data from making it into dissemination.

PR #3665, opened 2024-04-17
3571 handle accidental uploads of findings associated workbooks

sambodeme:
## Description
This PR allows users to delete a worksheet after they have uploaded it and realized it is not needed.
The change only applies to optional worksheets. When users upload an optional worksheet, a new link appears near the edit link to allow users to delete the worksheet.
Upon clicking on the worksheet removal link, a new page is displayed, warning the user about the deletion action and asking for confirmation before proceeding. If confirmed, the worksheet will be deleted and the user will be redirected to the submission progress page. In case of an error, the application stays on the deletion page and displays a user-friendly error message to indicate the issue.
See #3571 for context.

## To Test:
- Upload an optional worksheet. Note that the removal link appears.
- Click on the link and delete the worksheet.
- Confirm the deletion when prompted.
- Verify that you are redirected to the submission progress page.
- Ensure that the worksheet is correctly removed.

Noticed new links from the submission checklist page 
image

And from the upload page

image

## PR checklist: submitters

## PR checklist: reviewers


backend/report_submission/views.py
```diff
@@ -453,3 +479,119 @@ def post(self, request, *args, **kwargs):
...
+                return redirect(request.path)
+            path_name = request.path.split("/")[2]
+            section = self.additional_context[path_name]
+
+            try:
+                excel_files = ExcelFile.objects.filter(
+                    sac=sac, form_section=section["form_section"]
+                )
+                logger.info(f"Deleting {excel_files.count()} files.")
+                excel_files.delete()
```
timoballard, line 572: Deleting the `ExcelFile` record does not, it turns out, delete the actual file from S3. The data is correctly nulled out in the `SingleAuditChecklist` record, and the `ExcelFile` record is gone, but the S3 file remains (see minio screenshot after deleting the CAP upload):
image

sambodeme: Yes, this is what we discussed during dev huddle last week and my understanding was that we are not removing from s3 as part of this PR but we'll have another task/ticket to handle the deletion as part dissemination (and possibly clean up files for the already disseminated records).

timoballard: Right- couldn't 100% recall where we landed last week so I wanted to be sure - thanks for the clarification!

backend/report_submission/test_views.py
```diff
@@ -758,3 +763,105 @@ def test_post_validates_general_information(self):
...
+
+    def test_get_no_access(self):
+        request = self.make_request("get")
+        with patch("audit.models.SingleAuditChecklist.objects.get") as mock_get:
+            mock_get.side_effect = SingleAuditChecklist.DoesNotExist
+            with self.assertRaises(PermissionDenied):
+                self.view.get(request, report_id=self.report_id)
+
+    def test_get_successful_render(self):
+        sac = MagicMock(report_id=self.report_id)
```
timoballard, line 813: what are the tradeoffs here between using `MagicMock` and `Baker` (which we use in most of our test code)?

sambodeme: Baker could be used, but I did not find it necessary here as the focus is more on testing the behavior of the view rendering, so the database interaction was mocked. Using Baker would become relevant if it were necessary to test the database interaction with the Django test database.

Leighdiddy (commented):
Reviewing for UI only from screenshots - looks good to me!

lauraherring:
Looks good to me from a content perspective.

Issue #3704, opened 2024-04-19
Address historical audits with missing auditor EINs


gsa-suk:
This ticket addresses the issue related to invalid EINs. Some historical records are either missing the auditor EIN, or contain an auditor EIN value that will fail validation against the application's EIN schema with the error message 'does not match '^[0-9]{9}$''. The team-approved solution is to replace the missing or invalid value with 'GSA_MIGRATION', which is what this ticket will implement.
Changes will include:
- Update the schema to allow 'GSA_MIGRATION' in EIN fields.
- Update logic to only allow 'GSA_MIGRATION' in EIN fields when the migration flag is present.
- Error message: `'auditor_ein' is a required property`
- Example: 257745, 2022

sambodeme:
Close by #3726

PR #3726, opened 2024-04-19
3701 address errors caused by bad or missing EINs

sambodeme:
## Description
Some historical records are either missing the auditee's or auditor's EINs, or contain an auditee/auditor EIN value that will fail validation against the application's EIN schema with the error message 'does not match '^[0-9]{9}$''. The team-approved solution is to replace the missing or invalid value with 'GSA_MIGRATION', which is what this ticket will implement.
Changes include:
Updated schema to allow 'GSS_MIGRATION' in EIN fields.
Updated logic to only allow 'GSA_MIGRATION' in EIN fields when the migration flag is present.
#3701 
#3704 

## For Testing
- Find a report that migrates with no issue
- Delete the auditee or auditor EINs (or change the value to an invalid value) and re-run migration for that report.
- Observe that the migrated data contains `GSA_MIGRATION` in place of the missing/invalid EIN

## PR checklist: submitters

## PR checklist: reviewers


phildominguez-gsa (approved):

timoballard (approved):
Only concern was that this would technically allow users to enter "GSA_MIGRATION" as a valid EIN on the frontend general information form, but confirmed that [the form validation prevents that](https://github.com/GSA-TTS/FAC/blob/aba8a483b4e8aafa6d3cb1746581bde69933641d/backend/report_submission/forms.py#L75-L78)

Looks good!

sambodeme:
> Only concern was that this would technically allow users to enter "GSA_MIGRATION" as a valid EIN on the frontend general information form, but confirmed that [the form validation prevents that](https://github.com/GSA-TTS/FAC/blob/aba8a483b4e8aafa6d3cb1746581bde69933641d/backend/report_submission/forms.py#L75-L78)
> 
> Looks good!

Also, [this code](https://github.com/GSA-TTS/FAC/pull/3726/files#diff-ee11638a6683bf5050aa4a4cc679ae4ae767aa738864658283a06bda523299faR247) was meant to block `GSA_MIGRATION` in current submissions.

Issue #3725, opened 2024-04-19
Search field reset broken for basic search


danswick:
Per chat with @Leighdiddy, the "Reset Search" button doesn't work correctly on the basic search page. The underlying JS looks through a few hard-coded fields ([for example](https://github.com/GSA-TTS/FAC/blob/76987104dd20a994c19702cbff8f02c9db7aa118/backend/static/js/search-results.js#L65)), some of which no longer exist on the basic page.

PR #3744, opened 2024-04-23
Bug: Search Field Reset Broken for Basic Search

jperson1:
# Bug: Search Field Reset Broken for Basic Search

Issue:
Closes https://github.com/GSA-TTS/FAC/issues/3725

## Changes:
1. Ensure reset goes through if the `Cognizant or Oversight` fields aren't present (as in the case of basic search)
2. Break the cog/over and blanket checkbox resets into their own functions
3. Add JSDoc comments to most search related functions
    * These are the JS equivalent to a Python Docstring, which allows programs to access them nicely. (ex. VSCode tooltips, for many of us)

## How to test:
1. Switch to this branch and run normally. This should not require a fresh collectstatic, but it might.
2. Navigate to basic search. Ensure a fresh page load, so that the old JS isn't cached by Chrome/Firefox/Edge(?)
3. Fill out the filter fields. Hit "Reset Search", and verify that it does.
    * Audit year 2023 should be the only thing still selected/filled out
4. Do the same for advanced search. 

## Recording:
1. Basic search, bottom reset button
2. Advanced search, top reset button

https://github.com/GSA-TTS/FAC/assets/91098850/c5147eff-7614-4a90-b7a6-963443d9ffcd

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/static/js/search-results.js
+++ b/backend/static/js/search-results.js
@@ -8,10 +8,9 @@ const previous_page_link = document.querySelectorAll(
 const loader = document.getElementById(`loader`);
 const search_arrow = document.getElementById(`search_arrow`);
 
-/*
-  If any pagination links are clicked, set the page form element and submit it for a reload.
-  If the next or previous page buttons are clicked, set the page form element to be +/- 1 and submit for a reload.
-*/
+/**
+ * Attach event handlers for the pagination buttons. On click, set the appropriate form inputs and submit it for a reload.
+ */
 function attachEventHandlersPagination() {
   pagination_links.forEach((link) => {
     link.addEventListener('click', (e) => {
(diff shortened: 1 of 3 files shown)


phildominguez-gsa (approved):

backend/static/js/search-results.js
```diff
@@ -37,59 +36,88 @@ function attachEventHandlersPagination() {
...
-*/
+/**
+ * Reset the checkboxes. Clears audit year, entity type, etc.
+ * Re-check audit year 2023.
+ */
+function resetCheckboxes() {
+  var checkboxes = document.querySelectorAll(
+    '[type="checkbox"], [type="radio"]'
+  );
+  var audit_year_2023 = document.getElementById('audit-year-2023');
```
timoballard, line 82: nitpick: is it possible to refactor this so that it targets the second checkbox (skipping "all years") rather than explicitly targeting 2023? Might save our future selves an extra step/headache when rolling over to a new submission year

jperson1: For sure! I thought about doing this, and promptly forgot. Good call!

I looked at maybe sorting the list of audit years to grab the most recent audit year (the largest number), but that seems somewhat excessive here. So, I just recheck the second box.

timoballard (approved):
Looks good @jperson1 

Added [a ticket for covering "Reset Search" with a Cypress test](https://github.com/GSA-TTS/FAC/issues/3749) as a follow-on item

Issue #3702, opened 2024-04-19
Address historical audits with missing major programs


phildominguez-gsa:
- Error message: `Major program must be Y or N; cannot be left empty`
- Example: 191936, 2019
- Explanation: Some historic audits are missing values for the `MAJORPROGRAM` field in the electaudits table, which (correctly) fails our validations.
- Proposed solution: Set `MAJORPROGRAM` to “Y” if `TYPEREPORT_MP` is provided, and “N” otherwise.

PR #3784, opened 2024-05-01
Replace missing major program in historical audits

phildominguez-gsa:
Closes https://github.com/GSA-TTS/FAC/issues/3702

In this PR:
- Some historic audits are missing values for the `MAJORPROGRAM` field in the electaudits table, which (correctly) fails our validations.
- This fix sets `MAJORPROGRAM` to “Y” if `TYPEREPORT_MP` is provided, and “N” otherwise.
- Unit tests

Testing:
- "Y" case
  - Find (or create) an audit in the `elecaudits` table with a blank `MAJORPROGRAM` and has a `TYPEREPORT_MP` set.
  - The entry in the migrationinspectionrecord table should show "" for `MAJORPROGRAM` was replaced with "Y"
  - `is_major` in the dissemination_federalaward table should be "Y"
- "N" case 
  - Find (or create) an audit in the `elecaudits` table with a blank `MAJORPROGRAM` and does NOT have `TYPEREPORT_MP` set.
  - The entry in the migrationinspectionrecord table should show "" for `MAJORPROGRAM` was replaced with "N"
  - `is_major` in the dissemination_federalaward table should be "N"

## PR checklist: submitters

## PR checklist: reviewers


backend/census_historical_migration/test_federal_awards_xforms.py
```diff
@@ -588,6 +589,41 @@ def test_match_numbers_mixed_empty_and_non_empty(self):
...
+            self.MAJORPROGRAM = major_program
+            self.TYPEREPORT_MP = audit_type
+
+    def test_xform_normal_major_program(self):
+        """Test for normal major program"""
+        audits = [self.AuditMock("Y", "U")]
+
+        xform_missing_major_program(audits)
+
+        self.assertEqual(audits[0].MAJORPROGRAM, "Y")
```
gsa-suk, line 608: Shouldn't the result of the transformation be checked here?

backend/census_historical_migration/test_federal_awards_xforms.py
```diff
@@ -588,6 +589,41 @@ def test_match_numbers_mixed_empty_and_non_empty(self):
...
+
+    def test_xform_missing_major_program_with_audit_type(self):
+        """Test for missing major program with audit type provided"""
+        audits = [self.AuditMock("", "U")]
+
+        xform_missing_major_program(audits)
+
+        self.assertEqual(audits[0].MAJORPROGRAM, "Y")
```
gsa-suk, line 616: Shouldn't the result of the transformation be checked here?

backend/census_historical_migration/test_federal_awards_xforms.py
```diff
@@ -588,6 +589,41 @@ def test_match_numbers_mixed_empty_and_non_empty(self):
...
+
+    def test_xform_missing_major_program_without_audit_type(self):
+        """Test for missing major program without audit type provided"""
+        audits = [self.AuditMock("", "")]
+
+        xform_missing_major_program(audits)
+
+        self.assertEqual(audits[0].MAJORPROGRAM, "N")
```
gsa-suk, line 624: Shouldn't the result of the transformation be checked here?

backend/census_historical_migration/test_federal_awards_xforms.py
phildominguez-gsa: Some xforms modify the audit data directly, as `xform_missing_major_program` does

backend/census_historical_migration/test_federal_awards_xforms.py
phildominguez-gsa: See other comment

backend/census_historical_migration/test_federal_awards_xforms.py
phildominguez-gsa: See other comment

backend/census_historical_migration/test_federal_awards_xforms.py
timoballard: How do we decide which xforms return the modified data and which modify in place? Wondering if we might run into issues if/when we do stuff [like this](https://github.com/GSA-TTS/FAC/blob/967ddfa6b304304930ed0015842aa5f8ec76ca40/backend/audit/intakelib/transforms/runners.py#L56), which appears to assume that that there's data being returned from each function

backend/census_historical_migration/workbooklib/federal_awards.py
```diff
@@ -88,6 +88,32 @@
...
 
+def xform_missing_major_program(audits):
+    """Default missing major program by extrapolating from audit report type."""
+    change_records = []
+    is_empty_major_program_found = False
+
+    for audit in audits:
+        major_program = string_to_string(audit.MAJORPROGRAM)
+        if not major_program:
+            new_value = "Y" if audit.TYPEREPORT_MP else "N"
```
sambodeme, line 99:
```suggestion
            new_value = "Y" if string_to_string(audit.TYPEREPORT_MP) else "N"
```

phildominguez-gsa: Done

backend/census_historical_migration/test_federal_awards_xforms.py
phildominguez-gsa: A fair question, though the xforms we've been adding haven't been a part of those runner lists. A lot of the ones I was seeing altered the data instead of returning a value, so that's what I rolled with 🤷. Not opposed to standardizing our approach, though.

sambodeme (approved):

backend/census_historical_migration/workbooklib/federal_awards.py
```diff
@@ -88,6 +88,32 @@
...
+    """Default missing major program by extrapolating from audit report type."""
+    change_records = []
+    is_empty_major_program_found = False
+
+    for audit in audits:
+        major_program = string_to_string(audit.MAJORPROGRAM)
+        if not major_program:
+            new_value = "Y" if string_to_string(audit.TYPEREPORT_MP) else "N"
+
+            track_transformations(
```
sambodeme, line 101: We want to call track_transformations regardless of `if not major_program` condition. However, we want save the record (i.e., `InspectionRecord.append_federal_awards_changes(change_records)`) only when `is_empty_major_program_found` is true. @phildominguez-gsa

PR #3791, opened 2024-05-02
3718 address historical audits with missing planned actions

sambodeme:
## Description
Some historical reports have an empty `corrective action plan` text field. This field is mandatory and causes the migration to fail for those reports. This PR replaces empty `corrective action plan` text fields with `GSA_MIGRATION` keyword.

As of 05/07/2024, I updated the PR to include a similar fix for missing `text of finding`. See #3804 

## Test
- To test this PR, select an historical report and replace one of the `ELECCAPTEXT.TEXT` record with empty string.
- Try to migrate this report
- Confirm `GSA_MIGRATION` is in the corresponding `CapText.planned_action` field in dissemination. 
- Repeat the previous steps against `ELECFINDINGSTEXT.TEXT` to test the fix for missing `text of finding`

## PR checklist: submitters

## PR checklist: reviewers


phildominguez-gsa (dismissed):

backend/census_historical_migration/test_corrective_action_plan_xforms.py
```diff
@@ -59,3 +60,41 @@ def test_no_placeholder_addition_when_all_references_present(self):
...
+class TestXformAddPlaceholderForMissingActionPlanned(SimpleTestCase):
+    class CapText:
+        def __init__(
+            self, SEQ_NUMBER=None, FINDINGREFNUMS=None, TEXT=None, CHARTSTABLES=None
+        ):
+            self.SEQ_NUMBER = SEQ_NUMBER
+            self.FINDINGREFNUMS = FINDINGREFNUMS
+            self.TEXT = TEXT
+            self.CHARTSTABLES = CHARTSTABLES
+
```
gsa-suk, line 74: SEQ_NUMBER and CHARSTABLES are not used in the test cases.  Are these needed here?

sambodeme: They are not mandatory but rather provide information about the structure of this model. I can afford to do this here because the model does not have too many extra fields, as would have been the case for the ELECAUDITHEADER model.

gsa-suk: To keep things consistent with other test cases, may be the unwanted fields could be removed.

timoballard: I think it's ok to leave them, but maybe we could separate the fields that drive the test from the ones that are just there to give more context about the model -- maybe move the optional fields to the bottom with a comment clarifying that they're not strictly required for the test to avoid confusion in the future about what is/isn't necessary

timoballard (approved):
lgtm!

image

backend/census_historical_migration/test_corrective_action_plan_xforms.py
sambodeme: I would typically ignore extra fields in my mocks if there are too many of them and they are unused, as they could add confusion to the test cases. However, in this context, none of these concerns apply, as the test cases only address the fields used to avoid confusion. For instance, instead of something like self.CapText(FINDINGREFNUMS="123", TEXT="", SEQ_NUMBER="1", CHARTSTABLES="N"), the test cases simply did self.CapText(FINDINGREFNUMS="123", TEXT="").
That being said, I am open to keep or to remove the fields, as I only added them to inform readers of the real structure of the model, not to be used in the tests.

gsa-suk (approved):
Tested with (year='2020', dbkey='252699').  LGTM.

Issue #3716, opened 2024-04-19
Address historical audits with lowercase value for Y/N fields


phildominguez-gsa:
- Error message: `<b>y</b> is not accepted`
- Example: 234848, 2019
- Related validation: [has_invalid_yorn_field](https://github.com/GSA-TTS/FAC/blob/main/backend/audit/intakelib/checks/check_y_or_n__fields.py#L31)
- Explanation: Some historic audits have lowercase letters for various Y/N fields, which (correctly) fails our validations.
- Proposed solution: Convert these to uppercase Y/N. Transformation tracking is not required in this case.

PR #3797, opened 2024-05-03
Replace lowercase y/n with uppercase in historical audits

phildominguez-gsa:
Closes https://github.com/GSA-TTS/FAC/issues/3716

In this PR:
- Some historic audits have lowercase letters for various Y/N fields, which (correctly) fail our validations.
- This fix converts these to uppercase Y/N by replacing the default `str` transformation function in `SheetFieldMap`s with a new `uppercase_y_or_n` method.
- Transformation tracking is not required in this case.

Testing:
- Modify an audit by lowercasing all Y/N fields in the `eleccaptext`, `elecaudits`, `elecauditfindings`, and `electfindingstext` tables. There are a lot!
- The Y/N values in the `dissemination_finding`, `dissemination_captext`, `dissemination_findingtext`, and `dissemination_federalaward` tables should be re-capitalized
- If using preview data, dbkey 234848, audit year 2019 should pass

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/audit/intakelib/common/util.py
+++ b/backend/audit/intakelib/common/util.py
@@ -73,7 +73,7 @@ def is_yes_or_no_or_gsa_migration(v):
     # in historic data migration to pass validation.
     value = str(v).strip()
 
-    return value in ["Y", "y", "N", "n", settings.GSA_MIGRATION]
+    return value in ["Y", "N", settings.GSA_MIGRATION]
 
 
 def get_missing_value_errors(ir, range_name, message_key):
--- a/backend/audit/intakelib/common/util.py
+++ b/backend/audit/intakelib/common/util.py
@@ -73,7 +73,7 @@ def is_yes_or_no_or_gsa_migration(v):
     # in historic data migration to pass validation.
     value = str(v).strip()
 
-    return value in ["Y", "N", settings.GSA_MIGRATION]
+    return value in ["Y", "y", "N", "n", settings.GSA_MIGRATION]
 
 
 def get_missing_value_errors(ir, range_name, message_key):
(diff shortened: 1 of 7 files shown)


backend/audit/intakelib/common/util.py
```diff
@@ -73,7 +73,7 @@ def is_yes_or_no_or_gsa_migration(v):
     # in historic data migration to pass validation.
     value = str(v).strip()
 
-    return value in ["Y", "N", settings.GSA_MIGRATION]
+    return value in ["Y", "y", "N", "n", settings.GSA_MIGRATION]
```
sambodeme, line 76: We don't want to allow y/n here, as these are not expected at the schema level. In general, our Python checks are/should be less permissive than JSONSchema checks to prevent data from passing the former and failing the latter.

backend/census_historical_migration/transforms/xform_uppercase_y_or_n.py
```diff
@@ -0,0 +1,25 @@
+from ..exception_utils import DataMigrationValueError
+
+from django.conf import settings
+
+
+def uppercase_y_or_n(value):
```
sambodeme, line 6: `string_to_bool` will need a similar change to account for `y/n`

phildominguez-gsa: Isn't it being converted to Y/N before it gets to `string_to_bool`, though?

phildominguez-gsa: I thought that too, but it already performs an `upper()` before in that method

backend/audit/intakelib/common/util.py
phildominguez-gsa: Good call, done

gsa-suk (approved):
Tested with dbkey=234848, year=2019.  LGTM.

sambodeme (approved):

PR #3805, opened 2024-05-07
Replacing missing CPASTATE in historical audits

phildominguez-gsa:
Closes https://github.com/GSA-TTS/FAC/issues/3711

In this PR:
- Some historic audits are missing values for the `CPASTATE` field in the eleccpas table, which (correctly) fails our validations.
- This fix replaces the missing values with GSA_MIGRATION and updates the secondary auditors schema validations to allow this value
- Disallowing GSA_MIGRATION in uploaded workbooks by adding `secondary_auditor_address_state` to `check_gsa_migration_keyword.py`
- Creating `append_secondary_auditor_changes` for making inspection records
- Unit tests

Testing:
- Find an audit in the eleccpas table with a blank `CPASTATE`, or clear one that already exists
  - There are no modifiable entries when using local data, but you can create a row with these values, which are mostly borrowed from the one failing in preview: `7627,"243753201823516","2016","58817","","",FAC OVERRIDE,FAC OVERRIDE,FAC OVERRIDE,"","99999",FAC OVERRIDE,FAC OVERRIDE,"2133184987","",FAC OVERRIDE,"999999999"`
- The entry in the migrationinspectionrecord table should show it replaced "" with "GSA_MIGRATION"
- `address_state` in the dissemination_secondary_auditor table should be "GSA_MIGRATION"

## PR checklist: submitters

## PR checklist: reviewers


backend/census_historical_migration/test_secondary_auditors.py
```diff
@@ -0,0 +1,32 @@
+from django.conf import settings
```
gsa-suk, line 1: A different file name of 'test_secondary_auditor_xforms.py' might match the rest of the test files.

phildominguez-gsa: Done

gsa-suk (approved):
lgtm.

backend/census_historical_migration/workbooklib/secondary_auditors.py
```diff
@@ -59,6 +64,30 @@
...
 
 
+def xform_address_state(secondary_auditors):
+    """Default missing address_state to GSA_MIGRATION"""
+    change_records = []
+    is_empty_address_state_found = False
+
+    for secondary_auditor in secondary_auditors:
+        address_state = string_to_string(secondary_auditor.CPASTATE)
+        if not address_state:
```
sambodeme, line 74: We want to call `track_transformations` regardless of `if not address_state` condition. However, we want save the record (i.e., `InspectionRecord.append_secondary_auditor_changes(change_records)`) only if is_empty_address_state_found` is true.  @phildominguez-gsa

PR #3808, opened 2024-05-08
Replacing missing ZIPCODE, CPAZIPCODE in historical audits

phildominguez-gsa:
Closes https://github.com/GSA-TTS/FAC/issues/3711

In this PR:
- Some historic audits are missing values for the `ZIPCODE` and/or `CPAZIPCODE` fields in the elecauditheaders and eleccpas  tables, which (correctly) fail our validations.
- This fix replaces the missing values with GSA_MIGRATION and updates the general and secondary auditors schema validations to allow this value
- Disallowing GSA_MIGRATION in the General form by adding `auditee_zip` and `auditor_zip` to `validate_use_of_gsa_migration_keyword`
- Unit tests

Testing:
- Submitting an audit via app should pass cross validation
- elecauditheaders
  - Find an audit in the elecauditheaders table with a blank `ZIPCODE` and `CPAZIPCODE`, or clear one that already exists
    - The ones I saw in preview are missing more than zips, so while they will get past that error, they will probably fail on another field
  - The entry in the migrationinspectionrecord table should show it replaced "" with "GSA_MIGRATION"
  - `auditee_zip` and `auditor_zip` in the dissemination_general table should be "GSA_MIGRATION"
- eleccpas
  - Find an audit in the eleccpas table with a blank `CPAZIPCODE`, or clear one that already exists
    - There are no modifiable entries when using local data, but you can create a row with these values, which are mostly borrowed from the one failing in preview: `7628,"243753201823516","2016","145144","","",FAC OVERRIDE,FAC OVERRIDE,FAC OVERRIDE,CA,"",FAC OVERRIDE,FAC OVERRIDE,"2133184987","",FAC OVERRIDE,"999999999"`
    - The entry in the migrationinspectionrecord table should show it replaced "" with "GSA_MIGRATION"
    - `address_zipcode` in the dissemination_secondary_auditor table should be "GSA_MIGRATION"

## PR checklist: submitters

## PR checklist: reviewers


backend/audit/intakelib/checks/check_gsa_migration_keyword.py
```diff
@@ -32,6 +32,9 @@ def check_for_gsa_migration_keyword(ir):
         "contains_chart_or_table",
         "is_minimis_rate_used",
         "compliance_requirement",
+        "auditee_zip",
```
sambodeme, line 35: Here, we only track workbook fields (i.e., named ranges). `auditee_zip` and `auditor_zip` are part of `general_information` section.  This should go [here ](https://github.com/GSA-TTS/FAC/blob/main/backend/audit/validators.py#L247)

gsa-suk (approved):

sambodeme (approved):

Issue #3715, opened 2024-04-19
Address historical audits with missing firm/organization names


phildominguez-gsa:
- Error message: `Missing audit firm/organization name`
- Example: 246679, 2017
- Related section: [Secondary auditors](https://github.com/GSA-TTS/FAC/blob/main/backend/audit/intakelib/common/error_messages.py#L54)?
- Explanation:
  - Sudha's research - CPAFIRMNAME is blank in ELECCPAS table.
- Proposed solution: Use the GSA-MIGRATION keyword.

PR #3810, opened 2024-05-08
Sk/3715 missing firm names

gsa-suk:
**Closes**: https://github.com/GSA-TTS/FAC/issues/3715

**Testing**:

1. Run Test cases:

- Run fac test census_historical_migration.test_secondary_auditors_xforms.
- Result:
Screenshot 2024-05-08 at 3 32 36 PM

2. Test with a failed audit: (dbkey = 238797, audit_year = 2016) if access to census data in Preview is available.

- Run 'docker compose run --rm web python manage.py historic_data_migrator --years 'my_year' --dbkeys 'my_dbkey''
- Verify that auditor_name in dissemination_secondaryauditor is set to 'GSA_MIGRATION'.
- Verify that migration inspection record is updated in dissemination_migrationinspectionrecord table.

3. Test by changing CPAFIRMNAME in an audit:

- For an audit, change CPAFIRMNAME in census eleccpas table to ''
- Run 'docker compose run --rm web python manage.py historic_data_migrator --years 'my_year' --dbkeys 'my_dbkey''
- Verify that auditor_name in dissemination_secondaryauditor is set to 'GSA_MIGRATION'.
- Verify that migration inspection record is updated in dissemination_migrationinspectionrecord table.

## PR checklist: submitters

## PR checklist: reviewers


phildominguez-gsa:
I think you have to allow GSA_MIGRATION in [the schema](https://github.com/GSA-TTS/FAC/blob/main/backend/schemas/source/sections/SecondaryAuditors.schema.jsonnet#L26)

backend/census_historical_migration/workbooklib/secondary_auditors.py
```diff
@@ -94,6 +93,25 @@ def _get_secondary_auditors(dbkey, year):
     return sort_by_field(results, "ID")
 
 
+def xform_cpafirmname(secondary_auditors):
+    change_records = []
+    is_empty_cpafirmname_found = False
+    for secondary_auditor in secondary_auditors:
+        if secondary_auditor.CPAFIRMNAME == "":
+            is_empty_cpafirmname_found = True
+            track_transformations(
```
sambodeme, line 102: We want to call `track_transformations` regardless of `secondary_auditor.CPAFIRMNAME == ""` condition. However, we want save the record (i.e., `InspectionRecord.append_secondary_auditor_changes(change_records)`) only if `is_empty_cpafirmname_found` is true.

timoballard: could we clarify why this is the case as a comment in the code for future generations?

sambodeme: This is the convention we agreed upon before the first round of migration. For context, when a report section undergoes a transformation, the transformation function reviews each record in that section and modifies those that require transformation. At the end of the section, we need to determine which records were modified and which were not. Because we do not migrate any database primary keys along with the records, it becomes very challenging—and in some cases, nearly impossible—to accurately inform users which records have changed and which have not on the dissemination side. This is  because we lack a way to directly link each record in the `inspectionrecord` table to its corresponding record in the dissemination tables.
By saving all the records from the `inspectionrecord` for the given section, we can use the order of appearance in both the `inspectionrecord` and dissemination table records to establish the aforementioned link between records in `inspectionrecord` table and records in dissemination tables.

Also, We only save changes into the `inspectionrecord` table for a given section of a report if at least one record from that section has been modified.

phildominguez-gsa (approved):

sambodeme (approved):

PR #3854, opened 2024-05-17
Undo the workflow changes [Post Freeze]

asteel-gsa:
Undoes the changes to the workflow for after code freeze. 

As it turns out, we can just disable the workflows natively in git without making any changes to the workflow. Good to know for the future.

image


danswick (dismissed):
Do we need to remove `workflow_dispatch: null` from the [production workflow](https://github.com/GSA-TTS/FAC/blob/main/.github/workflows/deploy-production.yml#L7 )?

danswick (approved):
Looks great. Thanks, @asteel-gsa!

Issue #3881, opened 2024-05-24
[ADR] Access deletion


tadhg-ohiggins:
# Access deletion

## Areas of impact

- [ ]   Compliance
- [x]   Content
- [x]   CX
- [x]   Design
- [x]   Engineering
- [x]   Policy
- [x]   Product
- [ ]   Process
- [x]   UX

### Overview

There are occasional requests from users to remove other users from submissions—that is, to delete the `Access` instances that allow those other users to view and edit in-progress submissions.

This document only deals with in-progress submissions; submissions that have been disseminated are a separate category. The management of disseminated submissions on the intake side may need to be revisited when we deal with _**resubmission**_[^1].

### Current support

Note that while users have names recorded in the system, only user email addresses matter for purposes of authentication/access. Therefore all references to adding/removing/changing a user refer to adding/removing/changing an email address associated with a role.

Any reference to user management is only with respect to a specific submission; users cannot change the access of any users with respect to the application as a whole.

There are three roles:

*   **Audit Editor**
*   **Auditee Certifying Official**
*   **Auditor Certifying Official**

Users with these roles can all view the entire submission and make edits. Only users with the **Auditee Certifying Official** and **Auditor Certifying Official** roles can complete the certification step matching their role.

Submissions may have multiple users with the **Audit Editor** role, but only one each with the **Auditee Certifying Official** and **Auditor Certifying Official** roles.

In addition, submissions must have users for each of the **Auditee Certifying Official** and **Auditor Certifying Official** roles, and these cannot be the same user.

Only users with the **Audit Editor** role can modify access to the submission, and only these operations are possible:

*   Add a user with the **Audit Editor** role
*   Add the user for the **Auditee Certifying Official** role, if no such user exists
*   Add the user for the **Auditor Certifying Official** role, if no such user exists
*   Change the user for the **Auditee Certifying Official** role
*   Change the user for the **Auditor Certifying Official** role

In particular, the following three operations are not supported:

*   Changing or removing a user with the **Audit Editor** role
*   Removing a user with the **Auditee Certifying Official** role without supplying a new user to replace them
*   Removing a user with the **Auditor Certifying Official** role without supplying a new user to replace them

Note that the user who originally created the submission is automatically granted the **Audit Editor** role. So far, experience suggests that users who create submissions are users who tend to have responsibilities around managing those submissions; this is part of the rationale for restricting access management functions to users with that role.

### Removing Audit Editor access

[...]

PR #3897, opened 2024-05-29
Add access deletion ADR

tadhg-ohiggins:
Sometimes people leave  
Sometimes people get kicked out  
Either way: bye-bye.  

-----

Add Architecture Decision Record for adding ability to remove **Audit Editor** roles from submissions.

See https://github.com/GSA-TTS/FAC/issues/3881 for earlier discussion.

### DIFF
--- a/docs/architecture/decisions/0035-access-deletion.md
+++ b/docs/architecture/decisions/0035-access-deletion.md
@@ -84,12 +84,6 @@ We should support this use case by allowing users with the **Audit Editor** role
 
 For the moment, we will not support this use case, but will evaluate how much of a demand there is for removing Certifying Officials without requirement replacement users for their roles.
 
------
-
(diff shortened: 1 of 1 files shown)


jadudm:
Only one comment:

It looks like there are two "Story 2" entries? (I'm reading the merge, so perhaps I'm confusing myself.)

I would suggest removing story 3. It speaks to what we won't do, suggesting features we don't have and aren't discussing (adding users out-of-band). I think it can safely be removed, because the list of features we aren't discussing in this ADR is infinite. Story 1 and Story 2 capture what we're doing: providing the ability for people to remove Editors, and they can't remove themselves.

Unless I'm still missing why that story is critical for scoping the feature/decision.

tadhg-ohiggins:
@jadudm agreed, that was more of an exploratory segment and doesn't need to be in the ADR; removed.

jadudm (approved):
LGTM

PR #3904, opened 2024-05-29
3688 address historical federal program totals that are incorrect

sambodeme:
## Description

Some historical reports failed migration due to incorrect federal program total in the federal awards section. Our current validation rule does not allow incorrect federal program total in the federal awards workbook. The team has decided to migrate these reports as is by skipping that validation rule for those specific reports. This PR addresses that issue. The solution involves:

- Adding logic to look for incorrect federal program total in federal awards records before the creation of the federal awards workbook in the migration module.
- When a duplicate is detected, tracking the bad records and setting a global flag that will allow the report to bypass the corresponding validation rule once it reaches that logic.
- If the report does not break any other validation rule, saving the tracked bad records into the InvalidAuditRecord table and updating the ReportMigrationStatus table; otherwise, clearing the track.
- Clear the flag

## To test:

1. Launch the stack from the main branch.
2. Attempt to migrate the audit report with `dbkey = 182165` and `audit_year = 2022` and notice the migration failure.
3. Switch to this PR branch and re-run the migration for the same report and observe that it processes properly.
4. Check the InvalidAuditRecord table and ReportMigrationStatus table for records related to this audit.

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/audit/intakelib/checks/runners.py
+++ b/backend/audit/intakelib/checks/runners.py
@@ -1,6 +1,8 @@
 from django.core.exceptions import ValidationError
 import logging
 
+from census_historical_migration.invalid_record import InvalidRecord
+
 from .check_finding_award_references_pattern import award_references_pattern
 from .check_cluster_names import check_cluster_names
 from audit.fixtures.excel import FORM_SECTIONS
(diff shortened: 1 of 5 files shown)


backend/audit/intakelib/checks/runners.py
```diff
@@ -146,43 +148,90 @@
...
+    """Run the validation check if it is not skippable, or if it is skippable, only run if this is not a data migration."""
+    fun_name = get_key_by_value(skippable_checks, fun)
+    if fun == verify_auditee_uei_match:
+        return fun(ir, auditee_uei)
+    elif (
+        is_data_migration
+        and fun_name
+        and fun_name in InvalidRecord.fields["validations_to_skip"]
+    ):
+        return None
```
phildominguez-gsa, line 216: This should be the first case, IMO

sambodeme: The order would matter if they were related logic/conditions but in this case they are not.

phildominguez-gsa: Discussed offline (readability)

phildominguez-gsa (approved):

Issue #3933, opened 2024-06-05
Configure and Run ESLint on All JS Files


jperson1:
# Configure and Run ESLint on All JS Files

### Context:

During dependency update PR https://github.com/GSA-TTS/FAC/pull/3932, ESLint was migrated to its v9.0.0 configuration system. Here, it is noticed that ESLint:
1. Is run via `npx run check:eslint`, which in turn only covers the static/js library
2. Does not inherently ignore any files (which is possibly why we specify just the one directory)

This means ESLint does not cover any Cypress files, despite being configured to do so. It also doesn't cover standalone/separate JS files, such as the ESBuild dedicated file, `build-assets.js`. 

### Tasks

```[tasklist]
- [x] Have `npx run check:eslint` cover all* JS files (Cypress, `build-assets.js`)
- [x] Remove the `--ignore-pattern` argument from `check:eslint` to rely on the config file instead
- [x] Configure `eslint.config.mjs` to ignore the proper directories (compiled directories such as `staticfiles`, `node_modules`, and `static/compiled`)
- [x] Choose appropriate rules for the Cypress ESLint plugin (defined [here](https://github.com/cypress-io/eslint-plugin-cypress/tree/master/docs/rules)) - maybe, the default are fine
- [x] _Resolve any issues discovered by ESLint, now that it covers everything_
```

A casual enabling of ESLint for the Cypress directory reveals about 100 errors:

```
jperson@FM-C02DX829ML7J backend % npx eslint cypress/

/Users/jperson/Documents/GSA/FAC/backend/cypress/e2e/audit-info-form.cy.js
  20:5   error  It is unsafe to chain further commands that rely on the subject after this command. It is best to split the chain, chaining again from `cy.` in a next command line  cypress/unsafe-to-chain-command

...

/Users/jperson/Documents/GSA/FAC/backend/cypress/support/workbook-uploads.js
  16:38  error  'request' is defined but never used  no-unused-vars

✖ 100 problems (100 errors, 0 warnings)
```

PR #3965, opened 2024-06-10
ESLint Reconfiguration and Related Fixes

jperson1:
# ESLint Reconfiguration and Related Fixes

Issue: https://github.com/GSA-TTS/FAC/issues/3933

## Changes:
1. Have a standard `npx eslint` cover all the files we need covered
    * Update `eslint.config.mjs` to handle standard options, such as ignored directories
    * Update the package scripts and Makefile to use the standard execution
    * Now covers Cypress, it's own config file, etc.
2. Migrate a number of files to ES modules format
    * ES modules have been supported by node since 2017, so we're good there
    * Cypress config and plugin index files are updates, as well as the ESBuild specific `build-assets.mjs`
    * _The future is now_
3. Fix all the ESLint errors. 
    * Most are a Cypress error akin to `It is unsafe to chain further commands that rely on the subject after this command.` This means we shouldn't clear a text box, type into it, and blur it in rapid succession. These just get broken into several statements where re recapture the text box each time to verify that it's okay. 
    * A bunch of ES module changes
    * That's pretty much it? We already used prettier for just about everything, so there's more actual errors/warnings than formatting errors. 

Also, I included the ESLint Babel parser (@babel/eslint-parser). ESLint only officially supports the latest final ECMAScript standard, so Babel helps keep everything non-ECMA in check.

## How to test:
Locally:
2. Bring up a fresh stack, including refreshing the `node_modules` directory and running a `collectstatic`
3. Run `npx eslint` to verify the new config works.
5. Run the Cypress tests, to verify that the not-insignificant changes didn't change the behavior.

Other things:
1. The GitHub Actions PR tests should still pass.
2. Cypress through GitHub Actions should also still pass.

## PR checklist: submitters

## PR checklist: reviewers


jperson1:
This Cypress run shows that these changes didn't break the Cypress config: https://github.com/GSA-TTS/FAC/actions/runs/9455963528

They're failing, but it's likely because they were run against dev, which is likely not equipped to handle the authorization pieces. The important part is that they load up and start the tests with no issue.

backend/cypress.config.mjs
```diff
@@ -1,13 +1,14 @@
-const { defineConfig } = require('cypress');
-const { downloadFile } = require('cypress-downloadfile/lib/addPlugin');
+import { defineConfig } from 'cypress';
+import { downloadFile } from 'cypress-downloadfile/lib/addPlugin.js';
+import indexTasks from './cypress/plugins/index.mjs';
 
-module.exports = defineConfig({
+export default defineConfig({
```
danswick, line 5: 🙏

backend/cypress/e2e/check-eligibility.cy.js
```diff
@@ -78,7 +78,7 @@ describe('Create New Audit', () => {
     });
 
     it('should navigate to next step after successful submission', () => {
-      testValidEligbility();
+      testValidEligibility();
```
danswick, line 81: Nice catch. This one was hard to spot, even in the diff.

backend/cypress/e2e/check-ueid.cy.js
```diff
@@ -39,22 +40,28 @@ describe('Create New Audit', () => {
...
-        cy.get('#auditee_uei').type('ASDF').blur();
+        cy.get('#auditee_uei').type('ASDF');
+        cy.get('#auditee_uei').blur();
         cy.get('#auditee_uei-length').should('be.visible');
       });
 
       it('should indicate when the supplied input is too long', () => {
-        cy.get('#auditee_uei').clear().type('ASDFASDFASDFA').blur();
+        cy.get('#auditee_uei').clear();
+        cy.get('#auditee_uei').type('THIRTEENLONG0');
```
danswick, line 56: :heart:

backend/package.json
```diff
@@ -9,10 +9,10 @@
...
   "scripts": {
     "build": "run-p build:*",
-    "build:esbuild": "node build-assets.js",
-    "dev": "node build-assets.js --watch",
-    "check:eslint": "npx eslint --ignore-pattern ../.gitignore static/js",
-    "fix:eslint": "npx eslint --fix --ignore-pattern ../.gitignore static/js",
+    "build:esbuild": "node build-assets.mjs",
+    "dev": "node build-assets.mjs --watch",
+    "check:eslint": "npx eslint",
+    "fix:eslint": "npx eslint --fix",
```
danswick, line 15: I'm not sure why they were like this, but we don't need `npx` to run these commands as npm scripts. 

```suggestion
    "check:eslint": "eslint",
    "fix:eslint": "eslint --fix",
```

danswick (approved):
This looks great. Thanks, @jperson1! I left one small suggestion inline. Feel free to incorporate if you agree or merge if you don't.

backend/package.json
jperson1: Oh, fantastic! Went ahead and axed the `npx`'s for the other scripts as well.

danswick (approved):
Looks good to go!

Issue #3977, opened 2024-06-12
[ADR] Validation Waivers


tadhg-ohiggins:
# Validation waiver process

## Areas of impact

- [ ]   Compliance
- [x]   Content
- [x]   CX
- [x]   Design
- [x]   Engineering
- [x]   Policy
- [x]   Product
- [ ]   Process
- [x]   UX

## Related documents/links

Earlier discussions in [this doc](https://docs.google.com/document/d/1YzlmnzOQJM-uKjin4k90m8E01avqzKuA9YynTRh_UnQ/edit) and [this doc](https://docs.google.com/document/d/16Oo-no6cOQvtgjzVVJp-KUJIrEwAX8j1XX7HJj3gZu4/edit).

### Overview

The current FAC system has hard requirements for submissions. There are specific instances where these requirements need to be waived, but we currently lack support for that. This document proposes an approach.

#### Waivers

We introduce the notion that submissions have a set of waivers, one waiver per requirement waived[^1]. The typical submission has zero waivers.

Waivers require staff intervention, probably initiated via the help desk and performed using the Django Admin interface.

[^1]: In unusual cases there may be multiple waiver entries per submission per requirement waived; our code should be able to handle this.

#### Transparency

Any waivers will be public and will be published along with the rest of the FAC data for a disseminated audit. This ADR does not deal with the concept of waiver justifications that should not be public.

##### Example 1

An auditee certifying official may not be available to certify, and there may be no prospect of anyone being able to certify a given submission at any point. In such cases, the Census FAC practice was to stand in as the certifying official for the purposes of completing the submission. GSA FAC would take a similar approach: the submission would be able to proceed, and the summary report would note that GSA FAC issued a waiver for that submission and that no certifying auditee associated with the submission certified it.

##### Example 2

Circumstances may make it impossible to reactivate the UEI for an auditee, preventing the start of a submission. GSA FAC in this case would waive the requirement for an active UEI (note that a _valid_ UEI would still be required) and again note in the summary report that a waiver was issued and that the UEI was not active at the point the submission started.

## Decision

We will add a process for handling waiver requests and add application features to support this.

### Process

1.  Via help desk or other channels, FAC staff learn about the possible need for a validation waiver.
2.  A FAC staff member is assigned to handle this particular case.
3.  The FAC staff member either denies the request (in which case this is the last step) or confirms that a validation waiver is appropriate.
[...]

PR #3981, opened 2024-06-12
Add ADR for validation waivers

tadhg-ohiggins:
Your philosophy  
Does not contain all the things;  
Exceptions needed.  

-----

ADR for validation waivers.

Primary difference from [the ADR issue](https://github.com/GSA-TTS/FAC/issues/3977): rather than leaving ambiguous who the official “requester” might be, this version insists that the authorizing NSAC/KSAML is the requester for our purposes. The alternative here would be to have fields for both the original correspondent and the NSAC/KSAML, and to either leave the original correspondent blank or duplicate it in the case where that individual is the NSAC/KSAML.

### DIFF
--- a/docs/architecture/decisions/0037-validation-waivers.md
+++ b/docs/architecture/decisions/0037-validation-waivers.md
@@ -1,0 +1,145 @@
+# 37. Validation waivers
+
+Date: 2024-06-12
+
+## Status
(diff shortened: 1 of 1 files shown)


jadudm (approved):
LGTM.

PR #3995, opened 2024-06-17
Remove `data_distro`

tadhg-ohiggins:
Some experiments  
Lead to other approaches  
And are then cut down.  

-----

Remove `data_distro` directory, and references to it in the documentation. This isn’t being used anywhere, so shouldn’t break anything. However, @jadudm  , you may want to look at the changes to `development.md` to see if there’s now a different way to load data or if we should actually keep some of the `data_distro` code for that purpose.


backend/config/settings.py
```diff
@@ -126,7 +126,6 @@
     "api",
     "users",
     "report_submission",
-    # "data_distro",
```
timoballard, line 129: that this was already commented out makes me feel even less concerned

jadudm:
No data is loaded via `distro`. It happens via a completely different mechanism at deploy time. This should be fine.

timoballard (approved):

asteel-gsa:
This branch will need to be sync'd with main to ensure it has the latest terraform changes

phildominguez-gsa (approved):

PR #4051, opened 2024-07-02
Terraform provider update

asteel-gsa:
- Adds `delete_recursive_allowed = false` to to ensure the default `true` value does not get applied
```terraform
resource "cloudfoundry_space" "space" {
  name                     = var.name
  org                      = data.cloudfoundry_org.org.id
  allow_ssh                = var.allow_ssh
  delete_recursive_allowed = false
}
```
- Brings all module versions to `v1.0.0`, bringing us up to date with https://github.com/GSA-TTS/terraform-cloudgov
- Brings terraform provider to `~>0.53.1` [for compatability with `v1.0.0` above](https://github.com/GSA-TTS/terraform-cloudgov/blob/main/cg_space/providers.tf)
```terraform
  required_providers {
    cloudfoundry = {
      source  = "cloudfoundry-community/cloudfoundry"
      version = "~>0.53.1"
    }
  }
```
- removes `recursive_delete = true` from modules as it has been depreciated


asteel-gsa:
So... basically what is being done here, and im not entirely sure this is the _correct_ way of doing this, but due to me being unable to bootstrap things locally, I had to rely on grabbing specific things from cgov..

- created a space deployer account
- grabbed the production s3 state credentials
- forced the version upgrade to `~>0.53.1` in `providers-managed.tf` files
- ran `terraform init -upgrade` in `preview, staging, production, dev, meta`
- forced the lock files to be upgraded
- version bumped all modules we import from https://github.com/GSA-TTS/terraform-cloudgov to `v1.0.0` from `v0.9.1`
- removed `recursive_delete` as its no longer used in the modules
- applied to preview -> _no configuration changes found_
- created pr
- dev plan -> _no configuration changes found_
- meta plan -> _recognized the new template version of `~>0.53.1` and is attempting to overwrite the resources with it, which is okay, because they are checked in the same version for the providers. 

I am a little hesistant on this and would like to get mogul's opinion on how we should do this, if there is a better way. We are unable to upgrade from `v0.9.1` to `v.1.0.0` otherwise, due to our locks being `~>0.51.3`, where this was accessible in `v0.9.1` release, however, no longer available in `v1.0.0` release.

asteel-gsa:
This appears to be operational, however, we should proceed with caution and discuss if we want to fully update everything at some point

asteel-gsa:
@mogul if you by chance have a moment, I would greatly appreciate your 👀 on this, as I would like to ensure the terraform upgrade was done in a safe and secure way, to avoid any possible system breakage.

asteel-gsa:
OBE

PR #4052, opened 2024-07-02
Add option to only detected changed files

asteel-gsa:
Dan and I agree, the coverage report is good, but its bloated, hard to read, and often ignored.

Now the only coverage added to the PR is if the code is changed/added, and used with Django/Coverage. Should make things a lot easier to read.


jadudm (approved):
LGTM. Trims down the coverage report.

PR #4061, opened 2024-07-08
Change Coverage Reporting Tool

asteel-gsa:
The other major coverage tools output too much to the PR, and make it difficult to use for Review.
[cobertura-report](https://github.com/marketplace/actions/cobertura-report) has an outstanding issue with displaying modified files, as well as broke when the merge occurs due to this. It did not break when it output the entire coverage report to the PR
[code-coverage-report-action](https://github.com/marketplace/actions/code-coverage-report-action) while nice to have a diff on old coverage reports, outputs everything as if it were on the docker container and has no ability to hide 100% items.

The new action, while it doesnt display line numbers, gives the reviewer easy access to bring the testing up to 90% threshold, while keeping the pr clean.

### DIFF
--- a/.github/workflows/testing-from-build.yml
+++ b/.github/workflows/testing-from-build.yml
@@ -52,13 +52,13 @@ jobs:
         with:
           filename: ./coverage.xml
           badge: true
-          fail_below_min: false
+          fail_below_min: true
           format: markdown
           hide_branch_rate: false
           hide_complexity: true
           indicators: true
           output: both
-          thresholds: '85'
+          thresholds: '85 90'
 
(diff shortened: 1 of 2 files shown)


asteel-gsa:
@danswick What do you think of these thresholds for the new coverage report?

asteel-gsa:
Still unsure if the team is feeling the coverage necessity at this time. Thread on options [here](https://gsa-tts.slack.com/archives/C03QCP4FSN8/p1721403664051459)

jadudm (approved):
LGTM. And, if we want to change it/go back, we can.

Issue #3869, opened 2024-05-21
Add cross-validation for `total_amount_expended` < limit


tadhg-ohiggins:
Following from #3769, we may also need to catch the case where `total_award_expended` is some invalid amount.

We've primarily seen this in cases where there were no federal awards and `total_award_expended` is zero, but should we just reject if the total is below some limit, and should that limit be the same as the limit we ask submitters about in the initial steps? (Currently this is $750,000.)

jadudm:
I believe this is safe.

There is ambiguous policy floating around the world that suggests that, for some entities, they can/must submit at a lower threshold. This has never been resolved. Therefore, we will disregard it as an "unknown/unknown."

I think it is safe to set this to the threshold. If someone thinks they *should* be able to submit with a spend below the SA threshold, and we prevent them... they will reach out to us.

jperson1:
Moving this to "Blocked" for now. See the relevant Slack thread for why, but TL;DR:

Expenditures and reimbursements count against each other. If that brings the record below the threshold, we don't want to block it. How can we detect that, and is it our place to block a submission when the user has attested early in the process that they are above the threshold?

PR #4128, opened 2024-07-29
Add Cross Validation for total_amount_expended < DOLLAR_THRESHOLD

jperson1:
# Add Cross Validation for total_amount_expended < DOLLAR_THRESHOLD

Issue: closes https://github.com/GSA-TTS/FAC/issues/3869

## Changes:

Add `check_total_amount_expended` to the cross validation functions.

1. The actual function is in its own file
2. The error message references the UG and is included in `errors.py`
3. It is included in __init__ to ensure it runs with the rest
4. A test file covers a few cases for `total_amount_expended` - at, above, or below the threshold

## How to test:

1. Switch to this branch and run normally
2. Go through the submission process with a federal awards workbook with a too-low `total_amount_expended`
    * I edited the test workbook that Cypress uses to have $200 of awards total. That way the other workbooks are still usable.
3. Walk it through to cross validation
4. Ensure that the correct error message displays at the time of cross validation

## Screenshots:

The error message:
image

## PR checklist: submitters

## PR checklist: reviewers

### DIFF
--- a/backend/audit/cross_validation/__init__.py
+++ b/backend/audit/cross_validation/__init__.py
@@ -68,6 +68,7 @@ from .check_findings_count_consistency import check_findings_count_consistency
 from .check_has_federal_awards import check_has_federal_awards
 from .check_ref_number_in_cap import check_ref_number_in_cap
 from .check_ref_number_in_findings_text import check_ref_number_in_findings_text
+from .check_total_amount_expended import check_total_amount_expended
 from .sac_validation_shape import sac_validation_shape  # noqa: F401
 from .submission_progress_check import submission_progress_check
 from .tribal_data_sharing_consent import tribal_data_sharing_consent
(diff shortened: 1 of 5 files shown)


phildominguez-gsa (approved):

sambodeme (commented):
I suggest moving this to `intakelib/checks` instead. The reasoning behind cross-validation is that these validations perform checks involving more than a single section. In this case, all the information needed to perform the validation is available at intake time and only requires the federal award workbook, so it is more appropriate to perform the validation right when the workbook is being submitted. We also have a few checks in cross-validation, including `check_award_ref_existence`, `check_award_reference_uniqueness`, and `check_finding_reference_uniqueness`, that we will want to move into `intakelib/checks` due to the nature of their checks, but that is beyond the scope of this PR.

jperson1:
Closing this as superseded by 4198. The code can/will remain for reference, but it makes more sense to implement this when we have multiple thresholds to validate for.

Issue #3506, opened 2024-03-12
Fix Fiscal Period Start


sambodeme:
### Description:
The function [`xform_auditee_fiscal_period_start`](https://github.com/GSA-TTS/FAC/blob/main/backend/census_historical_migration/sac_general_lib/general_information.py#L232) aims to calculate the start date of an auditee's fiscal period based on the provided end date within the `general_information` dictionary. However, the current implementation assumes a fixed 365-day fiscal year, which does not account for leap years. This can lead to incorrect fiscal period start dates, especially around leap years.

#### Problematic Code:
```python
fiscal_start_date = xform_census_date_to_datetime(
    general_information.get("auditee_fiscal_period_end")
) - timedelta(days=365)
```

#### Suggested Fix:
Modify the function to check if the year of the fiscal period end date is a leap year and adjust the calculation accordingly. Here's a proposed adjustment:

```python
# Determine if the fiscal_end_date's year is a leap year
if isleap(fiscal_end_date.year):
    days_to_subtract = 366
else:
    days_to_subtract = 365

# Adjusted calculation for the fiscal start date
fiscal_start_date = fiscal_end_date - timedelta(days=days_to_subtract) + timedelta(days=1)
```

danswick:
We'll tackle this as part of the next batch of curation work.

PR #4239, opened 2024-08-28
Consider leap years when generating fiscal period start date

rnovak338:
Addresses #3506 

Previously, the fiscal period start date for audits generated from end dates in `/census_historical_migrations` did not account for leap years. The following PR changes take the fiscal period end date, and properly determines how many days there are in the fiscal period, based on the year.

@danswick and @sambodeme if this needs to be corrected please let me know - adjusting this should be trivial. Currently, the new logic handles the below end dates in the following way:

- If the end date is `10/1/2020`, the start date will be `10/1/2019` (366 days)
- If the end date is `10/1/2021`, the start date will be `10/1/2020` (365 days)
- If the end date is `2/1/2020`, the start date will be `2/1/2019` (365 days)
- If the end date is `2/1/2021`, the start date will be `2/1/2020` (366 days)

If the math is correct, I'll move this out of draft.

## PR Checklist: Submitter

## PR Checklist: Reviewer


backend/audit/utils.py
```diff
@@ -72,6 +73,22 @@ def remove_extra_fields(general_information_data):
             general_information_data.pop("audit_period_other_months", None)
         return general_information_data
 
+    @staticmethod
+    def check_leap_year(end_date):
```
jadudm, line 77: *Dates. I hate dates.* (With apologies to Indiana Jones.)

Is there a way we can use the date library to 

1. Do the calculation and
2. Put a test or two in place that asserts the things that you're doing in your code?

My concern is this: 

*Are there other cases you haven't considered?*

In this regard, if there's a way we can say

```
if calendar.isleap(...):
  return calendar.subtract_one_year(...)
```

so that the question of 365 vs. 366 is handled by someone else.

It would also be good for us to identify examples where this hit us, and make sure we have some tests that make certain that, when we use this function in the future, it really does the right thing on the edges we're worried about.

If that doesn't make sense, feel free to ask questions.

rnovak338: @sambodeme has found one of the tickets this morning related to this issue [here](https://fac-gov.zendesk.com/agent/tickets/4337). However, it is not fully clear to me that the audits involved are an issue _because of_ the leap year. If there are other helpdesk tickets related to this, getting some sample dates and expected outcomes would be very useful.

I did find a library **dateutil.relativedelta** which does seem to take into account the current year like you asked. From what I've read about it that means leap years too. That allows us to do something akin to:
```
start_date = end_date - relativedelta(years=1)
```

Whereas what we had before was `timedelta(days=365)`. But I would rather not continue to tweak on this until there are clear start/end dates from the helpdesk that I can experiment these changes on.

jadudm: I think pausing tweaking is reasonable; good call. We have a breadcrumb trail on your thinking here, which is good.

rnovak338: I think we are good to proceed. Speaking with Hassan and Dan, the new changes I am posting to this PR resolves the original issue.

https://app.fac.gov/dissemination/summary/2022-06-CENSUS-0000015322
The expected fiscal range is `7/1/2021` -> `6/30/2022`. I have adjusted one of the start dates from the `census_historical_migration` tests to be 6/30/2022. I am still unsure about the leap year piece, however, after running the above dates (from my PR description) against the new `dateutil` library, I am getting the output I expect.

rnovak338:
This is now ready for review after 10 lint commits 🥲

backend/census_historical_migration/sac_general_lib/general_information.py
```diff
@@ -310,14 +311,15 @@ def xform_auditee_fiscal_period_end(general_information):
 
 def xform_auditee_fiscal_period_start(general_information):
     """Constructs the fiscal period start from the fiscal period end"""
-    # Transformation to be documented.
-    fiscal_start_date = xform_census_date_to_datetime(
+    fiscal_end_date = xform_census_date_to_datetime(
```
sambodeme, line 314: I suggest adding a comment to this method to explain what the code was doing before this change and to highlight when and why this fix was implemented. Specifically, this fix was applied after the migration, so it's unlikely that the code will be run again for migration purposes. However, it was decided that everything related to the migration, including this transformation code change, should be treated as public records. Therefore, it will be helpful to indicate to reader that all migrations were run with the old version of the code.

sambodeme (approved):

PR #4291, opened 2024-09-17
Creating delete_blank_access_entries.py

phildominguez-gsa:
In this PR:
- Creating a command to delete Access entries that have blank email addresses. These are invalid, and we now enforce this in the app.
- The method [audit.models.delete_access_and_create_record](https://github.com/GSA-TTS/FAC/blob/180203905f8b1e34d6fef8f96bb5ddeee6cc5779/backend/audit/models/access.py#L136) was introduced some time ago, but not actually used as far as I can tell. [Tadhg's plan](https://github.com/GSA-TTS/FAC/blob/main/docs/architecture/decisions/0032-track-deletion-with-other-models.md) was to use this to create a DeletedAccess entry for each deleted Access, for record-keeping purposes, so I'm making use of that here.

Testing:
- Create a couple submissions (I just run the Cypress test)
- Go to the user access management pages for the submissions and add some users. You'll have to toss in some fake emails for now.
- In DBeaver, go to `audit_access` and delete the emails for the users you added
- Run `docker compose run web python manage.py delete_blank_access_entries --count`. It should log how many invalid Access entries there are, and exit.
- Run `docker compose run web python manage.py delete_blank_access_entries --limit 1`. It should only delete 1 Access entry, and add it to DeletedAccess (can confirm in DBeaver).
- Run `docker compose run web python manage.py delete_blank_access_entries` to delete the rest.

## PR Checklist: Submitter

## PR Checklist: Reviewer

### DIFF
--- a/backend/audit/management/commands/delete_blank_access_entries.py
+++ b/backend/audit/management/commands/delete_blank_access_entries.py
@@ -1,0 +1,71 @@
+from audit.models import (
+    Access,
+    DeletedAccess,
+    delete_access_and_create_record,
+)
+
+from django.core.management.base import BaseCommand
+
+import logging
+import sys
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Django management command for deleting Access entries with blank email fields.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--count", action="store_true", help="Only logs the blank Accesses count without deleting",
+        )
+        parser.add_argument(
+            "--verbose", action="store_true", help="Prints the report_ids that the Accesses are being deleted from",
+        )
+        parser.add_argument(
+            "--limit", type=int, help="Limits the number of blank Accesses to delete", default=None,
+        )
+
+    def handle(self, *args, **options):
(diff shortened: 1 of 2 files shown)


rnovak338:
FWIW (while awaiting Matt's approval), I was able to run through each of your commands successfully, using 4 sample audit accesses with no email. Nice work!

jadudm (changes requested):
I think this looks good. I'm working on the assumption that we'll test this in lower environments (dev/staging) before we use it in production. And, given that we have access controls in production, that will be easy enough to achieve.

I'm going to suggest one change; it's easy.

I think we shouldn't have a --verbose flag. I think the command should just be noisy. Why? Because it will send the logs of this work to NR. And, I think (given what it is doing), we absolutely want those log entries to go to NR. So, even though it will be a lot of rows (e.g. ~13K), we should definitely have the record of those DB manipulations in our production NR logs.

Does that make sense? In short, lets get rid of the `if` statements and `--verbose` flag, and just do the logging in all cases. (Happy to consider other possibilities...)

jadudm (approved):
LGTM, and thank you!

Issue #4259, opened 2024-09-06
Migrate from django-fsm to viewflow


danswick:
As of April this year, `django-fsm`, the state machine library we use to to manage submission events and the like, [has been deprecated](https://github.com/viewflow/django-fsm?tab=readme-ov-file#django-friendly-finite-state-machine-support). Its replacement, [viewflow](https://github.com/viewflow/viewflow/wiki/django%E2%80%90fsm-to-viewflow.fsm-Migration-Guide), is a backwards-incompatible re-write of the old system. While it seems to be a straightforward replacement, it will require some work to migrate.

**Update**:
As @rnovak338 pointed out, we import `django-fsm` in some old migrations, which means we can't easily remove the dependency without also doing some minor migration surgery. With that in mind, we'll probably do something like:

```[tasklist]
### Steps
- [x] Refactor the app code to use the viewflow API, add a new migration that changes the `FSMfield` to a standard `CharField`
- [x] Squash the existing migrations to get rid of the new-deprecated fsm reference. Or possibly some other approach to fixing the migrations.
```

PR #4296, opened 2024-09-20
Django FSM migration to Viewflow

rnovak338:
Part one of #4259

This branch migrates the use of the deprecating `django-fsm` library to using `django-viewflow` instead.

- Added new `django-viewflow` and `django-filter` dependencies to `requirements.txt`.
- New file `viewflow.py` under `/audit/models/` which contains the FSM logic for transitioning an SAC.
- Moved `STATUS` enumeration outside of the `SingleAuditChecklist` class. This required import/reference changes across many files and tests.
- Removed references of old deprecated library `django-fsm`.
- New migration to handle the changing the SAC's `submission_name` field to remove dependency on the deprecated `django-fsm`.

This should be blocked until #4292 (the race condition change) is implemented, since that PR includes some potential git conflicts with this branch that will need to be resolved before merging into `main`.

**IMPORTANT** - `django-fsm` is still referenced as an import for a historical migration in `/backend/audit/`. Since we are trying to remove the `django-fsm` library, this will need to be addressed in a later PR. While this deprecated library is no longer being used anywhere in the system, it still needs to be fixed in the old migration before it can be removed from `requirements.txt`.

## How to test
First, make sure that your local is able to install the new `viewflow` dependency with `docker-clean` and `docker-first-run`.

Ensure both backend tests with `python manage.py test` and the E2E tests succeed. For bonus points, run through the submission process manually to verify that each of the states are rendering the proper name for the submission as you go through the steps.

## PR Checklist: Submitter

- [X]   Link to an issue if possible. If there’s no issue, describe what your branch does. Even if there is an issue, a brief description in the PR is still useful.
- [X]   List any special steps reviewers have to follow to test the PR. For example, adding a local environment variable, creating a local test file, etc.
- [X]   Make sure you’ve merged `main` into your branch shortly before creating the PR. (You should also be merging `main` into your branch regularly during development.)
- [X]   Make sure you’ve accounted for any migrations. When you’re about to create the PR, bring up the application locally and then run `git status | grep migrations`. If there are any results, you probably need to add them to the branch for the PR. Your PR should have only **one** new migration file for each of the component apps, except in rare circumstances; you may need to delete some and re-run `python manage.py makemigrations` to reduce the number to one. (Also, unless in exceptional circumstances, your PR should not delete any migration files.)
- [X]   Make sure that whatever feature you’re adding has tests that cover the feature. This includes test coverage to make sure that the previous workflow still works, if applicable.
- [X]   Make sure the full-submission.cy.js [Cypress test](https://github.com/GSA-TTS/FAC/blob/main/docs/testing.md#end-to-end-testing) passes, if applicable.
- [X]   Do manual testing locally. Our tests are not good enough yet to allow us to skip this step. If that’s not applicable for some reason, check this box.

## PR Checklist: Reviewer


danswick (approved):
This looks good to me!

Let's keep an eye on the next staging deploy to make sure the migration and E2E tests work as expected.

JFYI: I also lightly edited your description so it wouldn't close the related issue (which still tracks the follow-on migration work) when this PR merges

PR #4345, opened 2024-10-02
Remove bpmn pr check

asteel-gsa:
Removes BPMN auto sync


jadudm (approved):
LGTM. This is an automation we do not need at this time.

Issue #4272, opened 2024-09-11
Fix Report Stuck in Submitted State and Update Code Logic


sambodeme:
**GitHub Ticket: Fix Report Stuck in Submitted State and Update Code Logic**

**Context:**
This issue is related to helpdesk ticket #4652. A user reported that their report is stuck in the "submitted" state and does not display the "accepted" badge. 

For reference, after a successful submission, a report should transition from `auditee_certified` to `submitted` while the dissemination process is ongoing. Once the dissemination completes successfully, the report should then transition from `submitted` to `disseminated`. However, in the case associated with ticket #4652, the final transition from `submitted` to `disseminated` never occurred.

**Issue:**
Upon investigation, it was unclear whether this issue was caused by human error (the user possibly clicked "submit" multiple times while the first submission was still processing) or a system issue (potential race conditions during the state transition and associated database operations). 

Regardless of the root cause, it is evident that the dissemination part of the process did not execute as expected. This led the submission process into the [else ](https://github.com/GSA-TTS/FAC/blob/main/backend/audit/views/views.py#L747) block of an `if` statement, where there was no logic to handle this specific scenario, resulting in the report being stuck in the "submitted" state.

**Proposed Fixes:**
1. **Update the Code:**
   - Modify the [else](https://github.com/GSA-TTS/FAC/blob/main/backend/audit/views/views.py#L747) block to properly handle the case where dissemination fails, ensuring that the process can continue without getting stuck. Add proper logic and error handling (beyond just `pass`), including logging for better traceability.

2. **Add a Django Command or Admin Process:**
   - Provide a Django command or admin process to allow administrators to transition reports that are stuck in the "submitted" state.
   - The process should follow sac waiver approach and:
     1. Perform full validation of the report.
     2. If validation is successful, trigger the dissemination and transition the report from `submitted` to `disseminated`.
     3. If validation fails, transition the report back to `auditee_certified` so the user can address any issues and resubmit.
   
This fix will prevent future reports from being trapped in this state and ensure smooth transitions between report states.

PR #4351, opened 2024-10-03
New management command - Manually move "submitted" report to dissemination

rnovak338:
Addresses part 2 of #4272 - providing an administrative command to move a SAC to the `disseminated` state.

The command is: `python manage.py move_to_disseminated --report_id ID_GOES_HERE`. This will look for a SAC with the `report_id` that you entered, and attempt to move it to the `disseminated` status - ONLY if it is stuck as `submitted` and passes cross-validation with no errors.

In the event that dissemination fails, or cross-validation fails, the SAC is returned to the `auditee_certified` state.

Additionally, this PR brings in some curation work from `jadudm/curation-api`. An app for `curation` has been created, which builts SQL scripts for tracking the before/after state of an audit. The above management command will track the audit's state before and after dissemination if it is successful.

## How to test

Before attempting the steps below, make sure you do a `docker-clean` and `docker-first-run`.

1. Create a SAC and bring it to the disseminated state.
2. Using a RDBMS, find this SAC, change `submission_status` to `submitted`, and ADD a row to `transition_name` and `transition_date`, which should be `disseminated` and a date for this action respectively.
3. Truncate the `dissemination_general` table to remove all dissemination data of this SAC. This will allow the management command to attempt to redisseminate without any issue.
4. Run the management command against your SAC.
5. Verify that the `fac_accepted_date` of the disseminated record matches the FIRST occurrence that the SAC was submitted (There should be two instances of `submitted` in `transition_name` and `transition_date` at this point).
6. Verify there is a record of this change in the `curation.record_version` table.

## PR Checklist: Submitter

## PR Checklist: Reviewer


rnovak338:
@jadudm Any thoughts when it comes to testing or test coverage for this new command?

My approach felt very hacky - I added manual transitions to the FSM in-line to bring a report to the `submitted` state (skipping through each step without any data) to verify that cross-validation would fail. I also allowed a report from the `auditee_certified` state to go through (and bypassing the `submitted` state condition that is required by the command).

In other words, I manipulated the logic in the command in order to unit test, but I'm scratching my head on how to "recreate" the user experience for testing this command end to end.

backend/audit/management/commands/move_to_disseminated.py
```diff
@@ -0,0 +1,105 @@
...
+        # check for validation errors.
+        errors = sac.validate_full()
+        if errors:
+            logger.info(
+                f"Unable to disseminate report with validation errors: {report_id}."
+            )
+            logger.info(errors["errors"])
+
+            # return to auditee_certified.
+            sac_revert_from_submitted(sac)
```
jadudm, line 67: `enable_audit_curation()` around this, because it is a curatorial act.

backend/audit/management/commands/move_to_disseminated.py
```diff
@@ -0,0 +1,105 @@
...
+            disseminated = sac.disseminate()
+            # `disseminated` is None if there were no errors.
+            if disseminated is None:
+                sac_transition(None, sac, transition_to=STATUS.DISSEMINATED)
+        # END ATOMIC BLOCK
+        disable_audit_curation()
+
+        # IF THE DISSEMINATION SUCCEEDED
+        # `disseminated` is None if there were no errors.
+        if disseminated is None:
```
jadudm, line 83: Collapse to an `if/else`, because it is testing a boolean condition. Only one of these two can happen.

backend/audit/models/viewflow.py
```diff
@@ -7,6 +7,25 @@
...
+    """
+    Transitions the submission_state for a SingleAuditChecklist back
+    to "auditee_certified" so the user can re-address issues and submit.
+    This should only be executed via management command.
+    """
+
+    if sac.submission_status == STATUS.SUBMITTED:
+        flow = SingleAuditChecklistFlow(sac)
+
+        flow.transition_to_auditee_certified()
```
jadudm, line 20: Wrap:

```
with curation_tracking():
  sac.save()
```

Perhaps we should always wrap our `with` as close to the action as possible?

jadudm: https://stackoverflow.com/questions/28157929/how-to-safely-handle-an-exception-inside-a-context-manager

We might want to do a bit more reading to make sure we don't accidentally leave tracking on.

We should probably `disable_audit_tracking` somwhere at startup, as an "always-just-in-case"...

backend/run.sh
```diff
@@ -19,6 +20,11 @@ source tools/seed_cog_baseline.sh
 setup_env
 gonogo "setup_env"
 
+#####
```
jadudm, line 23: Move to the end/late in the boot sequence. After the whole API standup/migrate/teardown dance.

backend/tools/curation_audit_tracking_init.sh
```diff
@@ -0,0 +1,9 @@
+source tools/util_startup.sh
+
+function curation_audit_tracking_init {
+    startup_log "CURATION_AUDIT_TRACKING" "BEGIN"
+    python manage.py curation_audit_tracking --init
```
jadudm, line 5: Always disable audit tracking right after the init.

backend/run.sh
jadudm: NEVERMIND. Leave this early in the sequence.

WHY?

What if we crashed, and the DB is in `enable` mode? We could enter the `migration` stage, run a big migration, and rewrite every row of `singleauditchecklist` That would all be documented as a curation, when it kinda isn't. (Eg. if we add a column with a new default value.) 

So, we should make sure it is disabled *before* the middle sequence, not after.

jadudm (dismissed):
Excellent stuff.

We're going to add a `with` context manager for the enable/disable, because it might help prevent a condition where audit tracking is left on, either accidentally or because of a system error.

Other notes inline. Small changes. Looks great overall.

Issue #3409, opened 2024-02-15
Enable removal of editors, certifiers from SF-SAC records


jadudm:
## What problems are we trying to solve?

Currently, it is possible to:

1. Add editors to an audit submission.
2. Change certifying officials.
3. Change editors.

It is not possible to *remove* editors. This is a problem.

## What can we do today?

* We will add a feature granting users with the Audit Editor role the ability to remove that role from others.
* We will not allow users with the Audit Editor role to remove themselves from that role. (In practice, this will mean that they must add another user with that role, and that user must then remove them.)
* We will not add the ability to remove Certifying Official roles without supplying new users for those roles; we will evaluate how many requests we get that are for exactly this operation.
* We will not add any out-of-band processes for adding or removing users from submissions.

## Context and further reading

We have discussed [via the ADR process](https://github.com/GSA-TTS/FAC/issues/3881). 

*Link to ADR when it is complete.*

```[tasklist]
### Requirements
- [x] Add a Remove link to the current access management table
- [x] Add a Remove page for deleting an Auditor Access role
- [x] Create a method for audit.views.manage_submission_access.ChangeOrAddRoleView to support deletion
- [x] Ensure that the new method does not allow deletion when the email in the Access object being deleted matches the email of the current user
- [ ] https://github.com/GSA-TTS/FAC/issues/4404
- [ ] https://github.com/GSA-TTS/FAC/issues/4403
- [ ] Track how many requests we get for removing Certifying Official roles without supplying a replacement
- [x] Potentially add documentation suggesting that larger organizations always add at least two of their staff in the Audit Editor role, to ensure that they do not get stuck without access to an audit if one of their staff members leaves
```

jadudm:
The "verify completed audits" story is supported in part by tickets like this:

https://fac-gov.zendesk.com/agent/tickets/2356

James-Paul-Mason:
### Main User Management Page

The table below lists the individuals with access to this single audit submission.

There must only be one Auditee Certifying Official and one Auditor Certifying Official. These users must have different email addresses. Any substitution for these roles will remove the prior user.

To change an individual's role, click the “Edit” link in the table. At this time, you can only change the certifying officials, and this can only be done by an Audit Editor.

It's recommended that larger organizations always add at least two of their staff in the Audit Editor role. This ensures that an audit does not become inaccessible due to the departure of a staff member.

James-Paul-Mason:
### Remove access page

Remove Editor
[...]

PR #4359, opened 2024-10-07
Allowing users to delete audit editors

phildominguez-gsa:
Draft: Coordinating merge with static site update

Addresses https://github.com/GSA-TTS/FAC/issues/3409

In this PR:
- Adding a link in the audit access management page that lets an editor remove another editor, as long as it's not yourself
- Fixed centering for the Edit column
- Unit tests

Testing:
- Go to the user management page for an audit you've created (this way you are an editor.) Removal links should only exist for those with the editor role.
- Click the Remove link for your user's row and you should get a page saying you're not allowed to. You can use devtools to un-disable the button and submit, but it will error. Now is also a good time to confirm that clicking the cancel button returns you to the user management page.
- Add a new auditor with some test email. Clicking Remove for that editor should show the normal removal page. Clicking the submit button should remove them and return you to the use management page. You can see that they've been added to the`audit_deletedaccess` table as well.
- Trying to get to a removal page for an editor that doesn't exist for this audit should 404. To do this, change the `id` param to some random number. This should also happen if you use the `id` for a non-editor user, which you can find in `audit_access`.
- If you're logged in as someone with audit access who is NOT an editor (i.e. a certifier), you shouldn't see the remove column/links at all. If you try to go to a removal url directly, it will error. I think the easiest way to test this is by having the Cypress test create an audit so it can make your normal login email a certifier instead of an editor.

## PR Checklist: Submitter

## PR Checklist: Reviewer


phildominguez-gsa:
> I was able to go through each of your points and confirm the expected behavior.
> 
> A small thing I caught (shown below). It seems that for the "Remove Access" page, some of the details in the footer are not visible. For example, My audit has an `Entity Name` and `UEI` (and it renders on other pages), but not here. Oddly enough, the report ID still displays.
> 
> image

Good catch, should be good now

sambodeme:
@phildominguez-gsa This works as described in the ticket description. One thing I noticed, which could either be included here or turned into a separate ticket, is that when adding an editor, the app does not check if that editor (identified by their email) already exists, nor does it prevent duplicate creation. I was able to add myself as an editor a second time, but the system would not allow me to remove the duplicate record.

phildominguez-gsa:
> @phildominguez-gsa This works as described in the ticket description. One thing I noticed, which could either be included here or turned into a separate ticket, is that when adding an editor, the app does not check if that editor (identified by their email) already exists, nor does it prevent duplicate creation. I was able to add myself as an editor a second time, but the system would not allow me to remove the duplicate record.

Totally agree. My preference would be separate in case this change needs to be reverted, as we don't want to also revert the dupe fix, unless @jadudm has any objections.

backend/audit/templates/audit/remove-editor-access.html
```diff
@@ -0,0 +1,52 @@
...
+                            name="editor_id"
+                            type="hidden"
+                            value="{{ editor_id }}"
+                        />
+                        <span class="usa-error-message" id="email-error-message" role="alert">{{ errors.email }}</span>
+                </fieldset>
+                
+                {% if is_editor_removing_self %}
+                <p class="font-sans-md">
+                    In order to avoid an audit from become inaccesible to an organization, editors do not have permission to remove their own access.
```
jadudm, line 36: Simplify to "Editors do not have permission to remove their own access."

jadudm:
Lets ticket the dupe fix separately, agreed. Reviewed. LGTM. My testing mirrors others; one suggestion on text, and its good to go.

backend/audit/templates/audit/remove-editor-access.html
phildominguez-gsa: Done!

jadudm (approved):
LGTM. Provides a long-sought feature. Tested locally.

phildominguez-gsa:
> Lets ticket the dupe fix separately, agreed. Reviewed. LGTM. My testing mirrors others; one suggestion on text, and its good to go.

Created a bug ticket here: https://github.com/GSA-TTS/FAC/issues/4375

PR #4383, opened 2024-10-15
Sandbox Environment

asteel-gsa:
This is a fairly extensive PR that I would like to have 1-2 more folks test before merging. This has been tested, built, destroyed, and rebuilt by Matt as well as myself.

### Description
Fundamentally, nothing changes except the following:
- Adds `helper_scripts` directory to `sandbox` to allow for easy `./` commands to plan, apply and destroy the environment
- Creates a clone of `https-proxy` called `sandbox-proxy` as the core change to this is what module manages the routes
  - Routes now located in `shared/sandbox/routes.tf` rather than where they were (clamav and https-proxy)
-  Detailed README on how to use sandbox
- Adds `"SANDBOX"` to `settings.py`


terraform/shared/modules/https-proxy/prepare-proxy.sh
```diff
@@ -13,11 +13,12 @@ tmpdir=$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
 cd "$tmpdir"
 
 # Grab a copy of the zip file for the specified ref
-curl -s -L https://github.com/GSA-TTS/cg-egress-proxy/archive/${GITREF}.zip --output local.zip 
+curl -s -L https://github.com/GSA-TTS/cg-egress-proxy/archive/${GITREF}.zip --output local.zip
 
 # Zip up just the proxy/ subdirectory for pushing
 unzip -q -u local.zip \*/proxy/\*
-zip -q -j -r ${popdir}/proxy.zip cg-egress-proxy-*/proxy
+cd "${tmpdir}/cg-egress-proxy-main/proxy/" && zip -r -o -X "${popdir}/proxy.zip" ./ > /dev/null
```
asteel-gsa, line 20: Can't remember why I made this change. I think it makes sense, but will confirm locally and in preview

asteel-gsa: NVM, can remove this change. Think I forgot about it after I made the `sandbox-proxy` module clone

terraform/shared/modules/stream-proxy/prepare-proxy.sh
```diff
@@ -1,11 +1,9 @@
 #!/bin/sh
 
 # Exit if any step fails
-set -e
```
asteel-gsa, line 4: Need to undo this

terraform/shared/modules/stream-proxy/stream-proxy.tf
```diff
@@ -1,97 +1,97 @@
...
+###
+
+resource "cloudfoundry_user_provided_service" "credentials" {
+  name  = "${var.name}-creds"
+  space = data.cloudfoundry_space.client_space.id
+  credentials = {
+    "domain" = cloudfoundry_route.egress_route.endpoint
+    "port"   = 8080
+  }
+}
```
asteel-gsa, line 97: CRLF -> LF Change

backend/Procfile
```diff
@@ -1,2 +1,2 @@
-# used by cloud.gov
-web: newrelic-admin run-program gunicorn config.wsgi -c gunicorn.conf.py
+# used by cloud.gov
+web: newrelic-admin run-program gunicorn config.wsgi -c gunicorn.conf.py
```
asteel-gsa, line 2: CRLF -> LF Change (I think)

terraform/shared/modules/stream-proxy/prepare-proxy.sh
asteel-gsa: Fixed

sambodeme:
After an initial walkthrough session with Alex, I ran destroy.sh, followed by plan.sh and apply.sh. The apply command initially failed with an error that appeared to be a random blip, so I reran plan.sh and apply.sh, and this time it worked. Wanted to do a complete rerun again from the start.

rnovak338:
Had a successful runthrough with Alex last week myself, and yesterday I was able to go through README on my own time and get the helper scripts running successfully.

sambodeme (approved):

Issue #4411, opened 2024-10-23
HTML/DOM interpretation fix


jadudm:
### Tasks

Again, a relatively straight-forward fix: make sure the content does not get redirected straight back.

```[tasklist]
- [ ] https://github.com/GSA-TTS/FAC/security/code-scanning/1557
```

PR #4416, opened 2024-10-24
Updated check-ueid.js to handle html elements

anagradova:
safely to prevent xss

## PR Checklist: Submitter

Issue: https://github.com/GSA-TTS/FAC/issues/4411
Steps to validate: 
- Change can be seen on /report_submission/auditeeinfo/ after submitting a valid UEI. 
- Inspect the javascript (check-ueid.js) and look for the function (showValidUeiInfo).

## PR Checklist: Reviewer

### DIFF
--- a/backend/static/js/check-ueid.js
+++ b/backend/static/js/check-ueid.js
@@ -43,15 +43,24 @@ function showValidUeiInfo() {
   const auditeeUei = document.getElementById('auditee_uei').value;
   const auditeeName = document.getElementById('auditee_name');
   const ueiInfoEl = document.createElement('div');
+  const dl = document.createElement('dl');
+  const dtUei = document.createElement('dt');
+  const ddUei = document.createElement('dd');
+  const dtName = document.createElement('dt');
+  const ddName = document.createElement('dd');
+ 
+  dl.setAttribute('data-testid', 'uei-info');
+  dtUei.textContent = 'Unique Entity ID';
+  ddUei.textContent = auditeeUei;
+  dtName.textContent = 'Auditee name';
+  ddName.textContent = auditeeName.value;
 
-  ueiInfoEl.innerHTML = `
-    <dl data-testid="uei-info">
-      <dt>Unique Entity ID</dt>
-      <dd>${auditeeUei}</dd>
-      <dt>Auditee name</dt>
-      <dd>${auditeeName.value}</dd>
-    </dl>
-  `;
+  dl.appendChild(dtUei);
+  dl.appendChild(ddUei);
+  dl.appendChild(dtName);
+  dl.appendChild(ddName);
+
+  ueiInfoEl.appendChild(dl);
 
   auditeeName.removeAttribute('disabled');
   auditeeName.parentNode.setAttribute('hidden', 'hidden');


phildominguez-gsa (dismissed):

jperson1 (approved):

danswick:
These changes look fine to me, but it's important to call out that they appear to be copied and pasted from the GitHub Copilot suggestion (with some reorganizing). I trust that the reviewers knew this and discussed it, but we need to be very careful about incorporating AI-derived code into the codebase. These changes are relatively minor, but it's still worth reminding everyone.

anagradova:
> These changes look fine to me, but it's important to call out that they appear to be copied and pasted from the GitHub Copilot suggestion (with some reorganizing). I trust that the reviewers knew this and discussed it, but we need to be very careful about incorporating AI-derived code into the codebase. These changes are relatively minor, but it's still worth reminding everyone.

That is correct @danswick, it is good to review AI generated suggestions, which I did. This is very similar to the co-pilot suggestion because there was no reason to change it very much other than some organization as its suggested implementation accomplished the goal, which was correcting the potential for XSS by placing in tags/strings into into the object model instead of a string with tags. The logic was not complex.

jadudm:
I'll be out of the room for a while, so the team will want to discuss some norms around the AI code.

1. We have a system that is CC0/public domain in the USA. The AI code is trained on code that has unknown licenses and unknown provenance. I have no idea what it means for us to use it. (Which is perhaps a way to say "given our license, perhaps we shouldn't.")
2. The license issue also applies to assets, FWIW, and might help frame the code issue. E.g. when we do public presentations, we can't use random images from the internet, because it re-licenses the content as public domain (or otherwise makes dissemination of our work _complex_.) So, similarly for us... licensing matters, and for our code as well.

(For reference, when you're reading SO, their license is pretty explicit, and generally compatible as long as we reference our work (BY-SA). https://stackoverflow.com/help/licensing)

In this regard... my personal inclination would be to never use it. However, the team can discuss and perhaps do an ADR about it.

asteel-gsa:
Dismissing reviews at this time for re-review. Please reference the Github Copilot original suggestion vs current adaptation for formal review.

asteel-gsa:
@anagradova spoke with @jperson1 and this would be my proposed change to get it through.
James was able (barring local build issues) to get this working and feels good about this as well. Let us know what you think.

CC: @danswick @jadudm 
```js
let dl; let dtUei; let ddUei; let dtName; let ddName;
dl = document.createElement('dl');
dtUei = document.createElement('dt');
ddUei = document.createElement('dd');
dtName = document.createElement('dt');
ddName = document.createElement('dd');

dl.setAttribute('data-testid', 'uei-info');
dtUei.textContent = 'Unique Entity ID';
ddUei.textContent = auditeeUei;
dtName.textContent = 'Auditee name';
ddName.textContent = auditeeName.value;

dl.append(dtUei,ddUei,dtName,ddName);
ueiInfoEl.appendChild(dl);
```

jperson1 (approved):

danswick (approved):
Pulled this down, built, and tested. Looks good!

gsa-jrothacker (approved):
I don't have any concerns with your actual change here.

However, I'm curious why it looks like we're doing a lot of dom manipulation inside this javascript. I'm not claiming to be a frontend expert, but I would imagine there is a framework, library, or just a better way?

PR #4443, opened 2024-11-08
Cross validation and waivers for prior references

phildominguez-gsa:
Closes https://github.com/GSA-TTS/FAC/issues/4444

In this PR:
- Verifying existence of prior findings reference number when listed in a finding. Here are the steps for doing this:
  - Waiver exists -> Valid.
  - No prior references in the report -> Valid.
  - The report is from before 2023 (because UEIs only consistent starting 2022) -> Valid.
  - Fetch all disseminated reports for the UEI and look for the prior reference in those. Not found -> Invalid.
- Able to create waivers in the admin panel
- Unit tests
- Tweaking Cypress tests and workbooks to use 2023 submission dates. This is because reports earlier than this skip prior reference validation.

image

Testing:
- Note: This has a migration, so be sure those run
1. Modify `full-submission.cy.js` so that it only runs one of the three tests, just to make things faster
2. Run the full-submission test and it should pass as normal. This will also create a submission that we will make use of in later tests.
3. Open up `backend/cypress/fixtures/test_workbooks/federal-awards-audit-findings-workbook.xlsx` and go to the Form tab. Set J2 to 'Y' and set K2 to '2023-999'. This indicates that the finding has a prior reference.
4. In `full-submission.js`, comment out everything after and including line 96. This way it will set up a new audit and we can just click the `Begin Validation` button for the remainder of the tests. Run full-submission again to get up to this point. Note the `report_id` for later steps.
5. ❌ Clicking the button now will cause cross validation to fail with an error saying the prior finding couldn't be found. This is correct because it we didn't set up the prior reference number on the report we made in step 2.
6. Go to http://localhost:8000/admin/audit/sacvalidationwaiver/add/ and create a waiver for the `report_id` from step 4, and select "Report has invalid prior reference numbers"
7. ✅ Clicking the button now should pass cross validation since the waiver lets it ignore errors
8. ❌ Delete the waiver you just created so that we can continue to set up a legit submission. Clicking the button should now fail cross validation again.
9. Go to `dissemination_finding`. Find the matching `report_id` from step 4 and modify the `reference_number` to be '2023-999'. This matches what we set up in the workbook in step 3.
10. ✅ Clicking the button now should pass cross validation

## PR Checklist: Submitter

## PR Checklist: Reviewer


backend/audit/cross_validation/check_finding_prior_references.py
```diff
@@ -0,0 +1,136 @@
...
+    general_information = all_sections.get("general_information")
+    auditee_uei = general_information["auditee_uei"]
+    previous_year = (
+        date.fromisoformat(general_information["auditee_fiscal_period_start"]).year - 1
+    )
+
+    # Prior reference numbers usually refer to findings from the previous
+    # year's report
+    try:
+        previous_year_report = General.objects.get(
```
sambodeme, line 47: This assumes that the combination of (audit_year, auditee_uei) corresponds to a single record, which is not always the case. For example, `ZZURKVEBSJL1` maps to two reports in 2022 and to two reports in 2023. There are cases where the same auditee_uei maps to multiple reports within a given audit year

sambodeme: image

backend/audit/cross_validation/check_finding_prior_references.py
```diff
@@ -0,0 +1,136 @@
...
+                {
+                    "error": err_bad_repeat_prior_reference(award_ref),
+                }
+            )
+            continue
+
+        try:
+            # The first four digits of a prior reference number are the year
+            prior_ref_year = int(prior_ref[:4])
+            prior_ref_report = General.objects.get(
```
sambodeme, line 106: Same comment as previous

backend/audit/cross_validation/check_finding_prior_references.py
```diff
@@ -0,0 +1,136 @@
...
+                audit_year=prior_ref_year,
+                auditee_uei=auditee_uei,
+            )
+            prior_ref_report_id = prior_ref_report.report_id
+        except General.DoesNotExist:
+            prior_ref_report_id = None
+
+        # UEIs only become reliable as of 2022, so don't bother invalidating
+        # prior references before that if we can't find it's report
+        if not previous_year_report_id and not prior_ref_report_id:
```
sambodeme, line 116: what is the expected behavior for cases where `(previous_year_report_id or prior_ref_report_id) and prior_ref_year < 2022`

backend/audit/cross_validation/check_finding_prior_references.py
phildominguez-gsa: Any objections to searching for the prior ref in all of those reports, then?

backend/audit/cross_validation/check_finding_prior_references.py
phildominguez-gsa: My thought process was that if the previous report was found, then it probably would/should contain the prior ref, regardless of the year.

phildominguez-gsa: Should this whole validation be skipped then if the submission is pre-2022?

backend/audit/cross_validation/check_finding_prior_references.py
phildominguez-gsa: Fixed

backend/audit/cross_validation/check_finding_prior_references.py
phildominguez-gsa: Fixed

backend/audit/cross_validation/check_finding_prior_references.py
phildominguez-gsa: Done

backend/audit/cross_validation/check_finding_prior_references.py
```diff
@@ -0,0 +1,122 @@
...
+def _validate_prior_refs(
+    prior_refs, award_ref, auditee_uei, previous_report_ids, errors
+):
+    """
+    Performs validation on the given list of prior reference numbers
+    """
+    for prior_ref in prior_refs:
+        prior_ref_year = prior_ref[:4]
+
+        if prior_ref == "N/A":
```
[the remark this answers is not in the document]
phildominguez-gsa: Done

sambodeme (approved):
Works great!

Issue #3903, opened 2024-05-29
Investigate cog/over outside base years


jadudm:
A question has been asked about the cog/over algorithm. 

Specifically, the question is: are we handling assignments outside of base years correctly?

The flow, as-is, is shown below. (The flow was previously expressed as a Mural; this is a Mermaid transcription, which would allow us to put it under version control.)

```mermaid
flowchart TD
    classDef startClass fill:#000,stroke:#CCC,color:#FFF
    classDef endClass fill:#000,stroke:#CCC,color:#FFF
    classDef info fill:#F4D03F,stroke:#CCC
    classDef question fill:#F39C12,stroke:#CCC
    classDef overInfo fill:#AED6F1,stroke:#CCC
    classDef cogInfo fill:#F5B7B1,stroke:#CCC
    
    A(Start):::startClass --- B[Record is <br />submitted to the FAC]
    B:::info --> C{Total expenditure <br />less than <br />or equal to $50M?}
    C:::question
    C -->|YES| O1
    C -->|NO| C0
    subgraph Oversight
    O1[Determine Oversight]:::overInfo
    O2:::question
    O1 --> O2
    O2{Does record have\nDA expenditures?}
    O2 -->|NO| O4
    O2 -->|YES| O3
    O4:::overInfo
    O3:::overInfo
    O3[Sum DA<br />expenditures to <br />unique ALN prefixes]
    O4[Agency with<br />predominant amount<br />of total funding\is oversight]
    O3 --> O6
    O4 --> O5[End]:::endClass
    O6:::question
    O6{Does agency<br />with largest DA<br />expenditure make up<br />at least 25%<br />of total expenditure?}
    O6 -->|YES| O8
    O6 -->|NO| O7
    O8:::overInfo
    O7:::overInfo
    O8[Agency with<br />predominant DA<br />expenditure is<br />oversight]
    O7[Agency with<br />predominant total<br />expenditure is <br />oversight]
    O8 --> O9[End]:::endClass
    O7 --> O10[End]:::endClass
    end
    subgraph Cognizance
    C0:::cogInfo
    C1:::question
    C2:::cogInfo
    C3:::cogInfo
    C4:::endClass
    C0[Determine Cognizance]
    C0 --> C1{Does<br />entity exist in<br />Cog_Ref_List}
    C1 -->|YES| C2[Listed Cog is assigned.]
    C1 -->|NO| C3[Look back at base year<br />for the submission]
    C2 --> C4(END)
[...]

PR #4489, opened 2024-11-27
Sk/cog rebase

gsa-suk:
## PR Checklist: Submitter

## PR Checklist: Reviewer

### DIFF
--- a/backend/support/cog_over.py
+++ b/backend/support/cog_over.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 import logging
+import math
 
 from django.db.models.functions import Cast
 from django.db.models import BigIntegerField, Q
@@ -14,7 +15,7 @@ COG_LIMIT = 50_000_000
 DA_THRESHOLD_FACTOR = 0.25
 
 
-def compute_cog_over(federal_awards, submission_status, auditee_ein, auditee_uei):
+def compute_cog_over(federal_awards, submission_status, auditee_ein, auditee_uei, audit_year):
     """
     Compute cog or oversight agency for the sac.
     Return tuple (cog_agency, oversight_agency)
@@ -25,6 +26,7 @@ def compute_cog_over(federal_awards, submission_status, auditee_ein, auditee_uei
     - sac.ein
     - sac.auditee_uei
     """
+    base_year = calc_base_year(audit_year)
     if not federal_awards:
         logger.warning(
             f"Trying to determine cog_over for a sac with zero awards with status = {submission_status}."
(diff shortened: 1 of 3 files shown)


phildominguez-gsa:
Lgtm. Is there a quick way to test this works in the app itself?

rnovak338:
> 2\. docker compose run web python manage.py test support.test_cog_over

+1 to Phil - I was able to get successful test results myself.

backend/support/cog_over.py
```diff
@@ -25,6 +28,7 @@ def compute_cog_over(federal_awards, submission_status, auditee_ein, auditee_uei
     - sac.ein
     - sac.auditee_uei
     """
+    base_year = calc_base_year(audit_year)
```
gsa-jrothacker, line 31: nitpick: I would move this line closer to where you use it. (around line 52)

backend/support/cog_over.py
```diff
@@ -45,13 +49,24 @@ def compute_cog_over(federal_awards, submission_status, auditee_ein, auditee_uei
...
 
 
+def calc_base_year(audit_year):
+    # Note: 2019 is the first supported baseline year in GSAFAC
+    # For audit years 2019 through 2023, baseline year is 2019
+    # For audit years 2024 through 2028, baseline year is 2024
+    # For audit years 2029 through 2033, baseline year is 2029
+    # For audit years 2034 through 2038, baseline year is 2034
+    # and so on
+    base_year = (math.floor((int(audit_year) - 2019) / 5) * 5) + 2019
```
gsa-jrothacker, line 66: 2019 and 5 appear to be magic numbers. I would prefer if pulled these out into constants. (2019 is also used below in line 97)

gsa-suk:
Added an advanced test.

backend/support/cog_over.py
```diff
@@ -13,8 +14,15 @@
...
+FIRST_BASELINE_YEAR = 2019
+BASELINE_VALID_FOR_YEARS = 5
 
-def compute_cog_over(federal_awards, submission_status, auditee_ein, auditee_uei):
+DBKEY_TO_UEI_TRANSITION_YEAR = "2022"
+
+
+def compute_cog_over(
+    federal_awards, submission_status, auditee_ein, auditee_uei, audit_year
+):
```
sambodeme, line 25: This comment is not directly related to the current change, but since the PR modifies this function, I’d like to take the opportunity to clarify some points and potentially simplify the logic for improved readability.
While reviewing this function, I noticed that it checks for None or empty awards at [line 36](https://github.com/GSA-TTS/FAC/pull/4489/files#diff-c25fd3c8576fba3f3c065573fc604b1501208e68bb87c074bb8daf3b46124d0dR36). For context, we compute cog_over as part of dissemination, and based on our current [cross-validations](https://github.com/GSA-TTS/FAC/blob/sk/cog_rebase/backend/audit/cross_validation/check_has_federal_awards.py#L4), we do not allow reports with no awards to proceed. This makes the check at line 36 seem unnecessary to me.
If anything, I believe it would be better to throw an exception here to indicate an issue with the report. Since I wasn’t part of the discussions that led to this logic, do you recall any context or conversations around this decision? I also noticed a similar check on [line 118](https://github.com/GSA-TTS/FAC/pull/4489/files#diff-c25fd3c8576fba3f3c065573fc604b1501208e68bb87c074bb8daf3b46124d0dL90).

sambodeme:
The code changes look good to me. After reviewing Matt's [diagram ](https://github.com/GSA-TTS/FAC/issues/3903) and the overall cog_over logic, I think we could consider refactoring the code to align more closely with the diagram. While the current implementation does this to some extent, having the flow in the code mirror the diagram more precisely could make it easier to understand what the code is doing (or is supposed to do) when compared to the diagram. Going through this exercise may also require us to update the diagram (or the code) whenever necessary. This is unrelated to this PR and would be more of a nice-to-have improvement.

Issue #4437, opened 2024-11-06
Bug: 500 on Search with Invalid Date


jperson1:
When a user makes a search with an invalid date, we throw a ValidationError, which results in a 500. This is the majority of 500s we see for search in the log reviews. 

Example:
```
Traceback (most recent call last):
  File "/home/vcap/deps/1/python/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/vcap/deps/1/python/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vcap/deps/1/python/lib/python3.12/site-packages/newrelic/hooks/framework_django.py", line 527, in wrapper
    return wrapped(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vcap/deps/1/python/lib/python3.12/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vcap/deps/1/python/lib/python3.12/site-packages/django/utils/decorators.py", line 48, in _wrapper
    return bound_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vcap/deps/1/python/lib/python3.12/site-packages/django/views/decorators/csrf.py", line 65, in _view_wrapper
    return view_func(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vcap/app/dissemination/views.py", line 237, in dispatch
    return super(Search, self).dispatch(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vcap/deps/1/python/lib/python3.12/site-packages/newrelic/hooks/framework_django.py", line 900, in wrapper
    return wrapped(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4497, opened 2024-12-02
Resolve 500 error with dissemination search/advanced

rnovak338:
Addresses #4437.
- Add form error labels for start and end dates in search.
- Re-render search template instead of raising validation errors.

## How to test

1. Boot up the application.
2. Attempt to search with invalid date formats under "FAC Accepted Date" from the search and advanced search pages.
3. Each of the pages should reload with an error message over the date field.

## PR Checklist: Submitter

## PR Checklist: Reviewer

### DIFF
--- a/backend/dissemination/templates/search-alert-error.html
+++ b/backend/dissemination/templates/search-alert-error.html
@@ -4,10 +4,12 @@
     <div class="usa-alert__body">
         <h2 class="usa-alert__heading">Error</h2>
         <p>The information you entered is invalid. Please double-check your information for errors.</p>
+        {% if errors %}
         <ul>
           {% for error in errors %}
           <li>{{ error }}</li>
           {% endfor %}
         </ul>
+        {% endif %}
     </div>
 </div>
--- a/backend/dissemination/templates/search-alert-error.html
+++ b/backend/dissemination/templates/search-alert-error.html
@@ -3,6 +3,11 @@
 <div class="usa-alert usa-alert--error">
     <div class="usa-alert__body">
         <h2 class="usa-alert__heading">Error</h2>
-        <p>Oops! It looks like some of the information you entered is invalid. Please double-check your information for errors.</p>
+        <p>The information you entered is invalid. Please double-check your information for errors.</p>
+        <ul>
+          {% for error in errors %}
+          <li>{{ error }}</li>
+          {% endfor %}
+        </ul>
     </div>
 </div>
--- a/backend/dissemination/templates/search-alert-error.html
+++ b/backend/dissemination/templates/search-alert-error.html
@@ -1,0 +1,8 @@
+{% load humanize %}
+{% load sprite_helper %}
+<div class="usa-alert usa-alert--error">
+    <div class="usa-alert__body">
+        <h2 class="usa-alert__heading">Error</h2>
+        <p>Oops! It looks like some of the information you entered is invalid. Please double-check your information for errors.</p>
+    </div>
+</div>
--- a/backend/dissemination/search.py
+++ b/backend/dissemination/search.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import time
 from .searchlib.search_constants import ORDER_BY, DIRECTION, DAS_LIMIT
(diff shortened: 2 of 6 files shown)


phildominguez-gsa:
This does work well, just a couple questions/suggestions!

No luck getting this to show up?
image

If it's not doable, I think we'd need to better indicate that the search failed due to an error. Right now you can't really tell until you expand the date section again. Maybe put a message where the search results show up, something like "Search submission failed" with a list of all the errors?

Ideally the form should retain all the info the user submitted as well, if possible. I could see it being pretty frustrating if everything got wiped because you entered a bad date 😬

rnovak338:
> This does work well, just a couple questions/suggestions!
> 
> No luck getting this to show up, like in the step 2 form? image
> 
> If it's not doable, I think we'd need to better indicate that the search failed due to an error. Right now you can't really tell until you expand the date section again. Maybe put a message where the search results show up, something like "Search submission failed" with a list of all the errors?
> 
> Ideally the form should retain all the info the user submitted as well, if possible. I could see it being pretty frustrating if everything got wiped because you entered a bad date 😬

Hey Phil,

Thanks for your feedback. That tooltip for the form field is generated by the browser, and it does not seem like we can easily guarantee a way to get it to generate without changing up some of the front-end logic quite a bit. I updated the PR with the following changes for both search AND advanced search. Please double check this on your end when you are ready!

New USWDS error component renders on invalid form field. It looks like this:
image

Your inputted fields will keep the values they had before submission (and the appropriate expanders are open/closed). The error will appear like this:
image

Give it a test on both basic and advanced search.

jperson1:
^ Relatedly, my browser won't give me an error the first time I put in a bad date. But if I get the bad date error from submitting and _then_ go in with a bad date, my browser will let me know. Not the first time, when I would have needed it. Weird. Oh well.

image

backend/dissemination/templates/search-alert-error.html
```diff
@@ -0,0 +1,8 @@
+{% load humanize %}
+{% load sprite_helper %}
+<div class="usa-alert usa-alert--error">
+    <div class="usa-alert__body">
+        <h2 class="usa-alert__heading">Error</h2>
+        <p>Oops! It looks like some of the information you entered is invalid. Please double-check your information for errors.</p>
```
jperson1, line 6: Not sure about "Oops." Might be too friendly for the FAC's voice? @James-Paul-Mason, how do you feel?

We have a similar sort of message here:
https://github.com/GSA-TTS/FAC/blob/53c8abdb5e8c387d7de5ec987d430d96a1b8339f/backend/audit/templates/audit/manage-submission-change-access.html#L34-L36

phildominguez-gsa: Once we get guidance from James on this I'm good with approving. Works great!

James-Paul-Mason:
In terms of the error message, I would agree with @jperson1 that we should drop the "Oops!" for tonal consistency. Also, based on our voice/tone guide, we want our messaging to be authoritative, so I would delete "looks like"

<s>Oops! It looks like some of</s> the information you entered is invalid. Please double-check your information for errors.

From my reading of the ticket, it sounds like this error pops up when a user inputs an invalid End date. If that is the case, can we get more specific about "the information" in the error message, and replace it with something like this: "The date you entered is invalid. Please double-check the **End date** field for errors." 

Open to any suggestions if that sounds too specific!

rnovak338:
> In terms of the error message, I would agree with @jperson1 that we should drop the "Oops!" for tonal consistency. Also, based on our voice/tone guide, we want our messaging to be authoritative, so I would delete "looks like"
> 
> Oops! It looks like some of the information you entered is invalid. Please double-check your information for errors.
> 
> From my reading of the ticket, it sounds like this error pops up when a user inputs an invalid End date. If that is the case, can we get more specific about "the information" in the error message, and replace it with something like this: "The date you entered is invalid. Please double-check the **End date** field for errors."
> 
> Open to any suggestions if that sounds too specific!

Thanks @James-Paul-Mason, I updated the language to be authoritative. Regarding your other point, the changes in this PR will detect form errors beyond just the start/end dates in this form. Because the errors provided by Django are not always clear to the end user, I added some custom context to handle when the start/end dates are invalid. Like so:

image

I do not think it is possible for the other fields to have invalid input. Either way, moving forward we can provide custom information for them if needed.

phildominguez-gsa (approved):

jperson1 (approved):

PR #4499, opened 2024-12-03
Add 'On Demand' Backup Operation

asteel-gsa:
Expands the backup utility to accept an `on_demand_backup` operation.

Fundamentally, operates the exact same way as other backup commands, however, this one, after discussion with Sudha, will be prep for the SAC load. Instead of digging through the `scheduled` folder for the backup, we can isolate the data load to be at the following location: `s3://$BUCKET/backups/on-demand/MM-DD-HH/....` to better facilitate a smooth data loading operation.


gsa-suk (approved):
lgtm.

PR #4504, opened 2024-12-04
Remove pre-deploy zap scan from staging and prod deploy

asteel-gsa:
Removes the dev zap scan from the staging deployment.

v14 of zap action upgraded their `upload-artifact` version, which seems to not like having an existing artifact name in place. During a dev deployment, zap is done as one of the last steps. Rescanning before a staging deploy feels unnecessary.


jperson1 (approved):
Makes sense to me, especially given the explanation in Slack!

Issue #4544, opened 2024-12-13
Add program name advanced search filter; Full cog/over name in results


danswick:
squad: @anagradova @phildominguez-gsa @analynd 

Since some users may not know the exact codes for program names, we should give them an option to search by the full program name. Start with simple exact-string matching, but also see what can be done with fuzzy searching. If searching by program name ends up being untenable or problematic, document this and be able to demonstrate why this is the case. Note that since a single report can have multiple program names, it's not really viable to include them in the search results table.

Additionally, display the full cog/over agency names in the search results table instead of just the code. For example, "OVER-00" should now be "00 - Some Example Name".

```[tasklist]
### Outcomes
- [ ] Full cog/over name appears in search results table
- [ ] Program name is implemented as a new search filter
```

Original text:
This is a long-standing request from our users, particularly those working on oversight. We use a lot of alphanumerical codes to identify things like programs or agencies and these can be difficult to navigate if you're scanning a long list of audit reports. Using the full name should help alleviate some of that difficulty. 

Note: as I'm writing this out, I'm realizing there's a pretty glaring issue: our search results are sorted by _audit report_, not by program. A given audit report can be associated with _many_ programs, not just one. The squad above should discuss, but I'm wondering if I accidentally conflated two different things at some point. We probably want to put the full cog/over _agency name_ in the search results table, but still add a search filter for program name.

PR #4558, opened 2024-12-19
Searching audits by program name search and display full cog/over agency

phildominguez-gsa:
Closes https://github.com/GSA-TTS/FAC/issues/4544

In this PR:
- Audits can now be searched by program name. Query terms can be separated by newlines, and it generally ignores other delimiters. For example, "Foo-Bar\nBaz" will search for program names containing (foo and bar) or (baz), ignoring case, word ordering, and delimiters.
- Full agency names are now displayed in the cog/over column

Testing:
- Make some submissions that have filings with different program names. An easy way to do this is run full-submission a bunch and modify `federal_program_name` entries in the `dissemination_federalaward` table.
- Searching by program name should work as described above
- Full agency name and number should now displayed in the cog/over column

## PR Checklist: Submitter

## PR Checklist: Reviewer


backend/dissemination/searchlib/search_federal_program_name.py
```diff
@@ -0,0 +1,49 @@
...
+    names = params.get("federal_program_name", [])
+
+    if not names:
+        return general_results
+
+    for name in names:
+        q_sub = Q()
+
+        for sub in _split(name, text_input_delimiters):
+            q_sub.add(Q(federal_program_name__icontains=sub), Q.AND)
```
gsa-jrothacker, line 24: I'm curious about the need to do this? 

I like that we `OR` on each line, but should we search for matches of what they enter instead of splitting it up? I might be overthinking it but if someone searches for "A-Plus School" (let's assume this is the exact name of a non-profit), I imagine the query will hit on a ton of school districts too (Let's assume this exists too: **Plus**h **A**laska **School**)

Again, I'm not concerned with how the code is written I think it would work, but just curious if we'll encounter some weirdness.

backend/dissemination/searchlib/search_federal_program_name.py
```diff
@@ -0,0 +1,49 @@
...
+
+    filtered_general_results = general_results.filter(q).distinct()
+
+    t1 = time.time()
+    report_timing("federal_program_name", params, t0, t1)
+
+    return filtered_general_results
+
+
+def _split(text, delimiters):
```
gsa-jrothacker, line 36: I _think_ we can eliminate this all together and just use regex. Note: we'd have to create the delimiter regex, but IMO that's better.

```
import re

re.split(delimiter_regex, name)
```

backend/dissemination/templates/search.html
```diff
@@ -149,10 +150,8 @@ <h1 class="font-sans-2xl">Search single audit reports</h1>
                                         </td>
                                         <td>{{ result.audit_year }}</td>
                                         <td>
-                                            {% if result.oversight_agency %}
-                                                OVER-{{ result.oversight_agency }}
-                                            {% elif result.cognizant_agency %}
-                                                COG-{{ result.cognizant_agency }}
```
gsa-jrothacker, line 155: Should we update the header? I _think_ it makes sense leaving it at as `Cog or Over`, but just want to ask the question.

backend/dissemination/test_search.py
```diff
@@ -627,9 +627,57 @@ def test_alns_no_findings(self):
...
+            federal_program_name="Yub Nub",
+        )
+        self.refresh_materialized_view()
+
+        adv_params = {"advanced_search_flag": True}
+
+        # Search for multiple program valid names
+        params = {"federal_program_name": ["Foo", "Bar"], **adv_params}
+        results = search(params)
+        self.assertEqual(len(results), 2)
```
gsa-jrothacker, line 660: Should we also validate that the results have the actual records we are expecting?

Something like "results contains federal program foo"; "results contains federal program bar"

backend/dissemination/views.py
```diff
@@ -1,47 +1,41 @@
-from datetime import date, timedelta
 import logging
 import math
+import textwrap
```
gsa-jrothacker, line 3: I didn't look closely at all the changes in imports... I am assuming all the changes are from an IDE automatically reordering.

backend/dissemination/views.py
```diff
@@ -134,6 +129,27 @@ def run_search(form_data):
     return search(search_parameters)
 
 
+# Function to do a dictionary lookup of the agency name to number
+def _populate_cog_over_name(results):
```
gsa-jrothacker, line 133: Is it possible to add this a method on form data itself? I don't love iterating over results and modifying it to populate the agency name, i would much prefer if there was method like `result.get_agency_name` and that was implemented like:

```
def get_agency_name():
        if result.oversight_agency:
            agency_name = agency_names.get(
                result.oversight_agency, result.oversight_agency
            )
            return "\n".join(
                textwrap.wrap(agency_name + " (OVER)", width=20)
            )
        elif result.cognizant_agency:
            agency_name = agency_names.get(
                result.cognizant_agency, result.cognizant_agency
            )
            return "\n".join(
                textwrap.wrap(agency_name + " (COG)", width=20)
            )
       return "n/a"   # Not sure what we'd want here
```

gsa-jrothacker (commented):
I think this will be greatly appreciated by our users.

Let me know if there is anything you want to chat about offline.

backend/dissemination/searchlib/search_federal_program_name.py
```diff
@@ -0,0 +1,49 @@
...
+from django.db.models import Q
+
+import time
+
+
+def search_federal_program_name(general_results, params):
+    """
+    Searches on Federal Award field 'federal_program_name'.
+    """
+    t0 = time.time()
```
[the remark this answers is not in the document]
phildominguez-gsa: I think you're correct here! Since we do this in all the `search_*` methods though, I think we should make another ticket to address this all at once. Thoughts?

backend/dissemination/searchlib/search_federal_program_name.py
```diff
@@ -0,0 +1,49 @@
...
+    q = Q()
```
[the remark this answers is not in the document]
phildominguez-gsa: Agreed, maybe we address all these in the same work as https://github.com/GSA-TTS/FAC/pull/4558#discussion_r1910711672?

backend/dissemination/searchlib/search_federal_program_name.py
phildominguez-gsa: Hmm that's a fair point. Do you think removing `-` from the [list of delimiters](https://github.com/GSA-TTS/FAC/blob/06e398694bf95b4cc78c4f6100d505f6752fa313/backend/dissemination/searchlib/search_constants.py#L20) would be sufficient?

backend/dissemination/templates/search.html
phildominguez-gsa: Fair question! Yes, "Cog or Over" still works 👍

backend/dissemination/searchlib/search_federal_program_name.py
phildominguez-gsa: Note: By that I mean making a different list, since that one is used elsewhere

backend/dissemination/searchlib/search_federal_program_name.py
phildominguez-gsa: Done

backend/dissemination/test_search.py
phildominguez-gsa: Done

backend/dissemination/searchlib/search_federal_program_name.py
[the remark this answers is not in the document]
gsa-jrothacker: I wouldn't be opposed to creating a new ticket address it all at once, I would suggest we just leave out the timing here though.

backend/dissemination/searchlib/search_federal_program_name.py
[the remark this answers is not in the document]
gsa-jrothacker: Honestly, I'd prefer we do this now while you're already in here. Would give us one less method we need to fix.

backend/dissemination/searchlib/search_federal_program_name.py
```diff
@@ -0,0 +1,49 @@
...
+    names = params.get("federal_program_name", [])
+
+    if not names:
+        return general_results
+
+    for name in names:
+        q_sub = Q()
```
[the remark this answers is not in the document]
gsa-jrothacker: Honestly, I'd prefer we do this now while you're already in here. Would give us one less method we need to fix.

backend/dissemination/searchlib/search_federal_program_name.py
gsa-jrothacker: I would love to get input from anyone with more UX experience, on how a user would expect the search to work. I think it's safe to assume new lines = OR, but would the person want each line to pick up anything that matches all terms, or want the term was exactly?

If I was writing the sql directly:
`select * from table where name like "%A-Plus School%"`
vs
`select * from table where name like "%A%" and name like "%Plus%" and name like "%School%"`

Thinking more on this, it's probably the "pick up anything that matches in any order" since that's what a google search would do. I'd probably only split on space?

I really don't know which is best, it was just something that popped into my head while reading the code

backend/dissemination/searchlib/search_federal_program_name.py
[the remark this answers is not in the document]
phildominguez-gsa: Done!

backend/dissemination/searchlib/search_federal_program_name.py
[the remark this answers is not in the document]
phildominguez-gsa: Done!

backend/dissemination/searchlib/search_federal_program_name.py
phildominguez-gsa: > I'd probably only split on space?
Just to continue the comparison, Google doesn't care about commas/semicolons/periods though, no? Dashes I'm less sure about.

backend/dissemination/views.py
gsa-jrothacker: This makes sense to me, I appreciate the deep dive.

backend/dissemination/searchlib/search_federal_program_name.py
gsa-jrothacker: You do bring up a good point that those delimiters are more or less ignored. I think this is fine as is, though I do think it would be worth while to re-visit if users are (or are not) using search as we expect.

gsa-jrothacker (approved):

PR #4566, opened 2024-12-26
Remove report flag for removal

sambodeme:
## Description
This PR introduces the logic change necessary to allow staff users to delete reports flagged for removal for over 6 months using the Django admin interface.

### How to Test:
1. **Switch to this PR branch**: In your local environment, check out this branch and launch the application.
2. **Create Test Data**: Create four reports:
   - Leave three reports in the "In Progress" state.
   - Complete the submission process for the fourth report.
3. **Flag Reports for Removal**: Use the delete icon to flag two "In Progress" reports for removal.
4. **Update Transition Dates**:
   - Use your database client to connect to your local database.
   - Update the `transition_date` field for one of the flagged reports:
     - Ensure the `transition_date` is at least six months prior to the current date.
5. **Run the Admin Action**:
   - Access the Django admin interface.
   - Navigate to **SF-SACs**.
   - Run the action titled `Delete reports flagged for removal for over 6 months`.
   - 
image

### Expected Outcome:
- The report in the "Submitted" state is **not removed**.
- The report in the "In Progress" state is **not removed**.
- The report flagged for removal with a `transition_date` less than six months from the current date is **not removed**.
- Only the report flagged for removal with a `transition_date` greater than six months from the current date is **removed**.

image

## PR Checklist: Submitter

## PR Checklist: Reviewer

### DIFF
--- a/backend/audit/admin.py
+++ b/backend/audit/admin.py
@@ -27,6 +27,7 @@ from audit.validators import (
 )
 from django.contrib.admin import SimpleListFilter
 from django.utils.translation import gettext_lazy as _
+from django.db import transaction
 from django.utils.timezone import now
 
 from dissemination.remove_singleauditreport_pdf import remove_singleauditreport_pdf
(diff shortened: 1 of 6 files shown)


backend/audit/admin.py
```diff
@@ -24,9 +27,73 @@
...
+            UeiValidationWaiver.objects.filter(uei=auditee_uei).delete()
+        SacValidationWaiver.objects.filter(report_id=sac).delete()
+        remove_singleauditreport_pdf(sac)
+        remove_workbook_artifacts(sac)
+        Access.objects.filter(sac=sac).delete()
+        SubmissionEvent.objects.filter(sac=sac).delete()
+        ExcelFile.objects.filter(sac=sac).delete()
+        SingleAuditReportFile.objects.filter(sac=sac).delete()
+        report_ids.append(sac.report_id)
+        sac.delete()
```
gsa-jrothacker, line 87: I am still getting caught up to speed on Django ORM...

It looks like we iterate over all the SAC records, then one by one delete entries with a foreign key relation, followed by deleting the single sac.

Does Django handle all those delete as a single transaction? If no, I am a bit concerned about failures in the middle of this process. For example, what if the SubmissionEvent fails to delete, won't that leave us in an inconsistent state?

If the database has cascading deletes on all the foreign key constraints, we could get away with just deleting all the sacs at once rather than in loop.

backend/dissemination/remove_singleauditreport_pdf.py
gsa-jrothacker, line 1: I really like small self contained files like this.

backend/audit/admin.py
sambodeme: Placing this process under a transaction is generally a good idea to avoid unexpected issues. The current implementation, even without a global transaction, will not leave the database in an inconsistent state. This is because the sac record, which acts as the parent for all the other records in the list, is only removed at the end of the process. If an unexpected issue occurs midway through the process, the sac record will remain intact. When the action is re-run, it will clean up any remaining child records before finally removing the sac. As long as the parent sac record is not removed prematurely, the child records can be handled in subsequent runs of the deletion action.
That said, I still add a global transaction as an extra layer of protection.

phildominguez-gsa:
I can't seem to get any to delete. Here is my sample set:
image
The non-disseminated ones set in 2023 should be deletable, right?

I also have a question. If an audit is created and never gets past the `in_progress` status, will it never be deletable because it has no transition dates?

sambodeme:
> I can't seem to get any to delete. Here is my sample set: image The non-disseminated ones set in 2023 should be deletable, right?
> 
> I also have a question. If an audit is created and never gets past the `in_progress` status, will it never be deletable because it has no transition dates?

You missed step 3. Reports should be flagged for removal before you can delete them. This holds for your two questions.

phildominguez-gsa (approved):

gsa-jrothacker (approved):

Issue #4561, opened 2024-12-23
Bulk CSV export


danswick:
This is a stub for now. @sambodeme and @gsa-jrothacker will flesh out the details below, but the gist is that we want some kind of basic CSV export downloads while we focus on building out better search and oversight features. It might look something like SAM.gov's ALN exports, but it could take other forms as well. 

@sambodeme's [previous notes](https://docs.google.com/document/d/1UlcJqYW5zI9V9d7nmom4R04AJnsxGc4WWG-mJMIyv3g/edit?tab=t.0 )

[Previously](https://github.com/GSA-TTS/FAC/issues/4508).

gsa-jrothacker:
Code change merged in to main. We will want to do additional testing on the exports in additional environments. As a team we should determine how often to run the export job then update the github action to run on a schedule.

jadudm:
This is now being tracked elsewhere. Some of this work completed under this ticket; closing this ticket, although it was later re-ticketed.

PR #4572, opened 2024-12-31
[FEATURE] Creating a new command to export database data to CSV.

gsa-jrothacker:
This is adding a new command which will allow us to export data from the database to S3.

To test: 
1. pull down this code, rebuild the application due to the added dependency on sling.
1. ~~Goto Minio dashboard (localhost:9002) and create a new bucket called `export`~~
1. Then you can run the new command: `docker compose run web python manage.py export_data`
1. Assuming you have data in your local database, you should see files in the ~~newly created export~~ `gsa-fac-private-s3` bucket in a new directory called `export`

## PR Checklist: Submitter

- [x]   [Link ](https://github.com/GSA-TTS/FAC/issues/4561) to an issue if possible. If there’s no issue, describe what your branch does. Even if there is an issue, a brief description in the PR is still useful.

## PR Checklist: Reviewer

### DIFF
--- a/backend/support/management/commands/export_data.py
+++ b/backend/support/management/commands/export_data.py
@@ -33,13 +33,13 @@ DEFAULT_OPTIONS = {
 
 
 class StreamGenerator:
-    PRIVATE_QUERY = (
+    EXCLUDE_NONPUBLIC_QUERY = (
         "select * from {table_name} where report_id in ("
         " select dg.report_id from public.dissemination_general dg"
         " where dg.audit_year = '{audit_year}' and dg.is_public = 'true' )"
     )
 
-    PUBLIC_QUERY = (
+    UNRESTRICTED_QUERY = (
         "select * from {table_name} where report_id in ("
         " select dg.report_id from public.dissemination_general dg"
         " where dg.audit_year = '{audit_year}')"
@@ -53,7 +53,9 @@ class StreamGenerator:
             "dissemination_" + model for model in restricted_model_names
         ]
         default_query = (
-            self.PRIVATE_QUERY if table_name in restricted_tables else self.PUBLIC_QUERY
+            self.EXCLUDE_NONPUBLIC_QUERY
+            if table_name in restricted_tables
+            else self.UNRESTRICTED_QUERY
         )
         self.query = query_override or default_query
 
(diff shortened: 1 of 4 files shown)


phildominguez-gsa:
image
Any idea why I'm getting this error?

phildominguez-gsa:
Not sure what I'm doing wrong, but I'm not seeing objects being created in my export bucket. I have one disseminated submission but it's not showing up. I created the materialized view as well (the command errors otherwise), you can see it in the advanced search results here:

image

It has an audit year of 2023 but the command doesn't pick anything up:
image

Here's the empty bucket:
image

gsa-jrothacker:
> Not sure what I'm doing wrong, but I'm not seeing objects being created in my export bucket. I have one disseminated submission but it's not showing up. I created the materialized view as well (the command errors otherwise), you can see it in the advanced search results here:
> ...
> Any ideas?

Discussed over slack, the issue was the example audit was marked as `is_public = false`, and therefore correctly excluded from the export. Will spend some time diving into which tables need the is_public flag or not and adjust the queries if needed.

backend/support/management/commands/export_data.py
```diff
@@ -0,0 +1,136 @@
...
+    PRIVATE_QUERY = (
+        "select * from {table_name} where report_id in ("
+        " select dg.report_id from public.dissemination_general dg"
+        " where dg.audit_year = '{audit_year}' and dg.is_public = 'true' )"
+    )
+
+    PUBLIC_QUERY = (
+        "select * from {table_name} where report_id in ("
+        " select dg.report_id from public.dissemination_general dg"
+        " where dg.audit_year = '{audit_year}')"
```
gsa-suk, line 45: and dg.is_public = 'true' 
belongs here.

gsa-jrothacker: This query is specifically for any table where the data is public so it does not need the `is_public` = true check. The query above, `PRIVATE_QUERY` is intended to return _only_ public data (thus hiding the private data).

I am open to renaming these variables if it makes it easier to understand. Like they say, there are only three things difficult about computer science, naming things, and off by one errors.

rnovak338 (approved):
Given the above conversation, seems to work as intended.

gsa-suk (approved):

PR #4620, opened 2025-01-14
Update versions for backup util to match v0.1.10

asteel-gsa:
Follow on PR for #4601 
This brings some of the stray versions in sync with the latest backup util version of `v0.1.10` and adds back a conditional clause for backups. We don't want to backup every env before deployment, just prod

### DIFF
--- a/.github/workflows/deploy-application.yml
+++ b/.github/workflows/deploy-application.yml
@@ -70,6 +70,7 @@ jobs:
           cf_command: update-user-provided-service fac-key-service -p '"{\"SAM_API_KEY\":\"${{ secrets.SAM_API_KEY }}\", \"DJANGO_SECRET_LOGIN_KEY\":\"${{ secrets.DJANGO_SECRET_LOGIN_KEY }}\", \"LOGIN_CLIENT_ID\":\"${{ secrets.LOGIN_CLIENT_ID }}\", \"SECRET_KEY\":\"${{ secrets.SECRET_KEY}}\"}"'
 
       - name: Backup the database
+        if: startsWith(github.ref, 'refs/tags/v1.')
         uses: cloud-gov/cg-cli-tools@main
         with:
           cf_username: ${{ secrets.CF_USERNAME }}
--- a/.github/workflows/deploy-application.yml
+++ b/.github/workflows/deploy-application.yml
@@ -76,7 +76,7 @@ jobs:
           cf_password: ${{ secrets.CF_PASSWORD }}
           cf_org: gsa-tts-oros-fac
           cf_space: ${{ env.space }}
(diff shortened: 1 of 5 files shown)


gsa-suk (approved):

.github/workflows/deploy-application.yml
```diff
@@ -70,13 +70,14 @@ jobs:
           cf_command: update-user-provided-service fac-key-service -p '"{\"SAM_API_KEY\":\"${{ secrets.SAM_API_KEY }}\", \"DJANGO_SECRET_LOGIN_KEY\":\"${{ secrets.DJANGO_SECRET_LOGIN_KEY }}\", \"LOGIN_CLIENT_ID\":\"${{ secrets.LOGIN_CLIENT_ID }}\", \"SECRET_KEY\":\"${{ secrets.SECRET_KEY}}\"}"'
 
       - name: Backup the database
+        if: startsWith(github.ref, 'refs/tags/v1.')
```
gsa-jrothacker, line 73: When we were dealing with the UEI issue, I recall us going back and forth on this line. I've only scratched the surface of github actions. Can you help me understand what our goal is with this, how this line achieves it?

asteel-gsa: @gsa-jrothacker It is a conditional clause to exclusively perform a backup operation exclusively on releases, as they are tags, with a `v1.` notation. More information on this specific clause can be found [here](https://stackoverflow.com/questions/58475748/github-actions-how-to-check-if-current-push-has-new-tag-is-new-release).

In [this pr](https://github.com/GSA-TTS/FAC/commit/ed9369f64350b828bd494c19ddb332e72cf2545d), the removal of this line actually was, unfortunately, not what y'all were expecting, based on what little I was told, as removing this enabled backup on **all** environments, rather than prod exclusively.

We don't want to, nor have a necessity, to backup the data in lower environments, and the 2 hour backup cron is substantial enough for us in the lower environments. We do this exclusively in production, as if a breaking change occurs on the "deploy" we want the last copy of the data to have some degree of integrity that if we need to rollback, the pre-deploy backup would be the last known "clean copy" of the data.

asteel-gsa: To have disabled the backup before deployment in production, this would have been acceptable
```yaml
      # - name: Backup the database
        # if: startsWith(github.ref, 'refs/tags/v1.')
        # uses: cloud-gov/cg-cli-tools@main
        # with:
          # cf_username: ${{ secrets.CF_USERNAME }}
          # cf_password: ${{ secrets.CF_PASSWORD }}
          # cf_org: gsa-tts-oros-fac
          # cf_space: ${{ env.space }}
          # command: cf run-task gsa-fac -k 7G -m 3G --name deploy_backup --command "./fac-backup-util.sh v0.1.10 deploy_backup"
```

Issue #4574, opened 2025-01-02
Add user session check redirection to login page


anagradova:
# At a glance

Currently there is no clear way in the platform for a user to know their session has been terminated without getting a 403 or other error when attempting an action.

"**In order to**: create a better user experience and reduce confusion
**as a**: user
**I want**: to be redirected to the login page so I don't attempt something I can't do

# Acceptance Criteria

**Given**: a user logged into the system
**when:** the user interacts with the system
**and**: the user's session has expired
**then**: the user will be redirected to the login page

# Background

This need was shown by a customer incident #[5879] and was discussed as a good improvement to the overall function of the application with @sambodeme.

# Security Considerations

We will need to investigate the session handling such that redirecting to the login page does not create an security issues.

  
# Sketch

- [ ] Investigate current session handling
- [ ] Review the checks associated with sessions and permissions
- [ ] Implement a function to redirect user to login screen

# Definition of Done

- [ ] Expired sessions or invalid sessions result in redirection to the login page

## Engineering Backlog

- [ ] Has clearly stated/testable acceptance criteria
- [ ] Has a sketch or list of tasks
- [ ] Can reasonably be done in a few days (otherwise, split this up!)

## Engineering Available

- [ ] There's capacity in the `In Progress` column
- [ ] An engineering shepherd has been identified

## Engineering In Progress

- [ ] Meets acceptance criteria
- [ ] Meets [QASP conditions](https://derisking-guide.18f.gov/qasp/)

## Engineering Done

- [ ] Presented in a sprint review
- [ ] Includes screenshots or references to artifacts
- [ ] Tagged with the sprint where it was finished
- [ ] Archived

</details>

analynd:
@anagradova For your reference, here's the finalized design that was shown during today's Design Sync with screenshots attached 😀: https://www.figma.com/design/fkjGQbAPlnPRIyNfY3Gt1W/%234574-Add-user-session-check-redirection-to-login-page?node-id=7-3640&t=56JrUmqrtoIFWfHR-4 

![image](https://github.com/user-attachments/assets/2a387d02-12f6-436e-897b-11cbf764595e)
![image](https://github.com/user-attachments/assets/de237775-56ae-4917-bb93-73abf2dd3981)


Please let me know if you have any questions. Thank you for your patience and collaboration! 🥳

PR #4636, opened 2025-01-16
4574 add session check redirection

anagradova:
[Link](https://github.com/GSA-TTS/FAC/issues/4574) to issue #4574 

This feature/fix is designed to solve a usability issue that certain users were experiencing as referenced in this [ticket](https://fac-gov.zendesk.com/agent/tickets/5879)

To test this feature, the user should begin an audit submission and continue to the upload SAC step. Wait for the session to expire with your current login and then attempt to upload the file. 
image

Upon submission, with an expired session, the user should be presented with this screen allowing them to log back in or proceed to the home page.

image

If the user goes back through the login cycle, the callback URL from login.gov should place them back on the upload submission step where their session expired. 

I have created two modals, one for session expiration (session-expired-modal) and one for session timeout warning (session-warning-modal). Each modal has a javascript snippet with it to trigger the modal due to the USWDS framework design requiring a <a> or <button> tag attribute and action to trigger the modal. The javascript snippet creates and simulates a click on an <a> tag for the modal. Due to the nature of how we manage sessions and interaction, the warning modal should never trigger, but I have left the code in place in the nav_primary.html template should we wish to use it in the future. 

The session expiration flag is set in mixin.py during the auth check, but could be triggered by anything that can pass the trigger to the template context.  

NOTE: A quick way to trigger this for testing is to login, reach the upload step, open a new browser window from the site, logout, and then attempt to upload the file. This should trigger the modal.

## PR Checklist: Reviewer


backend/audit/mixins.py
```diff
@@ -22,12 +24,10 @@ def __init__(self, message, eligible_users):
...
 
 def check_authenticated(request):
-    if not hasattr(request, "user"):
-        raise PermissionDenied(PERMISSION_DENIED_MESSAGE)
-    if not request.user:
+    if not hasattr(request, "user") or not request.user:
         raise PermissionDenied(PERMISSION_DENIED_MESSAGE)
     if not request.user.is_authenticated:
-        raise PermissionDenied(PERMISSION_DENIED_MESSAGE)
+        return {"session_expired": True}
```
gsa-jrothacker, line 30: Does it make sense to instead of returning an object like this, instead create a new exception `SessionExpired`?

backend/templates/includes/nav_primary.html
```diff
@@ -235,5 +235,124 @@ <h2 class="usa-modal__heading" id="login-modal-heading">You must log in to conti
                 {% endif %}
             </div>
         </div>
+        {% if session_expired %}
+        <script>
```
gsa-jrothacker, line 239: i would suggest moving all the modal html and js into it's own separate file.

backend/audit/mixins.py
anagradova: We could create a new exception, but passing the object into the context allows us to use it in a template and then conditionally trigger a code block for display. This would also allow us to carry more information in the object from the triggering functions in the future should we want to to make error modals more descriptive.

gsa-jrothacker (dismissed):
Thanks for the updates! I think this will be extensible to any additional exceptions that happen.

backend/templates/includes/nav_primary.html
anagradova: Done.

backend/audit/mixins.py
anagradova: Done.

backend/audit/mixins.py
```diff
@@ -22,12 +24,10 @@ def __init__(self, message, eligible_users):
 
 
 def check_authenticated(request):
-    if not hasattr(request, "user"):
-        raise PermissionDenied(PERMISSION_DENIED_MESSAGE)
-    if not request.user:
+    if not hasattr(request, "user") or not request.user:
```
gsa-jrothacker, line 27: Thank you.

backend/audit/test_manage_submission_view.py
```diff
@@ -82,7 +82,8 @@ def test_login_required(self):
             )
         )
 
-        self.assertEqual(response.status_code, 403)
+        self.assertTemplateUsed(response, "home.html")
+        self.assertTrue(response.context["session_expired"])
```
gsa-jrothacker, line 86: I think this is going to be a much better experience, the user will now see a nice page instead of the 403. Great work.

backend/audit/test_mixins.py
```diff
@@ -217,7 +218,7 @@ def test_anonymous_raises(self):
 
         view = self.ViewStub()
         self.assertRaises(
-            PermissionDenied, view.dispatch, request, report_id="not-logged-in"
+            SessionExpiredException, view.dispatch, request, report_id="not-logged-in"
```
gsa-jrothacker, line 221: I was worried we didn't have any remaining tests for PermissionDenied... but then realized I forgot to expand the diff. Again, I think giving more specific exceptions between a Permissions problem and a Session Expired is going be nice.

backend/config/context_processors.py
```diff
@@ -59,3 +60,9 @@ def maintenance_banner(request):
 
     # Base case - we are not within any of the given timeframes. Disable the banner.
     return context
+
+
+# Helper function to format time
+def format_time(seconds):
+    minutes, seconds = divmod(seconds, 60)
+    return f"{minutes} minutes, {seconds} seconds"
```
gsa-jrothacker, line 68: I didn't see this used anywhere else (I did a cmd+f to verify).

gsa-jrothacker (dismissed):
Looks good to me, all my comments were random thoughts I had while reviewing and no changes requested. Thanks!

backend/config/context_processors.py
anagradova: Removed.

gsa-jrothacker (approved):

PR #4651, opened 2025-01-22
Adding --wait to all cf run-task commands

gsa-jrothacker:
## Purpose

We recently found out that backup tasks were failing silently. Backup tasks are kicked off by github actions through a cloud foundry run-task command. By default this task is run asynchronously, and therefore the github action sees a success message when the task is started successfully. The purpose of this code change is ensure we can see the errors in the github workflow.

## How  

The cloud foundry run-task command accepts a --wait flag which will wait until the task is finished. If the task fails, the github workflow will fail.

## Testing

I created a temporary script which always fails. (Since removed from the commit). I tested it using the deploy to preview workflow. You can see that here: https://github.com/GSA-TTS/FAC/actions/runs/12913210660/job/36010804577


asteel-gsa (approved):
This looks good to me: https://github.com/GSA-TTS/FAC/actions/runs/12936852545/job/36083408728
image
image

Issue #4531, opened 2024-12-11
Revise layout + copy for Basic and Advanced search screens


Leighdiddy:
We are currently showing too much copy in the info alert on our Search and Advanced Search page layouts, which was raised as an issue in our recent EOP review. We need to revise the [Search](https://app.fac.gov/dissemination/search/) and [Advanced Search ](https://app.fac.gov/dissemination/search/advanced/) screens to remove the "Summary reports" and "Search" or "Advanced Search" sections of copy from the blue info alert. The info alert should only contain "Searching the FAC database" copy. 

The "Search" and "Advanced Search" sections of copy should be combined and revised to say: "Basic search is updated in real time and can be used to confirm that an audit submission has been accepted. Advanced Search is updated once daily at 12am EST and includes additional filters to support Federal audit resolution and oversight." 

The headlines on Basic and Advanced search screens should be updated to reflect which type of search you are currently viewing: "Audit submissions - Advanced search" or "Audit submissions - Basic search"  The info alert and additional search copy should be identical on both pages, containing links to their respective pages (basic vs. advanced.) 

<img width="1336" alt="Screenshot 2024-12-11 at 12 29 46 PM" src="https://github.com/user-attachments/assets/e7a36ca6-712e-47e4-b6bc-e568b890e377" />

### Tasks

```[tasklist]
- [x] Revise search screen layout for **Basic** search with updated headline and revised "Basic vs. Advanced" copy shown outside of the info alert.
- [x] Revise search screen layout for **Advanced** search with updated headline and revised "Basic vs. Advanced" copy shown outside of the info alert.
- [ ] Review with design teammate (can share screen grabs in ticket for async review)
- [ ] Implement change on frontend
```

Leighdiddy:
In progress: https://www.figma.com/design/D5LZmSTi50vnAoyVxCZSMM/Revise-layout-%2B-copy-for%234531--%234531-Basic-and-Advanced-search-screens%C2%A0?node-id=1-5520&node-type=frame&t=9iqjecCRgamWyuEM-0

anagradova:
@Leighdiddy Please see the adjusted screens for approval. 

<img width="873" alt="Image" src="https://github.com/user-attachments/assets/6b00e646-357e-4a1c-a2a5-97e28da4eee9" />

<img width="851" alt="Image" src="https://github.com/user-attachments/assets/09397d74-cda1-461c-a47e-8f190b31bb66" />

[...]

PR #4671, opened 2025-01-28
4531 revise layout copy for search screens

anagradova:
PR updates the search info text box on ticket #4531 . This can be seen when going to the search page (/dissemination/search/).

image

image


backend/dissemination/templates/search-alert-info.html
```diff
@@ -1,34 +1,12 @@
...
-        <p class="usa-alert__text padding-bottom-2">
-            For search results of {{ summary_report_download_limit|intcomma }} submissions or less, you can download a combined spreadsheet of all data.
-            If you need to download more than {{ summary_report_download_limit|intcomma }} submissions, try limiting your search parameters to download in batches.
+<div class="margin-y-2 display-flex">
+    <div class="text-family-public-sans">
+        <p class="usa-alert__text">
+          <a href="{% url 'dissemination:Search' %}">Basic Search</a> is updated in real time and can be used to confirm that an audit submission has been accepted. <a href="{% url 'dissemination:AdvancedSearch' %}">Advanced Search</a> is updated once daily at 12am EST and includes additional filters to support our Federal partners.
+        </p>
+        <p class="usa-alert__text">
+          Learn how Basic and Advanced search filters work on our <a href='https://www.fac.gov/search-resources/filters/' target='_blank'>Search resources page</a>.
```
phildominguez-gsa, line 9: Not sure why these didn't alert before, but I think we're supposed to use `rel="noopener noreferrer"` when using `target='_blank'`, otherwise the security scanner complains. See [here](https://github.com/GSA-TTS/FAC/pull/4007) for an example PR.

phildominguez-gsa (changes requested):

backend/dissemination/templates/search-alert-info.html
anagradova: Added

phildominguez-gsa (approved):

PR #4673, opened 2025-01-29
[Draft] Adds Cypress to Pull Request Checks

asteel-gsa:
Uses cypress to run full e2e suite against local build stack


.github/workflows/cypress-testing-from-container.yml
```diff
@@ -0,0 +1,89 @@
...
+      CYPRESS_LOGIN_TEST_PASSWORD_AUDITEE: ${{ secrets.CYPRESS_LOGIN_TEST_PASSWORD_AUDITEE }}
+      CYPRESS_LOGIN_TEST_OTP_SECRET_AUDITEE: ${{ secrets.CYPRESS_LOGIN_TEST_OTP_SECRET_AUDITEE }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
jperson1, line 46: Out of curiosity, how come v20 rather than the 18 we use elsewhere? Or, the most recents (22/23)?

jperson1 (approved):
11 minute Cypress testing on PRs is crazy work, very excited to mess around with them more

.github/workflows/cypress-testing-from-container.yml
asteel-gsa: Oh, I just grabbed this from [here](https://github.com/GSA-TTS/FAC/blob/b07382091435fed9ced2a5e533cdaee34f352a16/.github/workflows/testing-from-build.yml#L81-L84) I didn't realize we had mismatched versions.

Looks like we have 18 [here](https://github.com/GSA-TTS/FAC/blob/b07382091435fed9ced2a5e533cdaee34f352a16/.github/workflows/testing-from-ghcr.yml#L27) and [here](https://github.com/GSA-TTS/FAC/blob/b07382091435fed9ced2a5e533cdaee34f352a16/.github/workflows/testing-from-build.yml#L26) for django, but 20 [here](https://github.com/GSA-TTS/FAC/blob/b07382091435fed9ced2a5e533cdaee34f352a16/.github/workflows/testing-from-build.yml#L84) and [here](https://github.com/GSA-TTS/FAC/blob/b07382091435fed9ced2a5e533cdaee34f352a16/.github/workflows/testing-from-ghcr.yml#L86) for a11y

asteel-gsa: Looks like 18 & 20 are the major supported versions for the [`setup-node` action](https://github.com/marketplace/actions/setup-node-js-environment#supported-version-syntax)

jperson1: Makes sense to me. We could bump the others, but I seem to remember a headache or two when we upgraded from 16. If it works it works!

asteel-gsa: So I think for this, let's merge as is, and I will make a follow on PR that attempts to put that all at current stable, and see what happens

PR #4675, opened 2025-01-29
Bump setup-node versions to 23

asteel-gsa:
Follow on from addition to cypress to PR checks, we are running 2 different versions of node on the setup steps in the actions, and we would like to test to see if v23 will cause our actions to break.

To Test:
Pull down the branch, `cd backend` & `npm install`. Rebuild the container with `make docker-first-run`, then `docker compose -f docker-compose.yml up`. In a new terminal, `npx cypress open` and run `full-submission.cy.js` spec.
image

Alternatively, Cypress tests on the PR should pass.

### DIFF
--- a/.github/workflows/cypress-testing-from-container.yml
+++ b/.github/workflows/cypress-testing-from-container.yml
@@ -43,7 +43,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 23
 
       - name: Install Node dependencies
         working-directory: ./backend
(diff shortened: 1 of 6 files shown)


asteel-gsa:
@jperson1 @phildominguez-gsa leaving this for y'all to approve to deny. not seeing anything catastrophic with the versions across the board, and standardizes us to be running node 23 on all actions and dockerfile, so if you have any concerns thatwe shouldn't take this, we can close it down and ignore.

phildominguez-gsa (approved):

PR #4678, opened 2025-01-30
loosen the check to allow terraform plans to rerun on prs if failing to deploy

asteel-gsa:
Remove the conditional to exclude terraform plans


jperson1 (approved):
Thanks!

PR #4692, opened 2025-02-07
Terraform module updates

asteel-gsa:
This PR encapsulates the necessary changes to upgrade our modules from `v1.1.0` to `v2.1.0` in `sandbox`. The process is detailed in the new doc `docs/infra/terraform-module-upgrade.md` 

This will require 1-2 members of the team to do a call and walkthrough exactly what is happening behind the scenes. Once we move this into `main`, we will need to repeat these steps in `preview` on a call and ensure everything works the same way. Because we have to `import` the resource, I believe each environment will need changes to the `main.tf` file, and it is recommended that we proceed carefully.

My hunch at this time, is to perform these steps locally, run the apply locally, and then open a PR to keep things in sync.

For the new network policy for `app -> clamav`, I believe we will need to add this into the terraform for non-sandbox envs:
```tf
data "cloudfoundry_app" "fac-app" {
    name_or_id = "gsa-fac"
    space      = data.cloudfoundry_space.space.id
}

resource "cloudfoundry_network_policy" "app-network-policy" {
  provider = cloudfoundry-community
  policy {
    source_app      = data.cloudfoundy_app.fac-app.id
    destination_app = module.clamav.app_id
    port            = "61443"
    protocol        = "tcp"
  }
}
```


asteel-gsa:
Currently, we are expecting the meta module to have all of these changes listed in the terraform plan, however, this will not actually be the case.
```sh
[10:06 AM][asteel][meta]$ aws s3 ls s3://$BUCKET/
2025-02-14 10:06:59      73829 terraform.tfstate.meta
2025-02-14 10:06:49      76320 terraform.tfstate.meta.newprovider
```

I have made a backup of the terraform plan that is illustrated here under `.newprovider`, simply as a means to regenerate the provider files for use with the other environments while I begin prepping them for the upgrade. 

As part of the operation, we will need to rerun a `terraform plan` and `terraform apply` in `meta`, on my machine to update the `.tfstate` before merging this, and that will allow subsequent runs to generate the plan successfully. I have not updated the live state, as a means to not cause meta plans to show incorrectly on other PRs.

jadudm (approved):
We have tested this code against all lower environments, and carried out the manual update to each env (`sandbox`, `preview`, `dev`, `staging`, and `production`). This is documented as part of the commit. This is the final step: merging this into `main`, and then testing a deploy against each environment.

PR #4715, opened 2025-02-25
Adding new models for audit source of truth

gsa-jrothacker:
# Purpose

This adds new models for the source of truth: audit and history tables.

# Testing

Rebuilt the application locally including removing and recreating the database. The migration completed successfully.

# References

[Source of Truth Documentation](https://drive.google.com/drive/folders/1u-lqfgYEUKtRACSfl9VMx1P5aAtivPYP)

# Notes

1. I opted to remove the schema table. It was not being used in the POC and the existing validations in the excel templates seemed sufficient for our needs.


backend/audit/models/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    if int(month) not in range(1, 13):
+        raise Exception("Unexpected month value for report_id")
+    separator = "-"
+    report_id = separator.join([year, month, source, str(count).zfill(10)])
+    return report_id
+
+
+class JsonArrayToTextArray(Func):
+    function = "json_array_to_text_array"
+    output_field = ArrayField(models.CharField())
```
rnovak338, line 37: I noticed these (or most of these) are all duplicates of existing methods/variables - I assume the plan is to clean out the original copies around the same time we clean out SAC-related syntax?

backend/audit/models/__init__.py
```diff
@@ -19,14 +19,18 @@
...
 from .submission_event import SubmissionEvent
 
 # In case we want to iterate through all the models for some reason:
 _models = [
     Access,
+    Audit,
     DeletedAccess,
     ExcelFile,
     GeneralInformationMixin,
+    History,
```
rnovak338, line 33: Are we going to drop the original `curation_tracking` logic for tracking historical changes to intake?

backend/audit/models/utils.py
gsa-jrothacker: Yes, that is the plan. I thought I would use this opportunity to clean up the code base more so we don't have so many _huge_ files.

backend/audit/models/__init__.py
gsa-jrothacker: Yes and No. I documented this in the status google doc. Basically, I wasn't able to verify we were actually using the curation in production (I didn't see it enabled in staging which may be intentional). I figured it's easier to remove this table after the fact if we want to use the curation tables only. However adding it later is slightly more difficult (not hard, just requires an additional step).

gsa-suk: Curation work for historic data is pending, waiting on the sot implementation.

PR #4716, opened 2025-02-25
Converting SAC to Audit for logic pertaining to Access model

rnovak338:
- Re-arranged `__init__.py` in models (circular import issue with Access/Audit models.
- New migration to apply "audit" FK to Access model.
- Application of audit field to places in the code where "SAC" is referenced for an Access object.


backend/api/serializers.py
```diff
@@ -282,19 +282,19 @@ class AccessListSerializer(serializers.ModelSerializer):
     submission_status = serializers.SerializerMethodField()
 
     def get_auditee_uei(self, access):
-        return access.sac.auditee_uei
+        return access.audit.auditee_uei
```
gsa-jrothacker, line 285: I think we're going to want to also check the sac as well. Since a lot of the audits will be null on first push. 

Note: 
(i) We'll want to make the same change throughout this file
 (ii) We'll want to update tests to validate this as well.
 

```suggestion
        return access.audit.auditee_uei if access.audit else access.sac.auditee_uei
```

backend/api/views.py
```diff
@@ -138,23 +141,23 @@ def access_and_submission_check(user, data):
 
         # Create all contact Access objects
         Access.objects.create(
-            sac=sac,
+            audit=audit,
```
gsa-jrothacker, line 144: We will also need to keep the `sac=sac` until the migration is complete.

backend/audit/fixtures/single_audit_checklist.py
```diff
@@ -120,34 +120,33 @@ def fake_auditee_certification():
 
 def _create_sac(user, auditee_name, submission_status="in_progress"):
```
gsa-jrothacker, line 121: We will have to think through how this works with both an audit and SAC

backend/audit/views/manage_submission_access.py
gsa-jrothacker, line 1: Similar to previous comments, we'll need to account for the transition phase when we're saving/creating both audits and sacs.

gsa-jrothacker (changes requested):

backend/api/views.py
rnovak338: Re-introduced `sac` in user-facing logic.

backend/api/serializers.py
```diff
@@ -282,19 +282,21 @@ class AccessListSerializer(serializers.ModelSerializer):
     submission_status = serializers.SerializerMethodField()
 
     def get_auditee_uei(self, access):
-        return access.sac.auditee_uei
+        return access.audit.audit["general_information"]["auditee_uei"] if access.audit else access.sac.auditee_uei
```
gsa-jrothacker, line 285: we may want to do something like:
`access.audit.audit.get("general_information", {}).get("auditee_uei", None)` just in case the user hasn't inputed the general information yet (or we didn't save it properly). I don't think this _should_ be possible, but better to prevent errors.

backend/api/serializers.py
```diff
@@ -282,19 +282,21 @@ class AccessListSerializer(serializers.ModelSerializer):
...
 
     def get_auditee_fiscal_period_end(self, access):
-        return access.sac.auditee_fiscal_period_end
+        return access.audit.audit["general_information"]["auditee_fiscal_period_end"] if access.audit else access.sac.auditee_fiscal_period_end
 
     def get_auditee_name(self, access):
+        # TODO: 2/25 access audit
+        # Where is the equivalent of "auditee_name" in the Audit model?
```
gsa-jrothacker, line 292: I _think_ this would `access.audit.audit.get("general_information", {}).get("auditee_name", None)`

backend/audit/views/manage_submission_access.py
```diff
@@ -59,17 +60,22 @@ def get(self, request, *args, **kwargs):
...
+    # TODO: 2/25 access audit
+    # We will want to replace "sac" with "audit" under this endpoint when we are ready to deprecate SAC.
     def post(self, request, *args, **kwargs):
         """
         Change the current auditor certifying official and redirect to submission
         progress.
         """
         report_id = kwargs["report_id"]
         sac = SingleAuditChecklist.objects.get(report_id=report_id)
+        audit = Audit.objects.get(report_id=report_id)
```
gsa-jrothacker, line 78: We will want wrap any `.get` with a try/except block in case the audit hasn't been migrated yet.

backend/audit/views/manage_submission_access.py
```diff
@@ -117,8 +123,8 @@ def post(self, request, *args, **kwargs):
             context = {
                 "role": self.role,
                 "friendly_role": _get_friendly_role(self.role),
-                "auditee_uei": sac.general_information["auditee_uei"],
-                "auditee_name": sac.general_information.get("auditee_name"),
+                "auditee_uei": audit.general_information["auditee_uei"],
+                "auditee_name": audit.general_information.get("auditee_name"),
```
gsa-jrothacker, line 127: Shouldn't this be `audit.audit.get("general_information")` or `audit.audit["general_information"]`

backend/api/serializers.py
rnovak338: Yep, I realize the "general_information" updates further after completing more of the steps locally. I wasn't sure where this was coming from at first. This will get resolved on next push to the branch.

gsa-jrothacker (approved):

PR #4720, opened 2025-02-26
Adding the converters audit model

gsa-jrothacker:
## Purpose
Allow us to transform the uploaded excel files into the audit object.

## How
Add several new methods to existing intake library

## Testing
linting, end to end tests.

## Notes
These are not currently used, so there should be zero impact. The next PR as part of the Source of Truth will start using these and include creation/editing of Audits along with SACs.


rnovak338 (approved):
This LGTM!

PR #4722, opened 2025-02-27
Sot audit update

anagradova:
Fixed audit workflow UI issues.

### DIFF
--- a/backend/audit/views/views.py
+++ b/backend/audit/views/views.py
@@ -24,6 +24,7 @@ from audit.mixins import (
     CertifyingAuditeeRequiredMixin,
     CertifyingAuditorRequiredMixin,
     SingleAuditChecklistAccessRequiredMixin,
+
 )
 from audit.models import (
     Access,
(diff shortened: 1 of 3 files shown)


backend/audit/views/views.py
```diff
@@ -214,18 +214,18 @@ def post(self, request, *_args, **kwargs):
 
             form_section = kwargs["form_section"]
 
-            sac = SingleAuditChecklist.objects.get(report_id=report_id)
+            sac = Audit.objects.get(report_id=report_id)
```
rnovak338, line 217: We may want to leave SAC code here for now. Keeping it around will help us for the transitory period when we deploy everything and there are some prior SingleAuditChecklists that were created without an Audit.

I.E., wrapping the Audit up in a nested try/catch rather than replacing SAC... in case Audit is null or does not exist.

rnovak338 (changes requested):
I have a minor nitpick based on some feedback I had gotten from Jason with my own branch earlier this week.

rnovak338 (approved):

PR #4737, opened 2025-03-03
Creating/Saving Audits

gsa-jrothacker:
## Purpose

This PR will begin creating and updating Audits in the new Audit table.

## How

Audit should now be wired into everything all the way through to dissemination. Search and all UIs are still reliant on SAC/dissemination tables.

## Testing
Linting, Unit Tests, End to End tests

## Notes

I truly apologize about the size of this PR.
There is a large refactor as part of this change.
I will want to do additional thorough manual testing before we move this to production.
Due to the nature of this change, I'm going to request 2 approvals before I merge.


backend/audit/views/excel_file_handler.py
```diff
@@ -0,0 +1,175 @@
...
+
+        except SingleAuditChecklist.DoesNotExist as err:
+            logger.warning("no SingleAuditChecklist found with report ID %s", report_id)
+            raise PermissionDenied() from err
+        except ValidationError as err:
+            # The good error, where bad rows/columns are sent back in the request.
+            # These come back as tuples:
+            # [(col1, row1, field1, link1, help-text1), (col2, row2, ...), ...]
+            logger.warning("%s Excel upload failed validation: %s", form_section, err)
+            return JsonResponse({"errors": list(err), "type": "error_row"}, status=400)
```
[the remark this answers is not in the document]
gsa-jrothacker: This is not leaking stack information. This error list contains all the validation errors which gets displayed in a human readable format to the user on submitting the excel file.

This was in place already, and moved as part of the larger refactor.

backend/audit/models/viewflow.py
```diff
@@ -282,3 +372,141 @@ def transition_to_in_progress(self):
...
+    )
+    def transition_to_auditor_certified(self):
+        """
+        The permission checks verifying that the user attempting to do this has
+        the appropriate privileges will be done at the view level.
+        """
+        pass
+
+    @state.transition(
+        source=[STATUS.AUDITOR_CERTIFIED, STATUS.SUBMITTED],
```
phildominguez-gsa, line 456: Does this mean it's possible to go to back to "auditee certified" after already being submitted? That seems odd.

phildominguez-gsa (changes requested):

backend/audit/views/excel_file_handler.py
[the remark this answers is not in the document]
anagradova: It would seem most of these all scanning errors are heuristics-based rules that are just "keying" off the fact that it's dynamic instead of looking for the context of what the error could be. I think we could probably refactor this in the future, as you mentioned, but it should not impact the system now.

backend/audit/models/viewflow.py
gsa-jrothacker: This was copied from the existing viewflow for sacs.

There is comment [here ](https://github.com/GSA-TTS/FAC/blob/db3ad95c72faafc6a6fab8fd789bd0cd47c7595a/backend/audit/models/viewflow.py#L32-L37) which references that this is allowed through the management console in order for users to go back and fix "stuff" after it's been submitted

phildominguez-gsa (dismissed):

backend/dissemination/management/commands/migrate_audits.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+        DeletedAccess.objects.filter(sac__report_id=report_id).update(
+            audit=audit
+        )
+    except Audit.DoesNotExist:
+        logger.error(f"Unable to find an audit with report_id: {report_id}")
+
+
+# Copied from backend/audit/views/views.py:806
+def _convert_additional_fields(audit, sac):
```
gsa-jrothacker, line 97: If we're going to use this branch, you can just import `generate_audit_indexes` from audit\models\utils

backend/dissemination/management/commands/migrate_audits.py
```diff
@@ -0,0 +1,226 @@
...
+from audit.intakelib.mapping_secondary_auditors import secondary_auditors_audit_view
+from audit.models import Access, DeletedAccess, Audit, User, Schema, SingleAuditReportFile, SingleAuditChecklist
+from audit.models.constants import STATUS
+from audit.views.views import _index_awards, _index_findings, _index_general
+
+logger = logging.getLogger(__name__)
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **kwargs):
```
gsa-jrothacker, line 24: I had a couple of thoughts on this. Feel free to incorporate, or ignore however you feel is best. I trust you.
1. It would be cool if this script took in a parameter to migrate just disseminated vs in progress. Part of the launch plan is to migrate just disseminated first. It would prevent us from doing a code change.
2. It would be cool if we only had to run this once, instead of several times. My thought was a do...while type of loop where as long as there are more records to migrate, keep pulling them. If we can do that, instead of pulling 50k at time, we could pull something like 100 and it may be a bit smoother.

rnovak338: 1. Sure! I think for us to implement that I would just need to add a parameter to the command, and modify the SAC SELECT query accordingly. Simple enough
2. I have a couple questions on this. Why do we have a `limit 50000` in the first place instead of just pulling all unmigrated SAC data? And what would a do/while accomplish versus a "get all unmigrated SACs and then execute"

backend/dissemination/management/commands/migrate_audits.py
rnovak338: You got it!

backend/dissemination/management/commands/migrate_audits.py
gsa-jrothacker: The limit was there because we ran out of space when trying query _all_. 50k was the first round number I found that didn't fail because of running out of memory.

backend/dissemination/management/commands/migrate_audits.py
```diff
@@ -0,0 +1,226 @@
...
+    @staticmethod
+    def _migrate_sac(sac: SingleAuditChecklist):
+        audit_data = dict()
+        for idx, handler in enumerate(SAC_HANDLERS):
+            audit_data.update(handler(sac))
+
+        _convert_additional_fields(audit_data, sac)
+
+        # update existing Audit.
+        if Audit.objects.filter(report_id=sac.report_id).exists():
```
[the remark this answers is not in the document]
gsa-jrothacker: One reason, I wouldn't want to update a record that is already up to date. An example where this might happen, is in progress audits that haven't been migrated yet, but were created during the phase of launch where we are monitoring.

I'm not even convinced I'm right, just something I want us to think about.

PR #4766, opened 2025-03-11
First start at cross validation

gsa-jrothacker:
## Purpose

Add in the same logic for validations as in the SAC for the new SOT

## How

Translated the Audit object into the same shape as the sac validation object. Searched for all usage of the shape and added logic for validating the Audit.

## Testing Completed

1. Linting
2. Unit Tests
3. End to End Tests.

### DIFF
--- a/backend/audit/cross_validation/audit_validation_shape.py
+++ b/backend/audit/cross_validation/audit_validation_shape.py
@@ -1,0 +1,111 @@
+import logging
+
+from audit.cross_validation.naming import NC, SECTION_NAMES
+
+at_root_sections = (
+    NC.AUDIT_INFORMATION,
+    NC.GENERAL_INFORMATION,
+    NC.TRIBAL_DATA_CONSENT,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def audit_validation_shape(audit):
+    """
+    Takes an instance of Audit and converts it to the shape
+    expected by the validation functions.
+
+    This function exists so that as either the Audit or the
+    validation shape changes we only have to make adjustments in one place.
+
+    The sections that have spreadsheet workbooks all have root-level properties that
+    are the camel-case names of those sections. This function eliminates these names
+    and moves the actual values to the top level as part of returning a structure
+    that's appropriate for passing to the validation functions.
+
+    For example, if the Audit Information and Notes to SEFA sections have content,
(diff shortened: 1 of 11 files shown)


backend/audit/cross_validation/audit_validation_shape.py
```diff
@@ -0,0 +1,111 @@
+import logging
+
+from audit.cross_validation.naming import NC, SECTION_NAMES
+
+at_root_sections = (
+    NC.AUDIT_INFORMATION,
+    NC.GENERAL_INFORMATION,
+    NC.TRIBAL_DATA_CONSENT,
+)
```
jperson1, line 9: I suggest we add a ` # type: ignore` here, as we do here:

https://github.com/GSA-TTS/FAC/blob/3d7a87dd027ca498e04dafbaf0b1a66792da7d5a/backend/audit/cross_validation/sac_validation_shape.py#L8

Otherwise, I think mypy will complain.

jperson1 (approved):
Fantastic! Just a very minor suggestion, but everything is working as expected on my end.

backend/audit/cross_validation/audit_validation_shape.py
gsa-jrothacker: Thanks for the suggestion, added in.

backend/audit/views/cross_validation.py
```diff
@@ -57,3 +57,15 @@ def post(self, request, *args, **kwargs):
...
+# TODO: Post SOT Launch: Deletes
+def _compare_errors(sac_errors, audit_errors):
+    if (
+        (sac_errors and not audit_errors)
+        or (audit_errors and not sac_errors)
+        or (sac_errors) != set(audit_errors)
+    ):
+        logger.error(
+            f"<SOT ERROR> Cross Validation Errors do not match: SAC {sac_errors}, Audit {audit_errors}"
+        )
```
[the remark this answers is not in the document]
gsa-jrothacker: Let me take a closer look at what we can do here. I want to ensure we don't break the submitted_by for migrated audits. I was trying to avoid adding a new field, but we may need to.

jadudm (approved):
Reviewed with HDMS, RN, JR. Looks good for merging into the long-lived branch and continued effort.

PR #4774, opened 2025-03-12
Updating the excel file handling to save records into the history table.

gsa-jrothacker:
## Purpose

1. Deletes were not updating the Audit table... They should.
2. Deletes were not adding an entry to History Table... They should

## How

Updated the Excel Handlers to utilize the audit as well as the sac.

## Testing Done

Linting, End to End Tests, Manual Test of deleting additional UEIs and checking the database.

## Notes

1. I only tested one of the spreadsheets, but it should work for all
2. I did some more refactoring of the views in submissions, just because I like smaller files.


backend/audit/migrations/0020_merge_20250311_1702.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("audit", "0019_audit_auditee_name"),
+        ("audit", "0019_singleauditchecklist_migrated_to_audit"),
+    ]
+
+    operations = []
```
jperson1, line 13: Is this migration empty on purpose? Just making sure.

gsa-jrothacker: Django created it because there were two with the same number 019.

Before we merge into main, I hope merge all the migrations into a single file.

jperson1: Oh, woops! Makes sense.

jperson1 (approved):

PR #4789, opened 2025-03-14
Jadudm/source of truth/data load

jadudm:
This PR adds a Dockerfile, Makefile, and some instructions for repeatably loading data for testing.

Note that you do need to make sure the container that spins up has the right Docker network. (This is documented in the README). When you run this, you're standing up a container _adjacent_ to your `docker-compose` stack, and you have to make sure they're on the same network.


gsa-jrothacker:
You might need to rebase from sot/main ... there seems to be _a lot_ of changes.

jadudm:
This was a bad merge. Will do again Monday.

PR #4809, opened 2025-03-19
Fix for counting of findings to be in-line with previous method

gsa-jrothacker:
## Purpose

The beta summary page and current summary page should show the same data.

## How

Previously, findings were counted based on unique reference numbers. The initial beta page counted all findings regardless of unique reference numbers. Now we pre-compute the unique findings count, and add it to the search indexes.

## Testing Done

End to End Tests, Manual test.

## Notes:

This will require the audit's to be re-migrated from SACs. Until they are re-migrated, the page will show 0 findings.

### DIFF
--- a/backend/audit/models/utils.py
+++ b/backend/audit/models/utils.py
@@ -2,12 +2,16 @@
 The intent of this file is to group together audit related helpers.
 """
 
+import logging
+
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
 from django.db.models import Func
 
 from audit.models.constants import FindingsBitmask, FINDINGS_FIELD_TO_BITMASK
 from support.cog_over_w_audit import compute_cog_over
+
+logger = logging.getLogger(__name__)
 
 
 def generate_sac_report_id(count, end_date, source="GSAFAC"):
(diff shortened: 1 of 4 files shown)


rnovak338:
I completed a new complete manual submission with all the findings workbooks and the results look good now.
I need to blow away the current DB for the e2e tests, I'll hit an approve when those have good outputs too.

backend/audit/models/utils.py
```diff
@@ -40,9 +44,7 @@ class JsonArrayToTextArray(Func):
     output_field = ArrayField(models.CharField())
 
 
-# TODO:
-#    1) We'll want to calculate the cog/oversite for the audit same way as sac, for now just use the sac one
-def generate_audit_indexes(audit, sac):
+def generate_audit_indexes(audit):
```
rnovak338, line 47: This is used only for migrate_audits and the final step when a SAC is disseminated. Are we okay to remove reference to the SAC parameter for this method on the initial deploy?

rnovak338 (changes requested):
2/3 submissions went through on my e2e test (the other timed out), but the other two looked consistent on both beta and live. I am confident you got this working as intended now.

See pending comment before approval.

backend/audit/models/utils.py
gsa-jrothacker: Sudha's change recalculates the cog/over based on audit data only. We should be good with removing sac here. In fact, by removing sac here, it will give us higher confidence that audit is working because we can now compare the cog/over to ensure that is being correctly calculated.

rnovak338 (approved):
LGTM! I was able to verify the audit data matches with the dissemination data under the hood as well as on the UI with these changes.

PR #4815, opened 2025-03-21
File Uploads connection to Audit. Refactoring.

gsa-jrothacker:
## Purpose

Everything should be using the new audit table.

## How

The existing ExcelFile and SingleAuditReportFile tables did not have a reference to Audit. Now they do.

## Testing Completed

Linting, Unit Tests, End to End Testing, Manual Checks

## Notes:

To get the latest changes we'll need to throw-away existing database and rebuild. Shouldn't be a big deal with Matt's new data dumps.

I am still looking if there are more UIs needing beta endpoints.

### DIFF
--- a/backend/audit/exceptions.py
+++ b/backend/audit/exceptions.py
@@ -8,3 +8,10 @@ class SessionWarningException(Exception):
     def __init__(self, message="Your session is about to expire."):
         self.message = message
         super().__init__(self.message)
+
+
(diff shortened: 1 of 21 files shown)


rnovak338:
Added the changes to the migration script here... see the last commit. I figure this is a better place to put the change than in `sot/main`.

rnovak338 (approved):

PR #4816, opened 2025-03-22
Updates the script to handle additional utilities

jadudm:
Resets the `migrated_to_audit` field after data load.

Allows for truncating the `audit_audit` table during testing.

### DIFF
--- a/backend/util/load_public_dissem_data/README.md
+++ b/backend/util/load_public_dissem_data/README.md
@@ -69,14 +69,15 @@ Use the menu script in this folder.
 ./menu.bash
 ```
 
-This will give you three options:
+This will give you several options:
 
-```
-1) Truncate tables
-2) Load data
-3) Check row counts
-3) Quit
-```
+1. Truncate tables
+1. Load data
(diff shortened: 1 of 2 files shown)


backend/util/load_public_dissem_data/README.md
```diff
@@ -136,6 +139,16 @@ Checking counts
...
 ```
 
+## reset migrated_to_audit
+
+This option will set `migrated_to_audit` to `false` for all rows in the `singleauditchecklist`.
+
+This is run after data load, but is included in the menu as an option.
+
+## truncate audit_audit
+
```
rnovak338, line 149: I'm not sure how truncating the audit table with cascading may be useful, because it then wipes out a lot of the rows in tables that correlate to the SAC... which means we can't fully test the migration of audits without reloading the other tables again.

Perhaps I am missing something on a weekend brain.

rnovak338 (changes requested):
These scripts work and run as intended! I just had one comment regarding the new truncation of the audit table.

PR #4822, opened 2025-03-24
Minor tweak to the migration script

gsa-jrothacker:
Only disseminated records _need_ to have the search indexing step performed. Doing this too early causes errors when trying to calculate cog/over.

I ran the full script against the most recent data dump, and it worked.


rnovak338 (approved):
LGTM

Issue #4828, opened 2025-03-24
[SOT2] Test Admin View - Validation Waivers


rnovak338:
We need to replicate the `SacValidationWaiver` in order to support the new Audit model.

- [x] Create a new replica model `AuditValidationWaiver`.
- [x] Ensure that `AuditValidationWaiver` can perform the same functions as the `SacValidationWaiver` in django admin.
- [x] Migrate data from `SacValidationWaiver` to `AuditValidationWaiver` in the `migrate_audits` management command.

jadudm:
@rnovak338 , is this done, or does it have additional work/testing? Or, should this ticket close, and some followup be added to https://github.com/GSA-TTS/FAC/issues/4858? If that makes sense, go ahead and close this, and add a short followup there.

rnovak338:
This should now be tested on staging with the rest of the migration pipeline. I'll link it to the final migration issue.

Issue #4811, opened 2025-03-20
SoT: Admin View - Flagged for Removal Audits


rnovak338:
Similarly to how we can handle SAC records that are `flagged_for_removal`, we need to be able to perform the same actions for Audit records.

jadudm:
@rnovak338 , is this something that is done/will be done before the end of April? Or, should we close as "not planned?" Or, should we add this to https://github.com/GSA-TTS/FAC/issues/4858 as a part of the final migration? Take whatever action is appropriate from a ticket gardening perspective.

rnovak338:
This will be done before end of April and should be included in #4858. This is to handle the deletion of flagged Audits in the same fashion we can delete flagged SACs in Django admin.

I want to test this as part of our staging testing pipeline.

PR #4836, opened 2025-03-27
Introduce Audit actions and AuditValidationWaiver in Django Admin

rnovak338:
Addresses #4811 and #4828.

- Created new model: `AuditValidationWaiver`. Operates the same as `SacValidationWaiver` but focuses on Audits.
- Adjusted `test_admin` to account for the new model as well as testing actions on the audit instead of the SAC.
- Created duplicate methods for handling Audit actions that are equivalent to SAC actions (their original SAC counterparts are marked with a `TODO: SoT` for later reference).
- Created new method in `viewflow.py` for transitioning state of an Audit.
- Linting fixes.

### Setup
1. Clean and rebuild your DB with `make -i docker-clean` and `make docker-first-run`.
2. Boot up your local environment and create many submissions (manually and/or using e2e).

### Audit actions (vs. SAC actions)
1. Log in to Django Admin and attempt to perform the following actions for SOME of the records in the `Audit` table:
    - Flag selected report(s) for removal
    - Revert selected report(s) to In Progress
    - Delete reports flagged for removal for over 6 months. (For audits, you must modify the "flagged_for_removal" history to be at least 6 months ago).
2. Then, attempt to perform the same above actions some of the remaining records in the `SingleAuditChecklist` table. Note that modifying the SAC here will ALSO affect the Audit table, but not vice versa.

### AuditValidationWaiver
1. Create an `AuditValidationWaiver` for one of your records.
2. Validate that this waiver works for the audit during cross-validation.

### Permission handling
Modify your position in `staffusers.json` and re-run the `create_staffusers` script to test the following:
1. As a `read-only` user, I should only be able to VIEW the `Audit` and `AuditValidationWaiver` tables.
2. As a `helpdesk` user, I should be able to also add records to the `AuditValidationWaiver` table.

### Migrate_audits script
1. Boot up your local environment with the prod data load.
2. Validate that the records with a SacValidationWaiver successfully migrate to an AuditValidationWaiver.

## PR Checklist: Submitter

## PR Checklist: Reviewer

### DIFF
--- a/backend/audit/admin.py
+++ b/backend/audit/admin.py
@@ -55,6 +55,24 @@ logger = logging.getLogger(__name__)
 FLAGGED_REPORT_RETENTION_DAYS = 180  # 6 months
 
 
+def admin_message(request, message_type, ui_message, logger_message=None):
+    """Return a message to the user in the Django admin UI. (and optionally send a log)."""
+    messages.set_level(request, message_type)
+
+    if message_type == messages.INFO:
+        messages.info(request, ui_message)
+        if logger_message:
+            logger.info(logger_message)
+    if message_type == messages.WARNING:
+        messages.warning(request, ui_message)
+        if logger_message:
+            logger.warning(logger_message)
+    if message_type == messages.ERROR:
+        messages.error(request, ui_message)
+        if logger_message:
+            logger.error(logger_message)
+
+
 # TODO: SoT
 # No longer needed after SAC deprecation.
 @admin.action(description="Delete reports flagged for removal for over 6 months")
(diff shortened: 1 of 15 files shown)


backend/audit/admin.py
```diff
@@ -109,6 +126,75 @@ def delete_flagged_records(modeladmin, request, queryset):
...
+def audit_delete_flagged_records(modeladmin, request, queryset):
+    """
+    Admin action to delete records flagged for removal older than the specified months.
+    """
+    cutoff_date = now() - datetime.timedelta(days=FLAGGED_REPORT_RETENTION_DAYS)
+
+    # Filter records for deletion
+    records_to_delete = queryset.filter(
+        submission_status=STATUS.FLAGGED_FOR_REMOVAL,
+    )
```
gsa-jrothacker, line 139: Thinking out loud here... Can we update this query to also check the last updated date? It would simplify the loop, and you could then probably remove all the histories work you're doing?

I would imagine that the last_updated_date could be a good reference for when the item was flagged for deletion, since I assume once it's flagged for deletion, it won't be getting updated again?

backend/audit/admin.py
```diff
@@ -109,6 +126,75 @@ def delete_flagged_records(modeladmin, request, queryset):
...
+                        audit_remove_singleauditreport_pdf(audit)
+                        audit_remove_workbook_artifacts(audit)
+                        Access.objects.filter(audit=audit).delete()
+                        DeletedAccess.objects.filter(audit=audit).delete()
+                        SubmissionEvent.objects.filter(audit=audit).delete()
+                        ExcelFile.objects.filter(audit=audit).delete()
+                        SingleAuditReportFile.objects.filter(audit=audit).delete()
+                        report_ids.append(audit.report_id)
+                        audit.delete()
+                        count += 1
```
gsa-jrothacker, line 170: Would we ever do all this in any other spot? If so it might make sense to create a common method `delete_related_audit_entries(audit)`

backend/audit/admin.py
```diff
@@ -534,6 +694,182 @@ def handle_auditee_certification(self, request, obj, sac):
...
+    def has_delete_permission(self, request, obj=None):
+        return request.user.is_staff
+
+    def has_module_permission(self, request, obj=None):
+        return request.user.is_staff
+
+    def has_view_permission(self, request, obj=None):
+        return request.user.is_staff
+
+    def save_model(self, request, obj, form, change):
```
gsa-jrothacker, line 731: I suspect a lot of this was copied/pasted/updated from the existing SAC version. Consider how we can clean it up some. There are a lot of if/elif that _feel_ very similar, wonder if we can extract it out differently to simplify.

And maybe we can't... Just when I see long methods, my first thought is "can it be smaller"

backend/audit/models/viewflow.py
```diff
@@ -228,6 +248,42 @@ def _transition_audit(audit, user, submission_event, autoflow_action):
         )
 
 
+def audit_transition(request, audit, **kwargs):
```
gsa-jrothacker, line 251: Great improvement, thank you

gsa-jrothacker (commented):
No major concerns. I haven't tested it locally yet, but the code looks reasonable to me.

backend/audit/admin.py
rnovak338: I don't think there's a "last updated" date field we store for the `Audit` outside of the `History` table... or maybe I'm just missing it. However, I **think** I could make some adjustments to the queryset initialization to keep the ORM logic outside of the for loop. I'll take a look.

We do have the possible cases when a record is flagged for deletion and then gets unflagged. We still want to store that date when the original flagging occurred. TLDR it is possible for a record to flip flop in and out of the flagged state and we need to be able to retrieve the most recent occurrence.

backend/audit/admin.py
rnovak338: I think because of the gravity of what this logic does, we should not yet turn it into a common method.

However, because this is the only (and very nuanced) process for permanently deleting audit data, I think this code can very well become abstracted if the time comes to introduce another way to delete audits.

backend/audit/admin.py
rnovak338: Yep, we can clean this up. Django admin does require a lot of these small-ish routines when it comes to displaying fields a certain way from the UI, but the ones you've highlighted are not even needed as a matter of fact. We are now using `staffusers.json` to determine who can create/read/update/delete from tables.

backend/audit/admin.py
rnovak338: ^ Scratch the first part - thanks for reminding me of the `UpdatedMixin`. I'll simplify this method.

backend/audit/admin.py
rnovak338: I removed the unnecessary routines under this, but as for `save_model()`, I did not make it more compact - instead I removed some redundancy in the if/elif chain, and added some context for each case, which will help in understanding the logic when it is looked at in the future.

gsa-jrothacker (approved):
Thanks for the updates!

Issue #4826, opened 2025-03-24
Upgrade terraform components to use new provider


asteel-gsa:
As much as we can, we want all terraform to be using the new provider. There will be some items that cannot be ported at this time, and we will have to catalog them as debt, to be upgraded before the v2 depreciation or determine a workaround if they decide to not use them any more.

# Tasks
- [x] Upgrade common components for sandbox 
- [x] Upgrade common components for all envs
- [x] Upgrade meta providers
- [x] Upgrade meta
- [x] Documentation

# Status
### Common Components
 - [x] Initial Overhaul with system working - 7ec1f4e0b4f063b8c6d1c937ee3fb2f0031e5ceb
1. `https_proxy` variable needs to change due to new json
```diff
- export https_proxy="$(echo "$VCAP_SERVICES" | jq --raw-output --arg service_name "https-proxy-creds" ".[][] | select(.name == \$service_name) | .credentials.uri")"
+ export https_proxy="$(echo "$VCAP_SERVICES" | jq --raw-output --arg service_name "https-proxy-creds" ".[][] | select(.name == \$service_name) | .credentials.https_uri")"
```

### Application
- [x] Application overhaul working - 4178d301a4cc6a7b12816f64a9d550d9cb9dd1e7
1. Uses some hardcoded values that can be called as output from modules, pending `v2.3.0` release on terraform-cloudgov repo

### File Scanner
- [x] File Scanner overhaul working - 6c343c32a5a57326f95fbac52a4da24b8a0de494
1. Uses some hardcoded values that can be called as output from modules, pending `v2.3.0` release on terraform-cloudgov repo

### Logshipper
- [x] Logshipper overhaul working - ff49099cd5e9e658587a5425f3d8d8cd7d3dcf3b

### Live Environments
- [x] Shared modules overhaul - 59992ca12f183db5530a88e4f20c34106d0c506e

>[!IMPORTANT]
>This needs to be run locally to debug everything, remove modules from the `.tfstate` and rebuild everything by hand before merging into `main`

### Meta
- [x] Meta overhaul - b48015a08a65fa05985da17bb991abbca2927899

>[!WARNING]
> The following [logic](https://github.com/GSA-TTS/FAC/blob/b48015a08a65fa05985da17bb991abbca2927899/terraform/meta/bootstrap-env/deployer.tf#L32-L51) is not supported by the official provider, remediation will need to occur on these items.

# Documentation
Sandbox we needed to delete the following resources:
[...]

PR #4838, opened 2025-03-27
Chore: Provider Upgrade to 1.4.0

asteel-gsa:
Upgrades the provider to `1.4.0`
- Will fix the route issue in prod when I/Ryan do the same in the `terraform-cloudgov` repo

I will be performing this upgrade in https://github.com/GSA-TTS/FAC/tree/terraform-provider-overhaul as part of #4826 . This is more of a reminder that we can stuff this in prior before I start overhauling the live system terraform.

### Before Merge:
Apply the meta module locally from this branch, refresh the plan on git and validate no changes are expected for meta, then merge


asteel-gsa:
Merged into major overhaul branch

Issue #4827, opened 2025-03-24
[BULK] Automated CSV export to S3


jadudm:
We know that CSV exports were implemented against the old tables, but not the SOTs. 

1. What is the current state of the software?
2. How do we want to do the export? 
3. What do we want this to look like?

We'll assume that the bulk export team is aiming at CSVs. We also know we probably need to do year-by-year exports, or the federal awards table is too long to import into Excel. (There is a 1M row limit in XLSX documents. Also, a 65K column limit, but we won't hit that today.)

#3 is interesting, because we can either 1) dump the way Census used to (e.g. general, federal_awards), or we can do something new/different. :shrug: For now, we just want to evaluate the state of existing work, and we'll update this story based on that.

## What shape should things be?

Because we're using `sling`, we can only compress in one of a few formats. While `gzip` is available, Windows 10 cannot open it natively. So, to make sure the maximum number of users can use the data (including Federal users with limited ability to add software to their machines), we will export uncompressed CSV directly.

### what are the table names?

The table names should be identical to the API endpoints.

* `findings_text`
* `additional_ueis`
* `federal_awards`
* `corrective_action_plans`
* `notes_to_sefa`
* `passthrough`
* `general`
* `secondary_auditors`
* `additional_eins`

## Where should things end up?

The exports are of two kinds: the current data and the migration data.

We will put the current data at the following keys.

Note the naming always uses the full table name, and the filenames include enough information that it is obvious what the file contains.

### full dumps 

The full-table dumps (e.g. the `general` table in a single file, the `federal_awards` table in a single file) will end up at:

`public-data/gsa/full/<table>.csv`

### year-by-year based on audit year

The year-by-year dumps (e.g. the `general` table split out so that one file contains audit year 2016, and another AY 2017, etc.) should end up at 

```
public-data/gsa/audit-year/YYYY-ay-<table>.csv
```

This should yield 9 tables x 17 years = 153 files. 

### year-by-year based on federal fiscal year

The FFY dumps (e.g. the `general` table split so that one file contains Oct 1, 2015 through Sept 30, 2016, another Oct 1, 2016 through Sept 30, 2017, and so on...) should end up at 

```
public-data/gsa/federal-fiscal-year/YYYY-ffy-<table>.csv
```

### hashes for current data

If possible, every file should have a SHA1 hash computed as well.

```
public-data/gsa/federal-fiscal-year/YYYY-ffy-<table>.sha1
```

It would be a single-line file containing only a filehash.

[...]

PR #4842, opened 2025-03-31
Bulk export of audit data

gsa-suk:
This PR addresses bulk data export discussed in https://github.com/GSA-TTS/FAC/issues/4827.  The export_data_audit management command exports data from audit.audit table into S3 in the dissemination format as csv files.  

**Testing:**
1.  Pull branch.  Build the app from scratch.
2. On terminal, run 'docker compose run web python manage.py export_data_audit --year all".
3. Dissemination data from 2016 onwards will get exported to S3.  The console output looks like this:

Screenshot 2025-04-02 at 3 52 21 PM

5. Open minio console and verify that files exist in the gsa-fac-private-s3/bulk_export folder.  The csv files in S3 look like this:

Screenshot 2025-04-02 at 3 50 34 PM

6. To export for an audit year and the corresponding federal year, on terminal, run 'docker compose run web python manage.py export_data_audit --year 2016".  

_On terminal:_

Screenshot 2025-04-02 at 3 56 04 PM

_On Minio console:_
Screenshot 2025-04-03 at 1 19 38 PM

## PR Checklist: Submitter

## PR Checklist: Reviewer


.github/workflows/export-audit-data-to-csv.yml
```diff
@@ -0,0 +1,56 @@
+---
+name: Export Audit Data to CSV
+on:
+  workflow_dispatch:
```
gsa-jrothacker, line 4: is the plan to only run this adhoc?

.github/workflows/export-audit-data-to-csv.yml
```diff
@@ -0,0 +1,56 @@
...
+          - staging
+          - preview
+          - production
+      year:
+        required: true
+        type: string
+        description: Provide audit year.
+
+jobs:
+  scheduled-audit-data-export:
```
gsa-jrothacker, line 21: If we're not going to run the job on a schedule, this either isn't needed, or could be named better.

.github/workflows/export-audit-data-to-csv.yml
```diff
@@ -0,0 +1,56 @@
...
+    if: ${{ github.event_name == 'schedule' }}
```
gsa-jrothacker, line 22: Correct me if I am wrong, but this will never trigger since this action can only be run by hand.

backend/support/management/commands/export_data_audit.py
```diff
@@ -0,0 +1,310 @@
...
+            ReplicationStream(
+                object=f"bulk_export/{{MM}}/{self.friendly_name}.csv",
+                sql=self.query.format(table_name=self.table_name),
+                mode="full-refresh",
+                target_options={"format": "csv"},
+            ),
+        )
+
+    def generate_fed_year_stream(self, audit_year):
+        prev_audit_year = audit_year - 1
```
gsa-jrothacker, line 66: I'm questioning this logic. I think it is _mostly_ true but it is not safe to assume. For example, this [audit](https://app.fac.gov/dissemination/summary/2022-12-GSAFAC-0000360412) was submitted today (4/2/2025) but has an audit year of 2022

backend/support/management/commands/export_data_audit.py
```diff
@@ -0,0 +1,310 @@
...
+                    fac_accepted_date_start=fac_accepted_date_start,
+                    fac_accepted_date_end=fac_accepted_date_end,
+                ),
+                mode="full-refresh",
+                target_options={"format": "csv"},
+            ),
+        )
+
+
+STREAM_GENERATORS = [
```
gsa-jrothacker, line 84: Consider moving these into a new config file. This file is getting long

backend/support/export_audit_sql.py
```diff
@@ -0,0 +1,158 @@
+select_additional_eins = (
+    "select * from api_v1_1_0.additional_eins where audit_year = '{audit_year}'"
```
gsa-jrothacker, line 2: Suggestion (all the querries): you're hard coding the API version which will require a lot of updates if the api version changes.

At minimum: Introduce a constant at the top of the file: `API_VERSION = 'api_v1_1_0` then in all your queries use that. This way if the version changes we only have to change it once.

Even better: See if there is a constant already defined that we can use. Ideally we'd have 1: CURRENT_API_VERSION.

backend/support/export_audit_sql.py
```diff
@@ -0,0 +1,158 @@
...
+
+select_all_secondary_auditors = "select * from api_v1_1_0.secondary_auditors"
+
+
+select_fed_year_additional_eins = (
+    "select * from api_v1_1_0.additional_eins "
+    "where report_id in "
+    "(select report_id from api_v1_1_0.general "
+    "where fac_accepted_date>='{fac_accepted_date_start}' "
+    "and fac_accepted_date<= '{fac_accepted_date_end}')"
```
gsa-jrothacker, line 80: why use the fac_accepted_date instead of the audit_year? (Same with all the below)

backend/support/management/commands/export_data_audit.py
```diff
@@ -0,0 +1,310 @@
...
+    StreamGenerator(
+        friendly_name="General",
+        table_name="general_information",
```
gsa-jrothacker, line 87: You don't actually need a table name since all the queries were created separately. When I created this in the other export, there were only 2 queries defined.

I think you could actually get away with this too. All the queries above are very similar. I bet if you abstracted out the table name they'd be all almost the same.

backend/support/export_audit_sql.py
```diff
@@ -0,0 +1,158 @@
...
+    "select * from api_v1_1_0.federal_awards where audit_year = '{audit_year}'"
+)
+
+
+select_findings = "select * from api_v1_1_0.findings where audit_year = '{audit_year}'"
+
+
+select_findings_text = (
+    "select * from api_v1_1_0.findings_text where audit_year = '{audit_year}'"
+)
```
gsa-jrothacker, line 29: We need to be sure we're not exposing tribal data.  Does the API schema filter these out?

Applies to cap_text, notes_to_sefa, and findings_text

gsa-jrothacker (changes requested):
I feel like there could be a lot of improvements, let me know if you want to pair on some of this. I am particularly worried about usage of the fac_accepted_date and tribal data.

backend/support/management/commands/export_data_audit.py
danswick: "Audit year" isn't dependent on the submitted date; only on the fiscal year period the audit itself covers. Specifically the fy_start date. So I believe this logic should be ok!

gsa-jrothacker: I may have highlighted the wrong chunk of code... lines 67-68 seem a bit off.
```
fac_accepted_date_start = str(prev_audit_year) + "-10-31"
fac_accepted_date_end = str(audit_year) + "-09-30"
```

I admit I can be completely missing what the goal of these queries are. I just worry about tying fac_accepted_date with the audit year.

gsa-jrothacker: Nevermind... it's late on a Friday... I think I understand now. We are using audit_year for two slightly different meanings.

1. All records based off the audit year for the organization.
2. All records submitted in that federal year.

I am curious what is the use case for breaking it down this way?

jadudm: Through _mystical_ means, we have ascertained that Federal users would like both of these breakdowns to support their work. (That is, both an audit-year breakdown as well as a federal fiscal year breakdown.)

backend/support/export_audit_sql.py
gsa-suk: The new API PR creates these views only for public data - https://github.com/GSA-TTS/FAC/pull/4832/files

backend/support/export_audit_sql.py
gsa-suk: Matt mentioned that users need data pulled by audit year and also fac_accepted_date.

backend/support/export_audit_sql.py
gsa-suk: The SOT version is api_v1_2_0.  Current Prod version is api_v1_1_0.  When Prod switches from SAC to SOT, the plan is to rename the SOT api to api_v1_1_0 to not impact the users.

PR #4847, opened 2025-04-01
Copy over datasource during migration

gsa-jrothacker:
Minor fix for the data_source field in the SOT migration.


rnovak338 (approved):
Great catch.

phildominguez-gsa (approved):
Confirmed the data_source error no longer shows up when SOT validating 2020-12-CENSUS-0000255831

PR #4849, opened 2025-04-02
Updating the previous references validation to use Audit table

gsa-jrothacker:
## Purpose

The cross validation step for findings prior references was not using the Audit table.

## How

During the launch of SOT, if there is an audit with the report_id, we'll also check the audit table and compare any errors but continue to use the dissemination tables for SAC validation.

Post SOT launch, this logic can be removed and solely utilize the SOT.

## Testing Completed

Linting, Unit Tests, E2E, Manual.


rnovak338 (approved):
LGTM - and tests passed.

Issue #4862, opened 2025-04-07
[BULK] Downloading of 2015- and prior reports


jadudm:
We are trying to provide some more reports along with the historic data dumps.

We think we want to do the following.

1. We want a new URL/view for these that is separate from the "live" PDF downloads. This might be reused for the historic CSVs and bulk data dumps as well. 
    1. `.../public-data/report/YYYY-MM-CENSUS-###..` will be a PDF report. (YYYY must be 1999-2015)
    2. `.../public-data/current/full/<table>.csv`
    3. `.../public-data/current/audit-year/YYYY-ay-<table>.csv` TBD
    4. `.../public-data/current/fiscal-year/YYYY-ffy-<table>.csv` TBD
    5. `.../public-data/historic/census-YYYY.zip` TBD
2. Data needs to go into the bucket in the right places
3. Add a test or two.
4. Static site linkages (see [draft](https://drive.google.com/drive/folders/1vUUgWtSpHQ6aIZHWfexhw4EzufkobH2i))

The view might start to look like

```

class HistoricPdfDownloadView(ReportAccessRequiredMixin, View):
    def get(self, request, report_id):
        """
        Given a historic report_id in the URL, make sure it is old, and then provide it.
        """
        if re.match("[1998,1999,2000...2015]-CENSUS-\d{10}", report_id[0:4]):
            return redirect(get_download_url(filename))
        else:
            blah()
```

except we'll probably call it something like `PublicDataDownloadView` or similar, because we'll use it for all those things.

PR #4871, opened 2025-04-08
Bulk Download URL & View

jperson1:
# Bulk Download URL & View

Issue: https://github.com/GSA-TTS/FAC/issues/4862

## Changes:
Just the URL and view. Which is essentially a passthrough for users to download files in the `public-data` folder of our S3 bucket. No user feedback (unless there's a 404), and no file perusal. The intended access point is outside links (the static site). 

Some little tests for verification.

Also, a local linting fix. Not sure why that doesn't get hit in the PR checks or in the container. 

## How to test:
1. Switch to this branch, run the app normally. 
2. Populate Minio with some goodies to download
    a. The default path is http://localhost:9002
    b. The default login is `minioadmin` for both the user and pass
    c. Add some files to a folder named `public-data`
3. Attempt to download the files in the folder
    a. For example, `http://localhost:8000/dissemination/public-data/FILE_RELATIVE_PATH.EXT`
    b. If you bring in data from GDrive, try something like http://localhost:8000/dissemination/public-data/historic/2022.zip
    c. Attempts should succeed whether the file is directly in the folder or in a subfolder, like above

## Recording:

https://github.com/user-attachments/assets/09cb9f7c-fd78-4a3b-aeb9-5157a0af19f6

## PR Checklist: Submitter

## PR Checklist: Reviewer

### DIFF
--- a/backend/dissemination/views/download.py
+++ b/backend/dissemination/views/download.py
@@ -94,6 +94,7 @@ class PublicDataDownloadView(View):
         Ex. Given 'historic/2022.zip', attempt to serve the file located at 'BUCKET/public-data/historic/2022.zip'.
         """
         relative_path = f"/public-data/{relative_path}"
+        logger.info(f"Attempting to serve file {relative_path} from S3.")
         return redirect(get_download_url(relative_path))
 
 
--- a/backend/dissemination/views/download.py
+++ b/backend/dissemination/views/download.py
@@ -84,6 +84,19 @@ class XlsxDownloadView(ReportAccessRequiredMixin, View):
         return redirect(get_download_url(filename))
 
 
+class PublicDataDownloadView(View):
+    def get(self, request, relative_path):
+        """
+        Given a relative path, prepend '/public-data/' and attempt to serve the file.
+        The only user feedback is a file download or a 404 error.
+        This is meant to serve single files of at static, known locations.
+
+        Ex. Given 'historic/2022.zip', attempt to serve the file located at 'BUCKET/public-data/historic/2022.zip'.
+        """
+        relative_path = f"/public-data/{relative_path}"
+        return redirect(get_download_url(relative_path))
+
+
 class OneTimeAccessDownloadView(View):
     def get(self, request, uuid):
         """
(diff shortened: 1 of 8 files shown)


backend/dissemination/views/download.py
```diff
@@ -84,6 +84,19 @@ def get(self, request, report_id, file_type):
...
+class PublicDataDownloadView(View):
+    def get(self, request, relative_path):
+        """
+        Given a relative path, prepend '/public-data/' and attempt to serve the file.
+        The only user feedback is a file download or a 404 error.
+        This is meant to serve single files of at static, known locations.
+
+        Ex. Given 'historic/2022.zip', attempt to serve the file located at 'BUCKET/public-data/historic/2022.zip'.
+        """
+        relative_path = f"/public-data/{relative_path}"
```
jadudm, line 96: Do we want to process that value at all? E.g. 

* make sure that the only values in it are [a-zA-Z0-9.-_]
* make sure it starts with [a-zA-Z]

I mostly don't want it to start with `..`, and... not sure what else we want to do here.

jperson1: I believe we're protected from this, since the value comes from the URL? Maybe not, that's a good callout.

Django itself doesn't do any processing on the string, so I think it's just relying on it being a valid URL. https://github.com/django/django/blob/71a19a0e475165dbc14c1fe02f552013ee670e4c/django/urls/converters.py#L39-L41

I'll add a more restrictive converter.

jperson1: Done! I used

`[A-Za-z][A-Za-z0-9\/\-.\\_]+`

One character, followed by any number of either characters or "/", "-", ".", "\", or "_". 

I could have been less explicit, but this seemed okay since we know the actual names of all served files and this covers it.

jadudm: Excellent. Thank you. That was the spirit of what I was aiming for. If we ever expand this to cover other things, we'll know really quickly that this is the source of the problem. It assuages my paranoia, if nothing else. :D

PR #4890, opened 2025-04-11
<WIP> Removing SAC and Dissemination.

jrothacker:
As promised, I pushed this branch.

This is a lot of work in progress. Most tests are still failing. We need to work together to fix:
1. Unit Test Issues
2. Linting issues
3. End to End tests Issues.

Luckily fixing Unit Test issues, usually fixes linting issues at the same time.

We can sync in the next stand about who is working on what.


backend/curation/sql/enable_curation_auditing.sql
```diff
@@ -1,2 +1,2 @@
-select curation.enable_tracking('public.audit_singleauditchecklist'::regclass);
+select curation.enable_tracking('public.audit_audit'::regclass);
 select curation.enable_tracking('public.support_cognizantassignment'::regclass);
```
gsa-suk, line 2: Please remove line 2 related to cognizantassignment.

backend/requirements.txt
gsa-jrothacker, line 1: Don't commit this.

backend/audit/management/commands/move_to_disseminated.py
gsa-jrothacker, line 1: We maybe able to delete this.

jperson1: There are still a few (9) records that are in the "Accepted" state. Perhaps it's okay that they stay that way, but since this command will still work I think it's good to keep it.

backend/requirements.txt
jperson1: Will be resolved by https://github.com/GSA-TTS/FAC/pull/4940

PR #4927, opened 2025-04-17
Updating the submission process to go straight to dissemination

gsa-jrothacker:
## Purpose

We don't really need two steps for submit -> disseminate anymore. On submission go straight to disseminate.

## How

On disseminate, added an extra history object to keep the consistent event history, but otherwise we update directly to disseminate. This also removes all the race condition issues as dissemination is the same as any other event on audit.

## Testing Completed

Linting, Unit Tests, End to End tests.


jperson1 (approved):
Also looks good. Tested report goes right through, and `audit_history` reflects moving right on to `disseminated`.

backend/audit/models/audit.py
```diff
@@ -62,6 +62,20 @@ def create(self, **obj_data):
             }
 
             result = super().create(**updated)
+
```
gsa-jrothacker, line 65: This chunk needs to be in save

jperson1:
As suggested by the comment, I moved the History object create into the save. Now, it adds the "submitted" event and goes right through to "disseminated".

Screenshot from local DB:
image

PR #4944, opened 2025-04-21
Small fix

asteel-gsa:
App hasn't deployed, so it doesnt know about the new proxy variable. Fixes github in the allow list


jadudm (approved):
Fixes the sequencing post TF update. LGTM.

PR #4946, opened 2025-04-21
Uncomments the predeploy backup

asteel-gsa:
The application cannot do a backup because the application has not changed the `$https_proxy` variable. We commented this out to get a deploy going and get `prod` operational.


jadudm (approved):
LGTM.

PR #4964, opened 2025-04-24
Downscale prod to try and stop pushing against the quota

asteel-gsa:
Try to downscale prod to fix borked deploy


jadudm (approved):
LGTM. Avoiding RAM caps.

PR #4966, opened 2025-04-24
Add file download check + SOT `if` conditional

jadudm:
This does two things:

1. Adds some additional Cypress tests to make sure files can be downloaded.
2. Adds some more `if` conditionals around the `audit` table, which might not exist if the audit was created before last week (of which there are 400-ish).

This should allow users to see files and download them in cases where they have a `sac` but not a `sot` object.


jperson1 (approved):
Fantastic, thanks!

PR #4978, opened 2025-04-29
[Merge After 4/30] Contractor Offboarding

asteel-gsa:
🫡

### DIFF
--- a/backend/config/staffusers.json
+++ b/backend/config/staffusers.json
@@ -1,7 +1,6 @@
 {
   "readonly": [
-    "philip.dominguez@gsa.gov",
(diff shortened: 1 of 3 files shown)


backend/config/staffusers.json
```diff
@@ -1,7 +1,6 @@
 {
   "readonly": [
-    "philip.dominguez@gsa.gov",
-    "sudha.kumar@gsa.gov"
+
```
jperson1, line 3: The helpdesk ones below, as well

asteel-gsa: forgot to push this! ty!

Issue #4837, opened 2025-03-27
Chore: Upgrade all network policies to use the cloudfoundry/cloudfoundry provider


asteel-gsa:
### Upgrade Network Policies

Based on this commit, beff02dfdd55469fea6f9513dece971c1ff9d80c all network policies on the community provider are now in their own files. These will need to be updated to the official provider when the resource is available

PR #4993, opened 2025-05-29
Terraform provider update (1.6.0)

asteel-gsa:
`1.6.0` release includes a wonderful [item](https://github.com/cloudfoundry/terraform-provider-cloudfoundry/pull/171) from Ryan that will allow us to close the loop on #4837 

- [x] [Preview](https://github.com/GSA-TTS/FAC/actions/runs/15333704219/job/43146337446) Deploy
- [x] Sanbox Deploy


asteel-gsa:
> LGTM but is it OK to merge with those failing tests...?

Linting I was just able to fix. 
Django and Cypress however, probably need to consult with @jperson1 @rnovak338 and/or @phildominguez-gsa

Issue #5097, opened 2025-07-10
Management command for UEI changes


jadudm:
## problem

Despite our best efforts and the efforts of our users, sometimes a submission happens for an incorrect UEI. As an example, this is especially problematic for organizations with names like "Springfield Housing Authority." 

<img width="678" height="462" alt="Image" src="https://github.com/user-attachments/assets/e57ff874-44f0-44cb-922c-dc7a502b441e" />

A certifying official may contact us and indicate that they submitted for UEI `ABC` when they meant to submit for UEI `DEF`. 

Resubmission, as a process, will handle many other kinds of data errors, but not this. Changing the UEI requires administrative action by the FAC.

## solution

We will write a management command that takes as input

* the report id
* the old UEI
* the new UEI
* the email address who is in `staffusers` and has helpdesk or higher access

and do the following:

1. Load the SAC object
2. Confirm that the old UEI matches the command input
3. Confirm that the email address is in `staffusers`
4. Change the UEI
5. With curation tracking enabled, save the SAC
8. Add an event to the `submissionevent` table indicating the UEI was updated by  administrative action.
9. Delete records associated with that report ID from dissemination and re-disseminate the record.

After the command is run, we will add a note to the curation log indicating the UEI update: https://www.fac.gov/data/reliability/curation-log/

## controls

To do this, one must be able to enable SSH on the FAC in production.

This means that only developers on the FAC can make this change, and our SSH-enabling protocols will further prevent this from being casual or accidental. (Being able to enable SSH means that the user has onboard to GSA, TTS, the FAC, and the access controls underwent code review and approval by a Code Owner.) While this is heavy as a process, it is simple to implement, and solves a known problem *now*, and can be streamlined (if necessary) in the future.

Part of our controls will also need to be **standard operating procedure**. For further consideration:

1. A request to make such a change must come from a certifying (auditor/auditee) official. 
2. We provide an attestation template for them to use and return; that attestation goes into our GDrive.
3. The change is executed.
4. The change is logged to our curation log.

This is in keeping with other processes the FAC uses for exceptional circumstances.

## we'll know we're done when

- [ ] We have implemented the command and have tests that confirm its correctness
- [ ] We have updated the helpdesk SOP for dealing with UEI change requests

jadudm:
[...]

PR #5108, opened 2025-07-15
Update UEIs or EINs for submitted audits

jadudm:
## Updating UEIs and EINs for submitted audits

https://github.com/GSA-TTS/FAC/issues/5097

As we continue working on data quality, and especially as we implement resubmission, we're going to need a way to update audits where a critical linking value is incorrect. 

For example, a user recently wrote the helpdesk and said they discovered they had submitted under the wrong UEI. We should curate this, as it makes their audit submission *wrong*. Similarly, a day later, another user discovered their EIN was incorrect. Both of these are used by agencies as identifiers for entities.

This tool can only be run from the command-line as a management command. To be used in production, we must enable SSH, log in, run the command, and disable SSH. Therefore, it is not a light lift to do, and we expect this to be an infrequent ask. If that changes, we will change the way the tool is run.

We anticipate the following SOP:

1. Ask the entity to provide a statement attesting to the error, indicating the UEI they incorrectly filed under, and instead indicating the UEI they should file under. (Provide a template.) 
2. We confirm/check their UEI(s) in SAM.
3. Add that document to our GDrive.
4. Update the UEI/EIN in production.
5. Add a curation log entry on the static site.

### Running the command

```
python manage.py update_uei_or_ein_for_submitted_audit \
    --report_id 2023-05-GSARID-0000000005 \
    --email developer@gsa.gov \
    --old_uei ABC123DEF456 \
    --new_uei 456DEF123ABC
```

or

```
python manage.py update_uei_or_ein_for_submitted_audit \
    --report_id 2023-05-GSARID-0000000005 \
    --email developer@gsa.gov \
    --old_ein 555555666 \
    --new_ein 555555777
```

This should only work if:

1. Both UEIs or EINs validate in shape. (We do not check SAM for UEIs as part of this script.)
2. You include both old/new values for either UEI or EIN, but not a mix of both.
3. The email address is in the system as a valid user. (The only way a developer should end up in the system is via the `staffusers.json` file.)
4. The audit has been `disseminated`. (This does not work on audits that are in-progress.)

When it is complete, it will do the following:

2. Apply an administrative override to save the SAC (we don't normally save records that are already disseminated), 
3. Add an entry to the `submissionevent` table documenting the change.
4. Extend the `transition_name` and `transition_date` with a new `disseminated` entry.
5. Re-disseminates the record.

### demo

[update-ein-demo.webm](https://github.com/user-attachments/assets/34bd1e8e-5995-4252-a774-72a712232ab6)

## NOTE

This code also changes the `SingleAuditChecklist` model. 

1. It modifies `save()` to handle the administrative override.
2. It adds a `redisseminate()` method to handle deleting `dissemination_*` audits for this SAC, and then it calls `.disseminate()` on the SAC to publish the new information. This happens atomically.

### testing

```
python manage.py test curation.test_uei_ein_replacement
```

which runs tests against an invisible mocked DB, and manual testing against a live/local instance.

## PR Checklist: Submitter

## PR Checklist: Reviewer


backend/curation/curationlib/update_uei_or_ein.py
```diff
@@ -0,0 +1,144 @@
...
+    #     "General": General,
+    #     "Passthrough": Passthrough,
+    # }
+    return _dissemination_models
+
+
+def get_named_parts_containing_ueis(sac):
+    return {
+        "AdditionalEins": sac.additional_eins,
+        "AdditionalUeis": sac.additional_ueis,
```
phildominguez-gsa, line 37:
```suggestion
        "AdditionalEINs": sac.additional_eins,
        "AdditionalUEIs": sac.additional_ueis,
```
Tested on full-submission submissions

backend/curation/curationlib/update_uei_or_ein.py
```diff
@@ -0,0 +1,144 @@
...
+
+    sac.general_information["auditee_uei"] = THE_NEW_UEI
+    for json_field, object in named_parts.items():
+        # We might not have this workbook.
+        if object:
+            # If we do have the workbook, we MUST have auditee_uei as a field.
+            if json_field in object:
+                object[json_field]["auditee_uei"] = THE_NEW_UEI
+            else:
+                logger.error("Could not find auditee_uei in XLSX JSON object")
```
phildominguez-gsa, line 115:
```suggestion
                logger.error(f"Could not find auditee_uei in XLSX JSON object {json_field}")
```
This helped me debug

phildominguez-gsa (changes requested):

jperson1 (approved):
Works well for me! I changed a few records without issue. SubmissionEvents go through, invalid values are rejected.

I had a weird case where there were two admin users for my email, which broke. That's fine, I suspect. 

image

image

PR #5112, opened 2025-07-16
Update proxy allow list

asteel-gsa:
After noticing multiple failures on our cron jobs our proxy seems to be blocking a necessary endpoint/redirect for github content. Due to this, it was returning a 403 on the initial curl when attempting to get the necessary release binaries for the cron to run. Prior to this, `objects.githubusercontent.com:443` was enough to allow traffic to get said artifacts, and now, `release-assets.githubusercontent.com:443` seems to be a new redirect that it favors. Sporadically when running the job enough, it will go through, so presumably `objects.githubusercontent.com` has now become a secondary redirect.


jadudm (approved):
LGTM.

Issue #5119, opened 2025-07-18
[RESUB} Add Resubmitted tag to the search results


marissahenderson:

RanyeM:
Tags can be "Resubmitted" or "Most Recent".
- A SAC should only have a tag if it is a resubmission (i.e., it has resubmission metadata).
- Compare the previous SAC report ID to determine the relationship.
- If the SAC has a resubmission status of "Original or Deprecated", display the Resubmitted tag.
- If the SAC has a resubmission status of "Most Recent", display the Most Recent tag.
- Do not display tags for SACs that are not resubmitted or are old legacy SACs.
- Display only one tag per SAC — either 'Resubmitted' or 'Most Recent', never both.

PR #5193, opened 2025-08-15
Feat(ui): ] Add resubmitted and most recent tags to the search result…

RanyeM:
…s UI

#5119

## PR Checklist: Submitter

## PR Checklist: Reviewer

### DIFF
--- a/backend/dissemination/templates/search.html
+++ b/backend/dissemination/templates/search.html
@@ -156,7 +156,10 @@
                             <tbody>
                                 {% for result in results %}
                                     <tr>
-                                        <th scope="row" role="rowheader">{{ result.auditee_name }}</th>
+                                        {{ result.auditee.name }}
+                                        {% if result.resubmission_tag %}
+                                        <span class="usa-tag">{{ result.resubmission_tag }}</span>
+                                        {% endif %}
                                         {% comment %} Display UEI. If it's "GSA_MIGRATION", use the EIN instead. If no EIN, just show "GSA_MIGRATION". {% endcomment %}
                                         <td>
                                             {% if result.auditee_uei != "GSA_MIGRATION" %}
--- a/backend/dissemination/views/search.py
+++ b/backend/dissemination/views/search.py
@@ -2,7 +2,7 @@ from datetime import date
 import logging
 import math
 import time
-
+from audit.models.constants import RESUBMISSION_STATUS
 from django.core.paginator import Paginator
 from django.shortcuts import render
 from django.utils.decorators import method_decorator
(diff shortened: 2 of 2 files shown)


backend/dissemination/test_search_resub_tags.py
```diff
@@ -0,0 +1,96 @@
+from django.test import TestCase
+from dissemination.searchlib.search_resub_tags import (
+    build_resub_tag_map,
+    attach_resubmission_tags,
+)
+
+
+class ResubmissionTagTests(TestCase):
+    def test_tag_most_recent(self):
+        rows = [
```
jadudm, line 10: Using `baker.make(...)`, it is possible to create some generic `General` objects that have just enough fields that you can then test this code using the actual models/objects we'll see in production. That might be preferable? Happy to pair to flip these over.

backend/dissemination/searchlib/search_resub_tags.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def _normalize_status(val: Optional[str]) -> Optional[str]:
+    if val is None:
+        return None
+
+    status = str(val).strip().upper()
+
+    # Normalize known variants to canonical forms
+    if status in (
```
jadudm, line 20: (Out of order... leaving this note last...)

I asked a question to the dev team in chat about this one... I'm pretty sure we should not have to worry about multiple possible values, but I'm not 100% sure. Ideally, given that we control these constants, we should not have to resolve them/normalize them this way. If it is true that this is just a side-effect of the development process you went through, we should be able to get rid of this code, and use the value from the column directly. (E.g. because it is a constant, we can ask "does the value in this column equal `RESUBMISSION_STATUS.DEPRECATED` ... or whatever we named the constant. That way, if it ever changes in the future, the code would still "just work.")

I have a suspicion this is all because the code started before we had the model fully in-place. :( Lesson learned. Apologies on not getting the ticketing order right...

jadudm (changes requested):
Structure is good, it works well, but we have a few remaining cleanups that will help with maintainability, I think? Feel free to grab me to pair and knock these out.

backend/dissemination/searchlib/search_resub_tags.py
RanyeM: No worries! Yes, we do have constants, so this can be removed. Will used the RESUBMISSION_STATUS.DEPRECATED_VIA_RESUBMISSION

backend/dissemination/searchlib/search_resub_tags.py
```diff
@@ -0,0 +1,83 @@
...
+    ):
+        return "DEPRECATED"
+
+    if status in ("MOST_RECENT_SUBMISSION", "MOSTRECENT"):
+        return "MOST_RECENT"
+
+    return status
+
+
+def _resolve_report_id(row: Row) -> Any:
```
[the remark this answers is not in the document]
RanyeM: changed to row:General

backend/dissemination/searchlib/search_resub_tags.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Do not tag if status is missing or unknown
+    return None
+
+
+def build_resub_tag_map(rows: Iterable[General]) -> Dict[Any, Optional[str]]:
+    """
+    Builds a mapping of report_id → tag using already-loaded data from general search results.
+    """
+    tag_map: Dict[Any, Optional[str]] = {}
```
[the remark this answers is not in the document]
RanyeM: Yes, they are strings. Good catch. Replaced Any -> str

phildominguez-gsa (approved):

PR #5415, opened 2026-03-04
Disabling Metabase For a Bit

jperson1:
# Disabling Metabase For a Bit

## Related tickets

* N/A

## Description of changes

* Disable metabase to enable deploys. It shall return once we have the necessary permission. 

## How to test

* Vibe check it after the deploy


ccampbell-fearless (approved):
LGTM

PR #5554, opened 2026-04-15
Fixes for sanitized dumper

### LINKED ISSUE #5547: [RESUB DE-DUP] A New Data Dump for Local Testing
To ensure we get an appropriate breadth of testing done, we want a more up to date data dump for our local systems. We'll want to run the action on the full dataset, rather than the truncated 20,000.

Plus, we've added a column or two and that made the older dumps harder to use.

Follow the instructions laid out [here](https://github.com/GSA-TTS/FAC/blob/main/backend/util/load_public_dissem_data/README-prepare_prod_data_for_local.md) to generate a sanitized data dump usable by our team. Add this dump to the Drive to share out. Then, load it into our lower environments (Dev/Staging/Preview)

phildominguez-gsa:
## Related tickets

* Closes https://github.com/GSA-TTS/FAC/issues/5547

## Description of changes

* Adding tools back to the Dockerfile that are required for the dumper
* Replacing accidental `keepers` variable with correct `report_ids_to_delete` in `remove_tribal_audits.sql`
* Removing `DESTINATION` input check because it's already hard coded
* Renaming `DATE` to `TODAY` because that's what it is
* Renaming `test_sanitized_production_dump` to `reload_and_test_sanitized_production_dump` because that's what it does
  * Some other improvements here too, basically allowing the user to specify which dump they want to load
* README tweaks
* Not addressed in this PR, but we think we discovered a weird timing issue with the backups. In one instance, notes had a report_id that was not yet present in in disseminated_general. Created a ticket [here](https://github.com/GSA-TTS/FAC/issues/5556).

## How to test

* Warning: This dumping/loading takes a decent amount of time to complete
* Follow the quickstart instructions in `README-prepare_prod_data_for_local.md` to run options 1-5
* Download the 20260414 dump [here](https://drive.google.com/drive/u/1/folders/1WymwJtdQ287SdgrOx__aEraoVTx7ig9D) and put it in `backend/util/load_public_dissem_data/data`, then run option 7 to load it
* Your tables should generally be populated
* Querying `is_public=false` on disseminated_general should yield nothing


RanyeM:
Querying is_public=true on disseminated_general should yield nothing -> my table still shows rows but when  is_public=false, it shows nothing

phildominguez-gsa:
> Querying is_public=true on disseminated_general should yield nothing -> my table still shows rows but when is_public=false, it shows nothing

Oops! I've now fixed that in the instructions. Yes, `is_public=false` should yield nothing, since those are private/tribal.

RanyeM (approved):
Lgtm!

PR #5559, opened 2026-04-16
Using SAM response to populate Gen Info

### LINKED ISSUE #5467: [SPIKE] Determine What Info Can Be Pre-Populated From SAM
Pulling from SAM into the General Information form (and any other relevant sections).

As a user, I need have as much UEI information pre-populated from an official source so that I don't input an incorrect fields and can help validate I have the correct UEI.

We'll want to:
1. This ticket - Determine what SAM provides, and the places that it will be when it comes back from the API.
    * It's a different spot between active and inactive entities.
    * SAM may or _may not_ have everything for every entity.
2. Followup ticket - Store gathered info from SAM to the user profile, then pre-populate the general information form with it on SAC creation

phildominguez-gsa:
## Related tickets

* Closes https://github.com/GSA-TTS/FAC/issues/5467

## Description of changes

* The `AuditeeInfoForm` now stores the SAM response. This is used to populate some more fields in `user.profile.entry_form_data`, which was [already being used](https://github.com/GSA-TTS/FAC/blob/main/backend/api/views/access.py#L46) to populate gen info fields during sac creation.
* Specifically, it's populating `auditee_name`, as well as the auditee address fields. It uses their mailing address, as opposed to physical address, which was greenlit by Lynn.
* Some `AuditeeInfoForm` cleanup, like using `get`s, improving readability, and removing redundant `is_valid()`

## How to test

* Start making a submission as normal. When you get to gen info, the name and address fields should be populated

## Screenshots and recordings

Idk if this necessitates screenshots but here you go:
image
image


LynnMHouston:
Virginia State Audit Director, copying HHS OIG, made the case for mailing address (over physical address) in this ticket: https://fac-gov.zendesk.com/agent/tickets/8638

jperson1 (approved):
Looks good, thanks!

PR #5570, opened 2026-04-21
added filters for deploys and removed duplicate steps

ccampbell-fearless:
## Description of changes

* Removed jobs from `daily-deploy-staging.yml` since the same jobs are executed in the PR checks
* Removed duplicated linting and testing jobs
* Added `check-for-changes` job in all deployment workflows to prevent deploying when no code has changed
* Removed terraform linting from `pull_request-linting.yml` so that it runs only on terraform changes

### DIFF
--- a/.github/workflows/daily-deploy-staging.yml
+++ b/.github/workflows/daily-deploy-staging.yml
@@ -16,34 +16,9 @@ jobs:
       actions: read
       security-events: write
 
-  build-container:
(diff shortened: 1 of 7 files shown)


jperson1 (approved):
Reviewed in a big 'ole group with @daryll, @phildominguez-gsa, @LynnMHouston, @annie-maria, and @RanyeM.

PR #5578, opened 2026-04-22
More fixes for resubission meta

phildominguez-gsa:
## Related tickets

* Nope

## Description of changes

* Missed a few things from the last PR:
  * The old sac wasn't being saved with the new data
  * `submission_status` now becomes "resubmitted"
  * `resubmission_meta.resubmission_status` now becomes "deprecated_by_resubmission". This is the value we check on the resubmission start page.
  * Adds `resubmitted` to list of transitions
    * Had to refactor `sac_transition()` because of flake8 complexity

## How to test

* Run `resubmission-full`
* All `audit_singleauditchecklist.resbmission_meta` values should be correct
* All `audit_singleauditchecklist.submission_status` should be correct
* Try to start resubmissions for the deprecated submissions. It should fail.
* My Submissions list should look correct. Clicking on a resubmission should work and have the correct resubmission banner.


backend/audit/views/submissions.py
```diff
@@ -192,16 +192,24 @@ def post(self, request, *args, **kwargs):  # noqa: C901
...
                     )
                     old_sac.resubmission_meta = {
+                        "version": old_resubmission_meta.get("version", 1),
+                        "resubmission_status": RESUBMISSION_STATUS.DEPRECATED,
                         "previous_report_id": old_resubmission_meta.get(
                             "previous_report_id", None
                         ),
                         "previous_row_id": old_resubmission_meta.get(
                             "previous_row_id", None
                         ),
```
daryll: if these are null then we should leave them out of the JSON obj

daryll (changes requested):

daryll (commented):
this works great!

should we set submission_status == deprecated_via_resubmission or do we keep it as 'disseminated' in the 'audit_singleauditchecklist' table?

daryll:
> this works great!
> 
> should we set submission_status == deprecated_via_resubmission or do we keep it as 'disseminated' in the 'audit_singleauditchecklist' table?

https://github.com/GSA-TTS/FAC/blob/main/backend/audit/check_resubmission_allowed.py#L20 states we keep it as 'disseminated' and our design doc does not mention changing the submission_status

daryll (commented):
it seems the backend tables are all correct now 
although the first audit on the audit page now shows "In progress"

daryll (approved):
this all looks great!

many thx!

PR #5594, opened 2026-04-23
[DRAFT] Bumping `boto3`

jperson1:
We're seeing strange errors on file uploads in our testing environments. These run in a freshly built web container, into a "fake" s3 container (minio). I suspect there's a config issue somewhere that has only recently cropped up. I cannot get this issue to occur locally.

Partly as a sanity check, I'm bumping `boto3` here to see if it changes anything.

## Related tickets

* N/A

## Description of changes

* Pin `boto3`, bump `django-storages[boto3]` to 1.12.6

## How to test

* N/A

## Screenshots and recordings

* N/A

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -24,15 +24,15 @@ attrs==24.2.0 \
     # via
     #   jsonschema
     #   referencing
-boto3==1.35.41 \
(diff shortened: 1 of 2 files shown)


jperson1:
This decidedly did not fix the issue. A good check, though - it rules out a number of concerns.

PR #5600, opened 2026-04-27
[DRAFT] Add a Healthcheck to the ClamAV Container

jperson1:
Tests were failing on file uploads. This could point to lots of different issues. There could have been updates to; Minio, our local dependencies or connection config, the ClamAV upstream, or container connections. We've only really poked with ClamAV in the days before the issues appeared.

Something may have changed in ClamAV that had it start up slower, or cause issues if our tests begin before it spins up.

## Related tickets

* N/A

## Description of changes

* Add a healthcheck to `clamav-rest`. Require a positive healthcheck before `web` comes up.
* Add a -L flag to the `mc` install in the Dockerfile. Missing `mc` may cause the upload configs to be broken.

## How to test

* Spin things up locally, make sure it all runs right.
* Eyeball the CI/CD tests.

## Screenshots and recordings

* N/A

### DIFF
--- a/.github/workflows/testing-from-build.yml
+++ b/.github/workflows/testing-from-build.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Start services
         working-directory: ./backend
-        run: docker compose -f docker-compose.yml up
+        run: docker compose -f docker-compose.yml up -d
 
       - name: Run Django test suite
         working-directory: ./backend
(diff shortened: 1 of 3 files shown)


jperson1:
Drafting while I continue to poke at the flakey tests.

phildominguez-gsa (approved):

Issue #5550, opened 2026-04-13
[RESUB DE-DUP] *Revert* Resubmission Linking Command


jperson1:
In a world where everything goes wrong, we need to be able to revert the resubmission linking command. The command should have stored a list of changes in some way - the report_id and version, or just the report_id. This command will:

1. Accept the path to a list of changes.
2. Walk the list and reset the SACs `resubmission_meta` field.
3. Invoke a redissemination on these records to undo the change.

This is not necessarily a "panic button", it's also very useful for local testing.

Issue #5549, opened 2026-04-10
[RESUB DE-DUP] Manual Linking of Resubmissions


jperson1:
For those rare records that cannot be automatically linked, we would like to be able to do it manually.

Takes two or more records:
1. Unlinks them from wherever they are linked
    * Errors if it would orphan a record that is not passed in
2. Links them in order of their acceptance dates
3. Invokes redissemination on the records
4. The command is logged as an administrative action

Issue #5546, opened 2026-04-10
[RESUB DE-DUP] Resubmission Linking Command


jperson1:
A command to link assumed resubmissions with their earlier versions. This must be vigorously tested after the fact.

## Identifying a Resubmission
A resubmission will be identified by having identical metadata to a previous record. This data consists of the UEI, audit year, EIN, entity name, and auditee email. We believe this is sufficient to cover the vast majority of resubmissions.

## Linking Records
For each record in `dissemination_general`:
1. Filter for any records with matching metadata
2. Order the records by acceptance date
3. Add the appropriate `resubmission_meta` JSON to the SAC
    * Version number, linked records
    * Resubmission justification - FAC ADMIN CURATION ACTION
4. Invoke a redissemination on each record to publicize the change
5. LOG THE CHANGE
    * The report_id and version information should be stored
    * The log of changed records may be publicized, and we may need it to hit the "panic button" and undo the action

The action should be able to be run on a subset of `dissemination_general`; a particular record or set of records, an audit year, an audit period.

## Verification
For each resubmission chain, spot check some data. For example, if every record in a chain shares identical rows in `dissemination_federalaward`, we can say with some certainty that they were appropriately linked.

PR #5622, opened 2026-05-06
Resubmission Linking Command Updates

jperson1:
# Resubmission Linking Command Updates

I will not make such large commits a habit.

## Related tickets

* Directly related - https://github.com/GSA-TTS/FAC/issues/5546
* Directly related - https://github.com/GSA-TTS/FAC/issues/5550
* Only sort of related, but this includes the _opposite_ of https://github.com/GSA-TTS/FAC/issues/5549

## Description of changes

* The linking command:
    * Now uses equivalence to determine the chains. We may come back and use distance to catch stragglers.
    * The reviewable CSVs saw an update to enable the undo-ing, and all the files go into their own subdirectory.
* A new command to undo the linkage after the fact, just in case.
    * Uses a CSV generated from the linking command.
* A new command to wipe all resubmission metadata for an audit year or a set of specific records.
    * Very convenient while testing. Will be useful if any of these linkages proves false way down the line.
* A little Makefile change from last week.

## How to test
Switch to this branch, bring everything up normally. Ensure you have some local data to play with. Ideally, both Census and GSA records.

### The linking command
Try a few years, perhaps. The more recent years include more data. Those will take a little longer. Run the command with:
`python manage.py link_resubmissions --email {ADMIN_EMAIL} --audit_year 2020`

Check the `.md` and `.csv` files that are produced. Verify they look right. You'll see a lot of duplicate data - that's fine, and ideal, since it means the records are truly in the same resubmission chain. For the 2019-2022 range, verify there are some Census-GSA crossovers.

Press 'c' and 'enter' to continue. You'll see each record get redisseminated quickly. 

Verify a few records in search and/or through the API. You may want to re-up the materialized view to use Advanced Search. Also spot check the internal `singleauditchecklist` table. 

### The undo linking command
You've just run a few years, probably. Grab a related CSV, and run a command like:
`python manage.py undo_link_resubmissions --email {ADMIN_EMAIL} --csv curation/data/{FILENAME}.csv`

You'll see a "INFO CSV contains XYZ records." Verify it's right. Press 'c' and 'enter' to continue. You'll see each record get redisseminated quickly. 

Verify a few records after the fact. Any record that previously had "NULL" resubmission data will instead have version 0 and status "unknown_resubmission_status". If we were to redisseminate it with NULL, it would be assumed to be version 1. So we explicitly say "unknown" after taking curative action on it. 

### The reset resubmission metadata command
This command sets the resub version to 0 and the status to "RESUBMISSION_STATUS.UNKNOWN" for all records it hits. By passing `--set_most_recent True` it will instead use version 1 and "RESUBMISSION_STATUS.MOST_RECENT". It will do either a full AY or accept a list of report_id's. It's very nice for testing the above commands locally. We may eventually be using this to reset records that are incorrectly brought together - either by our own actions or by user error.

Run the command like:
`python manage.py reset_resubmission_metadata --email {ADMIN_EMAIL} --audit_year 2024`
`python manage.py reset_resubmission_metadata --email {ADMIN_EMAIL} --report_ids 2024-12-GSAFAC-0000383165 2024-12-GSAFAC-0000387718 2024-12-GSAFAC-0000398121 --set_most_recent True`

You'll see something like the following:
"INFO Found XYZ records for AY2024."
"INFO Found 3 records for report IDs: 2024-12-GSAFAC-0000383165, 2024-12-GSAFAC-0000387718, 2024-12-GSAFAC-0000398121."

Press 'c' and 'enter' to continue. You'll see each record get redisseminated quickly. Verify a few records after the fact.

## Screenshot
image


jperson1:
I've tested pretty extensively in the Preview environment, and things look good. I had to bump the memory to run the linking command - and it still couldn't get to some of the "larger" years. We may want to bump Production a bit before running the commands there. Something to keep in mind.

backend/curation/curationlib/sac_resubmission_records_postgres.py
```diff
@@ -1,53 +1,28 @@
+from django.db.models import Q
+
 from audit.models import SingleAuditChecklist
-from django.db.models import Count, Q, F
 
-# from config.settings import GSA_MIGRATION
 
+def fetch_sac_resubmission_records_postgres(AY, noisy=False):
```
phildominguez-gsa: Should this be renamed to just `fetch_sac_records_postgres` or something, since the resubmission filtering is done elsewhere now?

backend/curation/curationlib/generate_resubmission_clusters.py
```diff
@@ -15,14 +18,93 @@ class MinDist:
...
+    records = fetch_sac_resubmission_records_postgres(AY=AY, noisy=noisy)
+    sorted_sets = generate_clusters_from_records_by_distance(records, noisy=noisy)
     return sorted_sets
 
 
-def generate_clusters_from_records(records, noisy=False):
-    # For each record, compute its distance to the existing sets.
-    # If it is below the threshold, insert it into an existing set.
-    # Otherwise, insert into a new set.
+def generate_clusters_from_records_by_equivalence(records, noisy=False):
```
phildominguez-gsa: PEDANTIC ALERT (also, I know you inherited this code)

Some of these names bug me. "Cluster" instead of "chain" (only makes sense if you agree with my other sorting comment, though) and "record" instead of "audit" or "submission". For example, could this be `generate_audit_chains_by_equivalence`? Probably annoying to swap the verbiage everywhere but I figure it's worth checking and I'm down to discuss.

backend/curation/management/commands/link_resubmissions.py
phildominguez-gsa: Is the `lfilter` doing the same as [this](https://github.com/GSA-TTS/FAC/blob/3ac80f0008edae0ca982a819e71cdbbefdf7ac2c/backend/curation/curationlib/generate_resubmission_clusters.py#L91)?

backend/curation/curationlib/generate_resubmission_clusters.py
```diff
@@ -15,14 +18,93 @@ class MinDist:
 
 def generate_resbmission_clusters(AY=None, noisy=False):
     records = fetch_sac_resubmission_records_postgres(AY=AY, noisy=noisy)
-    sorted_sets = generate_clusters_from_records(records, noisy=noisy)
+    sorted_sets = generate_clusters_from_records_by_equivalence(records, noisy=noisy)
+    return sorted_sets
```
phildominguez-gsa:
```suggestion
    return generate_clusters_from_records_by_equivalence(records, noisy=noisy)
```
`sorted_sets` is a little confusing to me, since it IS sorted by AY but NOT per bucket, which we do later in `annotate_linked_reports`. I feel like we should just do all sorting in `generate_clusters_from_records_by_*`, honestly.

backend/curation/management/commands/undo_link_resubmissions.py
```diff
@@ -0,0 +1,165 @@
...
+    Records whose prior_resubmission_meta is empty were NULL before linkage.
+    They must be restored to version 0 rather than NULL,
+    because a NULL would be redisseminated as version 1.
+    """
+    if raw == "" or raw is None:
+        return {
+            "version": 0,
+            "resubmission_status": "no_resubmission_data",
+        }
+    return json.loads(raw)
```
phildominguez-gsa, line 36: Handle invalid json?

phildominguez-gsa (changes requested):

backend/curation/curationlib/sac_resubmission_records_postgres.py
jperson1: Yep!

backend/curation/curationlib/generate_resubmission_clusters.py
jperson1: I fully agree, and I'll take every opportunity to fix up the language around things. I'm trying to use "chain" pretty much everywhere. That may or may not be helpful, since they're not really "chains" until the sort and annotation is made? They are aspiring chains, and that seems... Fine. I've switched to "submission" in most places, I feel like that makes sense. I've updates some variables to use `sac` or `sacs`, which matches up with a lot of the backend code. Should be better, but I'm sure I missed stuff. Nothing broke!

backend/curation/management/commands/link_resubmissions.py
jperson1: Yes! Switched this one to a list comprehension as well.

backend/curation/curationlib/generate_resubmission_clusters.py
jperson1: Very fair. I moved the per-bucket (per-chain, now) sort over here.

backend/curation/management/commands/undo_link_resubmissions.py
jperson1: Done! Just by skipping it if an error comes back for the record. It's probably better than bailing halfway through the command and leaving things messy.

backend/curation/curationlib/generate_resubmission_chains.py
```diff
@@ -0,0 +1,146 @@
...
+
+def get_and_generate_submission_chains_by_equivalence(AY=None, noisy=False):
+    records = fetch_sac_disseminated_records_postgres(AY=AY, noisy=noisy)
+    sorted_chains = generate_submission_chains_by_equivalence(records, noisy=noisy)
+    return sorted_chains
+
+
+def get_and_generate_submission_chains_by_distance(AY=None, noisy=False):
+    records = fetch_sac_disseminated_records_postgres(AY=AY, noisy=noisy)
+    sorted_chains = generate_submission_chains_by_distance(records, noisy=noisy)
```
phildominguez-gsa:
```suggestion
    sacs = fetch_sac_disseminated_records_postgres(AY=AY, noisy=noisy)
    sorted_chains = generate_submission_chains_by_equivalence(sacs, noisy=noisy)
    return sorted_chains


def get_and_generate_submission_chains_by_distance(AY=None, noisy=False):
    sacs = fetch_sac_disseminated_records_postgres(AY=AY, noisy=noisy)
    sorted_chains = generate_submission_chains_by_distance(sacs, noisy=noisy)
```
Looks like it just returns a list of sacs

backend/curation/curationlib/sac_disseminated_records_postgres.py
```diff
@@ -0,0 +1,28 @@
+from django.db.models import Q
+
+from audit.models import SingleAuditChecklist
+
+
+def fetch_sac_disseminated_records_postgres(AY, noisy=False):
```
phildominguez-gsa: `fetch_disseminated_sacs_for_ay`? Including `postgres` doesn't seem pertinent for consumers.

backend/curation/curationlib/sac_disseminated_records_postgres.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Return all disseminated records for the given audit year.
+
+    Equivalence or distance-based clustering will decide which records actually form resubmission chains.
+    We just need the full population of disseminated records for the year.
+    """
+    records = SingleAuditChecklist.objects.filter(
```
phildominguez-gsa:
```suggestion
    sacs = SingleAuditChecklist.objects.filter(
```
And elsewhere in this method

backend/curation/curationlib/generate_resubmission_chains.py
```diff
@@ -0,0 +1,146 @@
...
+    return sorted_chains
+
+
+def generate_submission_chains_by_equivalence(records, noisy=False):
```
phildominguez-gsa: I think `records` is still just `sacs` here

phildominguez-gsa (changes requested):

backend/curation/curationlib/sac_disseminated_records_postgres.py
jperson1: Made the file just "fetch_sacs.py" for a world where we need more than one function in there

backend/audit/management/commands/reset_resubmission_metadata.py
```diff
@@ -0,0 +1,113 @@
...
+            logger.error("Exiting.")
+            sys.exit()
+
+        ok_user = User.objects.get(email=options["email"])
+        reset_count = 0
+
+        for sac in sacs:
+            sac.resubmission_meta = {
+                "version": 0,
+                "resubmission_status": "no_resubmission_data",
```
phildominguez-gsa: I feel like "no_resubmission_data" should be added to the list of resub status consts, otherwise it feels like we're tossing a random string in there. Maybe there's a more specific phrase we can use too? `data_reset` or something? Idk.

backend/curation/management/commands/link_resubmissions.py
```diff
@@ -202,25 +196,29 @@ def handle(self, *args, **options):
...
         )
 
-        export_sets_as_csv(options["audit_year"], sorted_sets, noisy=options["noisy"])
+        export_chains_as_csv(
+            options["audit_year"], sorted_chains, noisy=options["noisy"]
+        )
 
-        k = input("Review markdown and press `c` to continue...")
+        logger.info(f"Found {len(sorted_chains)} resubmission chains.")
+        k = input("Review markdown/CSV and press `c` to continue...")
```
phildominguez-gsa:
```suggestion
        k = input("Review markdown/CSV and enter `c` to continue: ")
```

backend/curation/management/commands/link_resubmissions.py
```diff
@@ -202,25 +196,29 @@ def handle(self, *args, **options):
...
-        export_sets_as_markdown(
-            options["audit_year"], sorted_sets, noisy=options["noisy"]
+        export_chains_as_markdown(
+            options["audit_year"], sorted_chains, noisy=options["noisy"]
         )
 
-        export_sets_as_csv(options["audit_year"], sorted_sets, noisy=options["noisy"])
+        export_chains_as_csv(
+            options["audit_year"], sorted_chains, noisy=options["noisy"]
+        )
```
phildominguez-gsa: Would be nice if these returned the filenames and we printed them out. When I saw this printout I wasn't sure where they went to.

phildominguez-gsa (changes requested):

backend/curation/management/commands/link_resubmissions.py
```diff
@@ -202,25 +196,29 @@ def handle(self, *args, **options):
...
 
-        k = input("Review markdown and press `c` to continue...")
+        logger.info(f"Found {len(sorted_chains)} resubmission chains.")
```
phildominguez-gsa: Exit here if 0?

phildominguez-gsa (changes requested):

backend/audit/management/commands/reset_resubmission_metadata.py
```diff
@@ -0,0 +1,113 @@
...
+                    f"resubmission metadata and will be skipped: {missing}"
+                )
+
+        if not sacs:
+            logger.info(f"No submissions with resubmission metadata found for {label}.")
+            sys.exit(0)
+
+        logger.info(f"Found {len(sacs)} submissions for {label}.")
+
+        k = input("\nPress `c` to continue...")
```
phildominguez-gsa:
```suggestion
        k = input("\nEnter `c` to continue: ")
```

backend/curation/management/commands/undo_link_resubmissions.py
```diff
@@ -0,0 +1,172 @@
...
+        rows = _load_csv(options["csv"])
+
+        # Display a little summary of what will be undone before touching any submissions.
+        report_ids = [r["report_id"] for r in rows]
+        logger.info(f"CSV contains {len(rows)} submissions.")
+        if options["noisy"]:
+            for rid in report_ids:
+                logger.info(f"  {rid}")
+
+        k = input("\nPress `c` to continue...")
```
phildominguez-gsa:
```suggestion
        k = input("\nEnter `c` to continue: ")
```

phildominguez-gsa (changes requested):

backend/audit/management/commands/reset_resubmission_metadata.py
jperson1: I've added a flag so that we either use v0 and "Unknown", or v1 and "Most Recent"

backend/curation/management/commands/link_resubmissions.py
jperson1: Done! They print the path. That can be a little confusing from something like VSCode, since the path is `src/` in the container rather than `backend/`?

backend/curation/management/commands/undo_link_resubmissions.py
```diff
@@ -0,0 +1,178 @@
...
+    Pull the JSON resub metadata from the CSV.
+
+    Submissions whose prior_resubmission_meta is empty were NULL before linkage.
+    They must be restored to version 0 rather than NULL,
+    because a NULL would be redisseminated as version 1.
+    """
+    if raw == "" or raw is None:
+        return {
+            "version": 0,
+            "resubmission_status": "no_resubmission_data",
```
phildominguez-gsa: Oh, the one-off resub status here, too

phildominguez-gsa (changes requested):

backend/curation/management/commands/undo_link_resubmissions.py
jperson1: Done, thanks!

phildominguez-gsa (approved):

Issue #5566, opened 2026-04-20
cloud.gov resource reduction - Stage 1


daryll:
With the [new pricing structure from cloud.gov ](https://cloud.gov/pricing) we should begin reducing costs.

The first stage is to remove resources that are no longer used and/or reduce redundant applications
- remove the sandbox environment entirely
- drop metabase usage to a couple environments
    - Production and staging

ccampbell-fearless:
Cloud gov usage before cleanup

```   
    "app_instances_count": 40,
    "apps_count": 36,
    "billable_app_instances_count": 40,
    "billable_services_count": 86,
    "memory_quota": 92160,
    "memory_usage": 66350,
    "name": "gsa-tts-oros-fac",
    "running_app_instances_count": 40,
    "running_apps_count": 36,
    "services_count": 86,
    "services_suite_for_pivotal_platform_count": 0,
    "spring_cloud_services_count": 0,
    "stopped_app_instances_count": 0,
    "stopped_apps_count": 0
```
Cloud gov usage after cleanup

```
    "app_instances_count": 29,
    "apps_count": 26,
    "billable_app_instances_count": 29,
    "billable_services_count": 69,
    "memory_quota": 92160,
    "memory_usage": 50392,
    "name": "gsa-tts-oros-fac",
    "running_app_instances_count": 29,
    "running_apps_count": 26,
    "services_count": 69,
    "services_suite_for_pivotal_platform_count": 0,
    "spring_cloud_services_count": 0,
    "stopped_app_instances_count": 0,
    "stopped_apps_count": 0
```

Usage reports include the following spaces:
- dev
- preview
- production
- sandbox
- sandbox-egress
- staging

PR #5635, opened 2026-05-14
added condition for deploying metabase resources

ccampbell-fearless:
## Related tickets

* https://github.com/GSA-TTS/FAC/issues/5566

## Description of changes

* Added condition for deploying metabase in staging and prod only
* Tf state migration was done locally for staging and prod resources so that existing instances will not be replaced


jperson1 (dismissed):

daryll (dismissed):
lgtm

daryll (approved):

Issue #5447, opened 2026-03-12
Trivy Scan Optimizations


ccampbell-fearless:
## Description

The current trivy scans run more than necessary. This ticket is to scope scans to only the relevant images and remove third-party container scans out of the PR checks.

## Tasks

- Remove FAC container from trivy scanning. This image is not deployed anywhere and should not be included in the scan targets
- Remove third-party container scans from PR checks since they are not modified by PRs
- Ensure job that pulls third party images and pushes to ghcr scans images successfully

PR #5659, opened 2026-05-28
cleaned up staging deployment and trivy scans

ccampbell-fearless:
## Related tickets

* https://github.com/GSA-TTS/FAC/issues/5447

## Description of changes

* renamed GH workflow job & file names
* removed FAC app from trivy scans
* consolidated workflow files


jperson1 (approved):
Looks right. Thanks!

Issue #5136, opened 2025-07-25
[VALIDATION] Require sensible auditor, auditee names


jadudm:
## problem

There are submissions to the FAC where the auditor firm name is the same as the EIN. This is not a firm name.

## solution

We should have some minimum requirements in the Python validations for some of the name fields.

- [ ] The field should be a minimum length. Assuming "A LLC" is the shortest possible firm name, we will say it must be 5 characters or longer.
- [ ] The field may not contain the EIN or UEI
- [ ] The field may not be made up entirely of digits and symbols (e.g. 0-9, -, ., _)

This ticket should be expanded with any other validations we wish to apply.

## we will know we are done when

- [ ] we have a list of validations we want to apply
- [ ] we have a validation in place
- [ ] we apply the validation to (auditee name, auditor name, auditor firm name) and any other fields that look like they should have similar constraints applied
- [ ] we have an error message that clearly guides the user to a solution

jperson1:
Triage-ing, so we can determine a list of validations before handing off for implementation.

phildominguez-gsa:
Previous character-limit-related PR: https://github.com/GSA-TTS/FAC/pull/4026. It includes a GDrive link to research findings.

daryll:
suggest more validations when working on this tix

PR #5677, opened 2026-06-04
Validation Warnings

### LINKED ISSUE #5170: Same EIN for Auditee/Auditor
## problem
Federal users have identified instances of an auditor putting their EIN into the auditee EIN field resulting in the same EIN for both auditee and auditor. 

## potential solution
Review validation options to potentially limit or block the ability to use the same EIN in auditor/auditees.
If further validations are not possible provide guidance to stakeholders to prevent user errors.

jperson1:
# Validation Warnings

Introduces the notion of a cross-validation "warning", which is not blocking for users but may let them know of any issues ahead of their final submission.

## Related tickets

* Closes https://github.com/GSA-TTS/FAC/issues/5170
* Hits a small part of https://github.com/GSA-TTS/FAC/issues/5136

## Description of changes

There are places where we the FAC may have some confidence that an input is incorrect, but can think of enough exceptional cases that we cannot block the submission. Introducing then, warnings!

Warnings are formatted in the same way as validation errors, they just have a different name and are therefore separate objects. Errors and warnings are returned separately from the full_validation functions. They are displayed in a different table from the errors. They do not prevent submissions. Their only purpose is to _warn_ a user of a potential issue.

To start, there are two:
1. Warning of duplicate EINs for auditees and auditors. There are potential exceptions for large entities like states, so it is a warning and not an error.
2. Warning of duplicate names in a number of fields in the gen info. There are potential exceptions here, because names are hard. For now this only acts on four fields, but it can be safely extended.

## How to test

* Switch to this branch, run normally.
* Navigate to or create a submission.
* Fill out the general information in order to get the two new warnings to display
    * The duplicate EIN one, by putting the same EIN into both inputs
    * The duplicate names one, by putting duplicate names into two or more of these four fields: auditee_name, auditee_contact_name, auditor_contact_name, auditor_firm_name

## Screenshots and recordings

image


backend/audit/models/models.py
```diff
@@ -248,6 +248,9 @@ def disseminate(self):
         ETL.
         """
         try:
+            # This will force a re-disseminated record to regenerate the cog/over. Perhaps we should do this always?
+            self.cognizant_agency = None
+            self.oversight_agency = None
```
phildominguez-gsa: This seems unrelated to the rest of the PR, no? Could we do it separately?

phildominguez-gsa (changes requested):

backend/audit/models/models.py
jperson1: Oops, sure is. Sorry! That was for debugging earlier in the week.

phildominguez-gsa (dismissed):

phildominguez-gsa (approved):

PR #5691, opened 2026-06-08
[BUG] Selecting Date from Calendar

RanyeM:
## Related tickets
* https://github.com/GSA-TTS/FAC/issues/5542

## Description of changes
* Fixed an issue where the **Validate UEI** button could remain disabled after users selected fiscal period dates from the USWDS date picker.
* Updated `validate.js` to guard against unknown validator keys so validation does not crash unexpectedly.
* Updated `check-ueid.js` to scope error checks to input validation errors only.
* Updated `check-ueid.js` to listen for the USWDS date picker’s custom event so button state updates when dates are selected from the calendar.
* Ran `collectstatic` so the compiled static files are available in the Docker container.

## How to test
* Start the app locally.
* Navigate to the Auditee Information page.
* Enter a valid UEI.
* Select fiscal period start and end dates using the calendar date picker.
* Confirm the **Validate UEI** button becomes enabled without manually editing or blurring the date fields.
* Confirm the button remains disabled when required fields are empty or date validation is invalid.
* Confirm UEI validation still proceeds successfully after the button is enabled.

## Screenshots and recordings
* Confirmed locally that the button enables once both `filled: true` and `valid: true` are met after selecting dates from the date picker.
Screenshot 2026-06-08 at 12 45 41 PM


backend/static/js/check-ueid.js
```diff
@@ -41,13 +41,22 @@ function requiredFieldsFilled() {
...
-  const inputsWithErrors = document.querySelectorAll('[class *="--error"]');
+  const inputsWithErrors = FORM.querySelectorAll('input[class*="--error"]');
   return inputsWithErrors.length === 0;
 }
 
 function updateValidateButtonState() {
-  const btn = document.getElementById('continue'); // bottom "Validate UEI" button
+  const btn = document.getElementById('continue');
+
+  console.log('BUTTON CHECK', {
```
jperson1: We can drop this log, now that the issue's been worked out

(Or comment it, in case this button gives us more issues in the future)

jperson1 (approved):

PR #5705, opened 2026-06-12
Prevent Unchanged Resubmissions

### LINKED ISSUE #5217: [RESUB] Prevent Unchanged Resubmissions
If a user has gone through the process of resubmitting, but there are no detectable changes from v1 to v2, should we prevent their submission? Is there ever a case where a user _should_ resubmit with no changes to the record?

If we want to prevent this case, we need to have full confidence in our ability to find all the meaningful differences between the records.

A simple implementation:
In the resubmission cross-validations, we can check the two SACs. We first omit SAC-unique data, like the row ID, report ID, dates, transition names, and certifications. Then, we hash the two SACs and compare. If they are the same, the SACs are identical and the resubmission should be disallowed.

phildominguez-gsa:
## Related tickets

* Closes https://github.com/GSA-TTS/FAC/issues/5217

## Description of changes

* Adding cross-val `check_resubmission_has_changes()` that errors if `compare_with_prev()` detects no changes between a resubmission and the original
* `compare_with_prev()` now always returns a tuple by using `None` for report_ids when appropriate
* Updating resubmission-full since they used be identical. Now the resubs have a randomized phone number.
* Added a blurb to the comparison page when no changes are found so that they're aware it isn't going to pass validation

## How to test

* Have a submission disseminated and begin the normal resub process without making any changes
* The comparison page should show no changes
* Presubmission validation should fail
* Now, go back and change a field somewhere
* The comparison page should show changes
* Presubmission validation should pass
* full-resubmission passes

## Screenshots and recordings

image
image

### DIFF
--- a/backend/audit/templates/audit/comparison_sections_only.html
+++ b/backend/audit/templates/audit/comparison_sections_only.html
@@ -18,7 +18,7 @@
     <div class="grid-row width-full">
       <div class="tablet:grid-col">
         <h2>There are no changes between Single Audit Reports</h2>
-        <p>Note that identical resubmissions are not permitted, so pre-submission validation will fail in its current state.</p>
+        <p>Note: Identical resubmissions are not permitted. Pre-submission validation will fail for this submission in its current state.</p>
       </div>
     </div>
   {% endif %}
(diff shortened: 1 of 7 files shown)


backend/audit/templates/audit/comparison_sections_only.html
LynnMHouston, line 1: Could we change "Pre-submission validation will fail for this submission in its current state" to "Pre-submission validation will fail for this submission unless changes are made."

phildominguez-gsa: Done!

RanyeM (approved):
LGTM!

PR #5713, opened 2026-06-16
Setting Django timezone to America/New_York

### LINKED ISSUE #5487: [BUG] Daylight Savings Timezone Error
It is currently 1:32 PM. The app is not accounting for daylight savings in the checklist, and displays 12:32 PM. 

Image

phildominguez-gsa:
## Related tickets

* Closes https://github.com/GSA-TTS/FAC/issues/5487

## Description of changes

* Setting Django timezone to America/New_York to fix off-by-one-hour issue

## How to test

* Start a submission as normal
* Check a timestamp under a completed submission checklist entry. It should show the correct time and not an hour too early.
image

### DIFF
--- a/backend/config/settings.py
+++ b/backend/config/settings.py
@@ -211,7 +211,7 @@ AUTH_PASSWORD_VALIDATORS = [
 
 LANGUAGE_CODE = "en-us"
 
-TIME_ZONE = 'America/New_York'
+TIME_ZONE = "America/New_York"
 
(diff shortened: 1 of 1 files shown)


jperson1 (approved):
Making a note to myself (and others, I suppose) that we should revisit some of the timezone shenanigans that come with a wider user base and daylight savings time. This fixes the issue though!

PR #5717, opened 2026-06-17
Creating dissemination_unified model and disseminating

### LINKED ISSUE #5680: [MV] Create new combined table and populate during dissemination
* Create a new model in `backend/dissemination` called `DisseminationUnified`. It will be almost identical to `DisseminationCombined`, but the `Meta` options should be omitted.
* Update `intake_to_dissemination.py` to insert into `unified` by "joining" the rows that are generated for relevant tables
  * Tables to be joined: `general_info`, `federal_award`, `finding`, and `passthrough`
* The data should appear identical to the `combined` MV
* Some preliminary profiling: Compare how long it takes to disseminate a submission or a sanitized dump

phildominguez-gsa:
## Related tickets

* Closes https://github.com/GSA-TTS/FAC/issues/5680

## Description of changes

* Creating `dissemination_unified` table and populating during dissemination
  * Having to copy/paste all the fields AGAIN from gen/awards/findings/PTs did not feel good, so I found a way to import them from the other models. This is nice because if any of the fields we care about change, `unified` will just pick it up. I noticed `combined` missed out on some model tweaks in the past because of the missed duplicated code, too.
    * Exceptions to this are `aln` (because it's unique to `combined`/`unified`) and `report_id` (because of foreign key weirdness)
    * Indexes can still be added later via `class Meta: ... indexes = [ ... ]`
* Fixing ADR filename :(

## How to test

* `make dfs` or whatever you like to run the new migration
* You should be able to see `dissemination_unified` in DBeaver. You might have to right click -> refresh Tables.
* Create a submission however you like
* Create the combined materialized view: `docker compose run --rm web python manage.py materialized_views --create`
* The submission's entries between `dissemination_unified` and `dissemination_combined` should match, though the field ordering is different.
* You can try search using it by changing `DisseminationCombined` to `Unified` [here](https://github.com/GSA-TTS/FAC/blob/main/backend/dissemination/search.py#L67) and adding the import
* I used `manage_local_data.bash` to load and disseminate 20k submissions successfully as well. Loading all the SACs takes so long I wouldn't blame you if you weren't up for that. Disseminating didn't take long at all though, so that's good.


RanyeM (approved):
LGTM

jperson1 (approved):
Excellent, thanks! Happy to see this moving along.

daryll (approved):
LGTM

Issue #5481, opened 2026-03-19
Backups Retention


ccampbell-fearless:
## Description
Currently, backup files are named using a month-day-hour convention, which causes backups to be overwritten on a yearly basis. While this could have been an intentional 1 year retention strategy, it isn’t the most intuitive and increases complexity when navigating backup history. To eliminate this, an S3 bucket retention policy (90 days) should to be established to explicitly determine how long backups are retained. 

## Solution

Add a gh action that applies a bucket retention policy via the aws cli. This approach is was decided on since the cloud.gov S3 broker does not support this level of bucket configuration. This workflow should run on a schedule (weekly??) to ensure the retention policy remains in place and is not inadvertently removed.

## Tasks/AC:
* A gh action workflow is created that applies a retention policy (90 days) to the backups bucket via the aws cli
* The workflow runs on a defined schedule
* The retention duration is defined and documented

*Note*: The naming convention will be address in a separate issue

PR #5767, opened 2026-07-06
imported prod resource into stae and documented S3 lifecycle policy

ccampbell-fearless:
## Related tickets

* https://github.com/GSA-TTS/FAC/issues/5481

## Description of changes

* added documentation for lifecycle policy
* imported backups s3 bucket into prod terraform state


jperson1 (approved):
Woo!

Issue #5615, opened 2026-05-04
Automate ALN Updates Using SAM.gov Assistance Listings API and Autogenerated Test Workbooks


LynnMHouston:
## Problem

The current process for updating Assistance Listing Numbers (ALNs) relies on manually downloading separate CSV files for active and inactive listings from SAM.gov and merging them into cfda-lookup.csv.

Additionally, Cypress test workbooks are manually updated when schema changes occur, which introduces extra overhead and risk of inconsistency.

This process is time-consuming, error-prone, and not easily reproducible.

## Proposed Changes

Introduce automation for both ALN updates and test workbook generation:

- Add scripts/fetch_alns_from_sam.py to retrieve ALNs from the SAM.gov Assistance Listings API (/assistance-listings/v1/search).
- Generate active-alns.csv and inactive-alns.csv automatically.
- Integrate the fetch step into the Makefile (make fetch_alns).
- Update make all to include the fetch step before merging.
- Add build_test_workbooks.py to auto-generate Cypress test workbooks from schema outputs and fixture data.
- Integrate test workbook generation into make all.

## Why This Matters

These updates will:

- Eliminate manual ALN download and maintenance.
- Ensure ALN data is always current and consistent with SAM.gov.
- Remove the need to manually maintain Cypress test workbooks.
- Improve reproducibility of schema builds and test artifacts.
- Reduce risk of stale or inconsistent data across the system.

## Files Affected

- backend/schemas/scripts/fetch_alns_from_sam.py
- backend/schemas/scripts/build_test_workbooks.py
- backend/schemas/source/data/test_workbook_fixtures.json
- backend/schemas/Makefile
- backend/schemas/source/data/aln_csvs_to_be_merged/ (generated files)
- backend/cypress/fixtures/test_workbooks/ (generated files)

## Acceptance Criteria

- [ ] make fetch_alns successfully retrieves ALNs from SAM.gov API.
- [ ] active-alns.csv and inactive-alns.csv are generated automatically.
- [ ] Existing merge_aln_csvs.py continues to function without modification.
- [ ] make all runs end-to-end without manual ALN file updates.
- [ ] Cypress test workbooks are automatically generated during make all.
- [ ] Test workbooks correctly include fixture data.
- [ ] Generated cfda-lookup.csv matches expected structure and content.

## Additional Context

This work leverages the SAM.gov Assistance Listings API:

https://api.sam.gov/assistance-listings/v1/search

[...]

PR #5775, opened 2026-07-08
updated readme to denote deprecation of manual aln steps

annie-maria:
## Related tickets

* https://github.com/GSA-TTS/FAC/issues/5615

## Description of changes

* Simply updated the readme to deprecate the manual running of the aln listing files. They are able to be extracted from the api endpoint and the execution process already doesn't use it. 

## How to test

* Only a readme change so no need for tests.


backend/schemas/README.md
LynnMHouston, line 1: Instead of this: "These steps do not need to be followed due to the above directions doing the work programmatically."

Maybe let's say something more like "These steps are no longer necessary because the script now calls the API and automatically downloads the required files."

LynnMHouston (approved):

Issue #5521, opened 2026-04-02
Replicate 499s in testing environment


phildominguez-gsa:
When we've tried to update Gunicorn past v22 we quickly start to see 499 errors in prod. We'd like to try to replicate these errors in a testing environment so we can safely figure out a fix, as opposed to experimenting in prod and breaking it a bunch. So far, we haven't had any luck triggering the 499s. This includes putting the app and API under load with multiple instances running in preview, similar to prod.

Idea: Can we run a private instance of the app in prod we can test on?

PR #5780, opened 2026-07-10
Forcing setuptools v83

### LINKED ISSUE #5743: Update pip group
* https://github.com/GSA-TTS/FAC/pull/5558
* sqlparse from 0.5.1 to 0.5.4
* setuptools from 75.2.0 to 78.1.1

phildominguez-gsa:
## Related tickets

* Closes https://github.com/GSA-TTS/FAC/issues/5743

## Description of changes

* The version of setuptools used by the app is pretty old. Normally it shouldn't be in a `requirements` file at all, but we're [forced to use an old version of Gunicorn](https://github.com/GSA-TTS/FAC/issues/5521) and it's old `zope-*` dependencies require it.
* The [Dependabot PR](https://github.com/GSA-TTS/FAC/pull/5558) also updates `sqlparse`, but we [already did that](https://github.com/GSA-TTS/FAC/blob/main/backend/requirements/requirements.in#L43)

## How to test

* App runs as normal

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -1574,10 +1574,11 @@ zope-interface==7.1.0 \
     # via gevent
 
 # The following packages are considered to be unsafe in a requirements file:
-setuptools==75.2.0 \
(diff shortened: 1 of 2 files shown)


ccampbell-fearless (approved):
LGTM

PR #5830, opened 2026-07-31
Added condition to check for service-account for local development

ccampbell-fearless:
## Description of changes

* There is a 500 error when running `rc admin service-account create` when the service account already exists. This PR adds a check to determine if that command should be run.


jperson1 (approved):
Looks great!
Image

PR #5833, opened 2026-08-03
Showing tribal step for all org types

### LINKED ISSUE #5604: [Submission] Display a Greyed-Out Tribal Data Release Link for All
Consider:
1. I am the representative of a tribal entity, which is also a non-profit or place of education.
2. I begin a record as entity-type "non-profit".
3. I do not see the tribal data release form, and have no way of knowing it exists from the web app.
    * The form is covered in the submission guide, but in a perfect world our process is so smooth that it is not necessary ;)
4. I submit a regular-public record that I would otherwise have chosen to be tribal-private

To help prevent this scenario, we'd like to show a greyed out tribal data release link for all users. We'll update the language to limit any confusion for non-tribal entities.

phildominguez-gsa:
## Related tickets

* Closes https://github.com/GSA-TTS/FAC/issues/5604

## Description of changes

* The checklist now displays the tribal release item for ALL submission types, but the link to access it is only enabled for tribal. I also tweaked the text to explain what the deal is.
* Creating a new template for selecting org type, since it's the same as the eligibility page's
  * Checking for report_id to determine if a required field `*` should be displayed

## How to test

* Submitting works as normal, including tribal
* Non-tribal
  * The tribal release step should be greyed out with appropriate text
  * Link for gen info works
* Tribal
  * The tribal release step should be clickable with the appropriate text

## Screenshots and recordings

In gen info:
image
For non-tribal:
image
For tribal (unchanged):
image


LynnMHouston:
Wording looks good!

jperson1 (approved):
Always appreciate the general cleanup. Looks good, thanks!

Issue #5800, opened 2026-07-16
[EIN VALIDATION] Validation Warning on "New" EINs


jperson1:
A user creating a new submission for an entity that has submitted in previous years should be warned when their EIN does not match up with a previous year. A previous years submission can be found by filtering `dissemination` for fully accepted (not in progress or deprecated) records with an identical UEI.

This is a beginning step towards further EIN validations.

## Acceptance Criteria
1. A new warning is added, matching our current cross validation formats. Start [here](https://github.com/GSA-TTS/FAC/blob/736f7e346e5b04add7cfec963004c506019d192c/backend/audit/cross_validation/__init__.py#L123-L126)
2. Warning text presented to the user is written or otherwise approved by our user managing folks, and contains both the old and new EIN for reference
3. A set of tests are added for the isolated validation function
4. The warning does not otherwise block a user from submitting

PR #5837, opened 2026-08-03
Rm/ein cross validation

RanyeM:
## Related tickets

https://github.com/GSA-TTS/FAC/issues/5800

## Description of changes

* Added a cross-validation warning when an Auditee EIN does not match the EIN from the most recent submission for the same UEI.
* Added an inline warning on the General Information page so users receive immediate feedback while entering an EIN, without needing to continue to the next page.
* Shared the EIN lookup and warning message logic through `Util.get_previous_ein_warning()` to eliminate duplicated logic between the UI warning endpoint and cross-validation.
* Added automated tests covering matching, mismatched, missing, and multiple previous submissions.
* Updated the frontend to only validate once a complete 9-digit EIN has been entered.

## How to test

* Start the application and navigate to the General Information page of a report submission.
* Enter an EIN that matches the most recent submission for the same UEI and verify no warning is displayed.
* Enter a different valid 9-digit EIN for a UEI with a previous submission and verify the warning appears below the EIN field.
* Verify no requests are sent while typing partial EIN values (fewer than 9 digits).
* Continue to the next page and verify the same warning is returned by cross-validation.
* Run the automated tests:
  ```bash
  python manage.py test audit.cross_validation.test_check_previous_ein
  
 ## Screenshots and recordings <!-- Optional for UI work. If there are none, delete this section. --> 
<img width="1499" height="827" alt="Screenshot 2026-08-04 at 12 39 26 PM" src="https://github.com/user-attachments/assets/a4ef3fe5-a9ec-4f3a-95fc-ad85a68cf788" />

### DIFF
--- a/backend/audit/cross_validation/check_previous_ein.py
+++ b/backend/audit/cross_validation/check_previous_ein.py
@@ -4,7 +4,7 @@ from audit.utils import Util
 def check_previous_ein(sac, sar=None):
     """
     Warn when the EIN for a new submission does not match the EIN
-    from a previous fully accepted dissemination record with the same UEI.
+    from the previous fully accepted dissemination record with the same UEI.
     """
     general_information = sac["sf_sac_sections"].get(
         "general_information",
--- a/backend/audit/cross_validation/check_previous_ein.py
+++ b/backend/audit/cross_validation/check_previous_ein.py
@@ -1,0 +1,25 @@
+from audit.utils import Util
+
+
+def check_previous_ein(sac, sar=None):
+    """
+    Warn when the EIN for a new submission does not match the EIN
+    from a previous fully accepted dissemination record with the same UEI.
+    """
+    general_information = sac["sf_sac_sections"].get(
+        "general_information",
+        {},
+    )
+
+    current_uei = general_information.get("auditee_uei")
+    current_ein = general_information.get("ein")
+
+    warning = Util.get_previous_ein_warning(
+        current_uei,
+        current_ein,
+    )
+
+    if not warning:
+        return []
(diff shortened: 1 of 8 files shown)


backend/audit/cross_validation/check_previous_ein.py
```diff
@@ -0,0 +1,25 @@
+from audit.utils import Util
+
+
+def check_previous_ein(sac, sar=None):
+    """
+    Warn when the EIN for a new submission does not match the EIN
+    from a previous fully accepted dissemination record with the same UEI.
```
phildominguez-gsa:
```suggestion
    from the previous fully accepted dissemination record with the same UEI.
```
Just to make sure it doesn't sound like _all_ previous submissions are checked

backend/audit/utils.py
```diff
@@ -92,10 +92,56 @@ def validate_redirect_url(url):
...
+            current_ein,
+        )
+
+        if not previous_ein:
+            return None
+
+        return (
+            f"The EIN entered for this submission ({current_ein}) "
+            f"does not match the EIN used in a previously accepted "
+            f"submission ({previous_ein}) for this UEI. "
```
phildominguez-gsa: I feel like we should include the report ID for the submission we're referring to so they don't have to hunt it down

phildominguez-gsa (changes requested):

backend/audit/utils.py
RanyeM: I like this idea!

phildominguez-gsa (approved):

PR #5846, opened 2026-08-05
Added terraform variable for metabase db plan

ccampbell-fearless:
## Description of changes

* Metabase was previously using the same db instance type as the FAC db. This PR adds a new variable allowing for different RDS instance types across the different services


jperson1 (approved):
Metabase DB plan was updated manually, and has completed the switch back to "redundant".

Looks good!

*Maybe we don't want those screenshots in GH. Not sure, just removing it in case.*

PR #5854, opened 2026-08-06
Updating linter packages for security findings

phildominguez-gsa:
## Description of changes

* We have a bunch of linter security findings I've been ignoring/dismissing. They just keep coming though, and we haven't updated the linter stuff in a while, anyway.
* I confirmed with the robot that `stylelint-config-scss` is not only redundant, it's an empty repo, so I removed it
* Updating `flat` import path
* I was getting an error that "yargs parser" requires Node 20, so I bumped that in the PR check
* Fixing `no-useless-assignment` errors for `key` and `row_array`
* If you `npm run build` there are some USWDS warnings, which we will update later

## How to test

* Linters still pass
* `npm run build`
* Error table still displays after a bad workbook upload, which you can see by uploading the wrong workbook
* Appending contact fields still works


RanyeM (approved):
LGTM

jperson1 (approved):

PR #3247, opened 2024-01-20
Jadudm/python deps 0120

jadudm:
This PR addresses a number of dependabot and related issues. 

* Bump `cryptography` to 41.0.6 or greater. https://github.com/GSA-TTS/FAC/security/code-scanning/1018
* Update aws-cli to v2. Should address https://github.com/GSA-TTS/FAC/security/code-scanning/865
* Update to `python-3.11-slim`, to enable a number of package updates, hopefully `pip` https://github.com/GSA-TTS/FAC/security/code-scanning/865
* `pycryptodome` update to at least 3.19.1 https://github.com/GSA-TTS/FAC/security/code-scanning/1119, https://github.com/GSA-TTS/FAC/security/code-scanning/1120
* Update pypdf to 3.17.0 or greater https://github.com/GSA-TTS/FAC/security/code-scanning/867
* urllib to 2.0.7 or greater https://github.com/GSA-TTS/FAC/security/code-scanning/858
* Django updated to v5, https://github.com/GSA-TTS/FAC/security/code-scanning/868
* gitpython to 3.1.41 or greater https://github.com/GSA-TTS/FAC/security/code-scanning/1122

Passes E2E Cypress tests and `make docker-test`.

## PR checklist: submitters

- [x]   Link to an issue if possible. If there’s no issue, describe what your branch does. Even if there is an issue, a brief description in the PR is still useful.
- [ ]   List any special steps reviewers have to follow to test the PR. For example, adding a local environment variable, creating a local test file, etc.
- [ ]   For extra credit, submit a screen recording like [this one](https://github.com/GSA-TTS/FAC/pull/1821).
- [x]   Make sure you’ve merged `main` into your branch shortly before creating the PR. (You should also be merging `main` into your branch regularly during development.)
- [x]   Make sure that whatever feature you’re adding has tests that cover the feature. This includes test coverage to make sure that the previous workflow still works, if applicable.
- [x]   Make sure the full-submission.cy.js [Cypress test](https://github.com/GSA-TTS/FAC/blob/main/docs/testing.md#end-to-end-testing) passes, if applicable.
- [x]   Do manual testing locally. Our tests are not good enough yet to allow us to skip this step. If that’s not applicable for some reason, check this box.
- [x]   Verify that no Git surgery was necessary, or, if it was necessary at any point, repeat the testing after it’s finished.
[...]

PR #5861, opened 2026-08-07
Python dependency updates for findings

phildominguez-gsa:
## Related tickets

* cyptography: https://github.com/GSA-TTS/FAC/security/dependabot?q=is%3Aopen+cryptography+
* aiohttp: https://github.com/GSA-TTS/FAC/security/dependabot?q=is%3Aopen+aiohttp+
* gitpython: https://github.com/GSA-TTS/FAC/security/dependabot?q=is%3Aopen+gitpython+

## Description of changes

* Removing gitpython from requirements. Bandit used to use a bad version so [we had forced it](https://github.com/GSA-TTS/FAC/pull/3247#:~:text=gitpython%20to%203.1.41%20or%20greater), but bandit doesn't use it anymore.
* I commented out [census_historical_migration/test_end_to_end_core.py](https://github.com/GSA-TTS/FAC/blob/main/backend/census_historical_migration/test_end_to_end_core.py)
  * [workbook_builder.py](https://github.com/GSA-TTS/FAC/blob/main/backend/census_historical_migration/workbooklib/workbook_builder.py#L2) uses [fs](https://github.com/PyFilesystem/pyfilesystem2), which tries to access `pkg_resources` from `setuptools`. This was removed from `setuptools` in v70. Instead of letting this test we don't care about anymore hold us hostage, I think it best to skip it.

## How to test

* App works as normal


jperson1 (approved):
Sometimes, you just have to banish the problem code. I appreciate the included explanation!

PR #1819, opened 2023-08-14
Sk/cog over assignment

gsa-suk:
This is a draft PR.
Completed 1462, 1699, 1738, 1762, 1811, 1812, 1813

PR #5867, opened 2026-08-11
Python dependency updates and cleanup

phildominguez-gsa:
## Related tickets

* https://github.com/GSA-TTS/FAC/security/dependabot/640
* https://github.com/GSA-TTS/FAC/security/dependabot/307
* https://github.com/GSA-TTS/FAC/security/dependabot/352

## Description of changes

* `pypdf==6.15.0` for finding
* Overriding `idna` and `python-dotenv` for findings
* `jsonnet==0.22.0` because of `pkg_resource` import issue
* Making an "Overrides" section in the reqs file. This way when we eventually get rid of a dependency we know what sub-dependency overrides we can get rid of.
* Removing these. I also confirmed they had no "via"s in the `requirements.txt` besides the `in` file, meaning they weren't there as overrides.
  * `peewee`: It was used at some point for Census stuff but was later considered dead and [deleted](https://github.com/GSA-TTS/FAC/pull/4109)
  * `cryptodome`: We already have `cryptodomex`, doesn't make sense to [add both](https://github.com/GSA-TTS/FAC/pull/3247/changes#diff-9cd5cf128c5b9f621ae51d2a84f3d69b7c31676db4c6e69a574a1a7235cc8519R25). We have a single import in [oidc.py](https://github.com/GSA-TTS/FAC/blob/main/backend/djangooidc/oidc.py#L9).
  * `python-slugify` and `types-python-slugify`: We used to use it to [slugify file names](https://github.com/GSA-TTS/FAC/pull/800/changes#diff-ee11638a6683bf5050aa4a4cc679ae4ae767aa738864658283a06bda523299faR114).
  * `sqlalchemy`: It was introduced [here](https://github.com/GSA-TTS/FAC/pull/1819). I see no uses of it anymore.
  * `uritemplate`: It was added [here](https://github.com/GSA-TTS/FAC/pull/118/changes#diff-9cd5cf128c5b9f621ae51d2a84f3d69b7c31676db4c6e69a574a1a7235cc8519) but not even used
  * `xkcdpass`: It was added [here](https://github.com/GSA-TTS/FAC/commit/8cd0fcbeac5d1b674b084856686fdfb5a3f9d9c6) for no apparent reason
  * `pyaml`: It was added [here](https://github.com/GSA-TTS/FAC/pull/118/changes) but again I see no use for it

## How to test

* `make docker-first-run && docker compose up`
* App works as normal


jperson1 (approved):
Great to clean up. I remember a handful of these being used briefly (uritemplate), or _maybe_ being used for a future feature (pyyaml), or just for local testing (xkcdpass). We'll bring them back if it matters!

Issue #5847, opened 2026-08-05
[INFRA] Remove the Pre-Deploy Backups


jperson1:
## Context
We take fresh backups every two hours or so. We also take a fresh backup as part of our production deployment process. This isn't a bad thing by any means, but it does raise some questions/concerns:

1. If a scheduled backup is running, the pre-deploy backup can run at the same time, which slows down both.
2. We run deploys midday, which means the two backups are somewhat likely to be running at the same time.
3. If a scheduled backup came in within the last hour or two, do we need a fresh backup before the deploy?
    * In a worst-case scenario, we run a botched migration on the deploy
    * We would have to load a backup and go through the full incident process anyway
4. Terraform plans require admin approval now, so catastrophic Terraform issues are much easier to catch than oh-so-long ago.
 
We think the pre-deploy backups are more a detriment than a boon at this point.

## Acceptance Criteria:
1. The production deploy action does not run a full backup before deploying.
2. The deploy step sits behind the Terraform deploy step, instead of in parallel with it.

PR #5869, opened 2026-08-11
removed database backup before prod deployment

ccampbell-fearless:
## Related tickets

* https://github.com/GSA-TTS/FAC/issues/5847

## Description of changes

* removed pre deploy database backup since backups are scheduled
* added workflow job dependency


jperson1 (approved):
Temp removal of the Terraform apply, to un-stick some planned actions we have upcoming.

Issue #5638, opened 2026-05-14
[RESUBv2] Disallow update of PDF for modification


marissahenderson:
If the user is doing a SF-SAC Data Collection form modification we should copy over their PDF from their previous submission, show it to them greyed out and disallow them to upload or change the PDF. We should also add a blurb that directs them back to the resubmission start to go through the full resubmission if they have updates to their PDF.

daryll:
- we should add a cross-validation check:
  - on whether theres a discrepancy on what the user say they did vs what they actually did (uploaded or not)
  - if the PDFs are the same

PR #5870, opened 2026-08-12
Disallow PDF Updates for SF-SAC Only Resubs

jperson1:
# Disallow PDF Updates for SF-SAC Only Resubs

## Related tickets

* https://github.com/GSA-TTS/FAC/issues/5638

## Description of changes

Overall - this is just for PDF upload behavior. when a user is making a SF-SAC change _only_, the previous SAR file is copied into the current record. Otherwise, they have to upload one themselves.

In rough order of importance:
* When the resubmission action for is submitted, it deletes the current PDF or copies in a fresh PDF as appropriate
* The PDF upload page no longer gives the option to copy the previous report
* The PDF part of the submission progress checklist will be disabled when appropriate. There's a little note for folks who may be confused at that one
* I added resub meta to the BabelFish naming scheme. We don't use it in exactly the same way, but I think it would be a good idea to, so I've started that process.
* A new SubmissionEvent for updating the resubmission action form.
    * This is what causes the migration
* Light test fixes. Something got them to act up, I'm not sure how those ones were passing as-is.
* Some light cleanup, typos, the works.

## How to test

1. Switch to this branch, run normally.
2. This is best done on _two_ submissions, to ensure that they both have PDFs in your local file system.
    * Create and submit an audit.
    * Make sure your RustFS is working properly, so that the PDF is sitting pretty in there.
    * Create a resubmission of your first submission.
3. On creating the resubmission, choose the "SFSAC_ONLY" option. Make sure the old PDF was copied into your new resubmission, and that the SAR form is disabled.
4. Switch the resubmission to the "AUDIT_PDF" option, in the resub action form. Make sure the current SAR (the one that was copied) is removed, and that the SAR form is enabled.
5. Verify that the SAR form works alright, and that the "use the old report" button and text is all gone.
6. Maybe have a quick crawl through the comments and text changes for validity.

## Notes/Questions:
1. Is it fine that the 503 page specifies "maintenance" as a possible reason for the error?
2. Do we want to delete the actual PDF when the SAR row is removed? This was not the case previously, and may have been an oversight. I suspect we do want to remove them, but we may also want to take a "flag for removal" approach as with the SAC rows.


backend/api/views/access.py
```diff
@@ -110,6 +111,25 @@ def access_and_submission_check(user, data):
...
+            == RESUBMISSION_ACTION.SFSAC_ONLY
+        ):
+            try:
+                copy_previous_report_data(
+                    previous_report_id=previous_report_id,
+                    current_sac=sac,
+                    current_audit=audit,
+                    user=user,
+                )
+            except Exception as err:
```
phildominguez-gsa: Not big on the catch-all exceptions. If specific exceptions can happen in `copy_previous_report_data` I'd prefer to have them caught in there. Also if this copying DOES fail, do we 500 and bail on the whole thing? They wouldn't be able to upload unless they changed the resub action, I think?

phildominguez-gsa (changes requested):

phildominguez-gsa (dismissed):

phildominguez-gsa (approved):

PR #5875, opened 2026-08-13
Unifying search

### LINKED ISSUE #5683: [MV] Remove basic search
Once we're confident in the new table's performance, we can remove basic search and only have advanced search available. Effectively, "basic search" will be gone and "advanced search" just becomes "search".
* Remove all related views, tests, etc., for basic search
* The new (advanced) search should now live at `dissemination/search/`, where basic search used to be
  * Make the `dissemination/search/advanced/` URL path redirect to `dissemination/search/` so that user bookmarks and whatnot don't break.

This change should be communicated with users.

Also, update any instructions and documents related to basic/advanced search. These can be found on the search pages themselves and on the static site. Some I've found:
* [FAQ page](https://support.fac.gov/hc/en-us/articles/29043839561869-I-just-submitted-my-audit-report-Why-can-t-I-download-my-report-from-Search-Is-there-an-error-with-my-submission)
* https://www.fac.gov/search-resources/filters/
* https://www.fac.gov/search-resources/ also seems weirdly empty
* Look for others!

This would also close out https://github.com/GSA-TTS/FAC/issues/5592

NOTE: Fix the foreign key -> string conversion bug that makes clicking the eyeball icon error out

### LINKED ISSUE #5684: [MV] Remove SF-SAC delay tooltip
Remove the tooltip/warning that says SF-SACs can take 24h to appear

Image

phildominguez-gsa:
Waiting on stakeholder communication before releasing this.

Static site PR [here](https://github.com/GSA-TTS/FAC-transition-site/pull/276)

## Related tickets

* Closes https://github.com/GSA-TTS/FAC/issues/5683
* Closes https://github.com/GSA-TTS/FAC/issues/5684

## Description of changes

* Removing basic search
* Advanced search is now just "search" using the unified table
* Migration to change `report_id` field to `report` since it's a foreign key, and Django returns it as a full General object in the results. It also makes `report_id` a field automatically, so this plays nicely whether General or Unified is used for searching.
* Tests now use unified instead of the materialized view
* Updating footer
* Rerouting deprecated `/search/advanced` to `/search` to bookmarks and whatnot don't break
* Removing the SF-SAC download-not-available tooltip, since they should always be available now

## How to test

* Head to http://localhost:8000/dissemination/search/
* Search should work as normal
* http://localhost:8000/dissemination/search/advanced reroutes to http://localhost:8000/dissemination/search/
* Buttons to view report summaries and downloads should work
* Footer search link works


jperson1 (dismissed):
This looks good to me. After much effort, thank you!

jperson1 (approved):

PR #5886, opened 2026-08-17
Updated app deployment workflow to use existing reusable workflow

ccampbell-fearless:
## Description of changes

* The manual app deployment workflow contained jobs that are no longer necessary. Instead of removing the jobs, this PR refactors the workflow to call the existing reusable workflow instead, eliminating duplicated code and keeping the logic in one place.


daryll (approved):
lgtm!

Issue #5685, opened 2026-06-05
[MV] Measure improved search and API performance


phildominguez-gsa:
Load a sanitized dump and compare `dissemination_combined` search times with `dissemination_unified`

PR #5929, opened 2026-09-04
Searching on General when possible for faster searches

phildominguez-gsa:
## Related tickets

* Related to https://github.com/GSA-TTS/FAC/issues/5683
* Related to https://github.com/GSA-TTS/FAC/issues/5685

## Description of changes

* When testing Unified search performance, it dawned on us that if you use advanced search to do not-actually-advanced searches, then it will be way slower than doing a basic search. This is because basic search hits the much smaller General table. This isn't terrible for the current app, but because we're removing basic search, this will soon be not great.
* This change checks to see if any advanced/non-basic/non-General fields are actually being searched on before querying DisseminationCombined or Unified. If the params are all basic, then it'll just stick to using the speedy General table.
* Specifying cog_or_oversight will only trigger an Advanced search if "either" is NOT used
* This effects both Advanced Search and Beta Search

## How to test

* I deployed the branch to preview, this way we can compare [prod beta search](https://app.fac.gov/dissemination/search/beta/) with [preview beta search](https://fac-preview.app.cloud.gov/dissemination/search/beta/)
* Searching by basic params in beta should now be as fast as using basic search
  * "Basic param" here means any fields that can be found on the basic search page, like audit years, states, etc.
* Advanced search should be faster for basic fields, too
* Searches generally working as normal
* Specifying cog_or_oversight will only trigger an Advanced search if "either" is NOT used
* If testing locally, you can tell what table/model is being used via the log line `INFO search Searching <model name>`

### DIFF
--- a/backend/dissemination/search.py
+++ b/backend/dissemination/search.py
@@ -54,6 +54,38 @@ def is_beta_search(params_dict):
     return params_dict.get("beta_search_flag", False)
 
 
+GENERAL_PARAMS = [
+    'audit_years',
+    'uei_or_ein',
+    'entity_name',
+    'start_date',
+    'end_date',
+    'auditee_state',
+    'fy_end_month',
+    'report_id',
+    'page',
+    'order_by',
+    'order_direction',
+    'advanced_search_flag',
+    'beta_search_flag',
+    'LIMIT',
+]
+
+
+def only_searching_on_general(params_dict):
+    """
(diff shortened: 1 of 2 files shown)


RanyeM (commented):
LGTM

LynnMHouston:
I did not do a time test, but it works great for me!

RanyeM (approved):
Lgtm!

